PR #182, opened 2016-01-11
Add GLSupport Module to share native parts of the different GL RenderSystems

paroj:
the added GLSupport module includes the platform specific parts that are needed to create a GL Context. (i.e. GLX, WGL, Cocoa)
They had been copy/ pasted but stayed largely the same across the different GL Rendersytems.

The result is the `GLNativeSupport` API that can be plugged in the exiting `GL*Support` classes (e.g. `GL3PlusSupport`) to handle the context creation.

The idea is to provide one library that just handles the platform native context creation. Consequently the SDL and GTK based GLSupport classes are gone now. You have to use the `externalWindowHandle` in these cases.

## Context Pofiles
The API was also extended to allow specifying a context profile. This way the GL3Plus RS creates a core profile, while the old GL RS uses the compatibility profile.

## The Advantage
The obvious one is that we do not have duplicated code any more. However we can now also easily create a GLES2 context using GLX/WGL on the desktop. A change of the GLES2 RS to do so is in the making.

## Reviewing

the interesting commits are f45bfd344acf80558fcef9b82b9435e88f19ebd2 (GLNativeSupport API) and b74217b9c60cad2a3cb758381658bbe7a73be047 (port of the GL rendersytem to the new API)


paroj:
@darksylinc @holocronweaver mind to take a look?

holocronweaver:
I am in the process of graduating so I can't look in detail, but I will say that in my OGRE window system framework the window system usually handles context creation, which mostly eliminates the problem you are trying to solve. However, it will be 2-3 months before I can work on it again, and I see no harm in cutting down on redundant code in the meantime. From a glance your approach seems reasonable.

BTW, do you do your work on Github with CI? How do you get it synced up with Mercurial + Bitbucket?

paroj:
@holocronweaver https://github.com/fingolfin/git-remote-hg

PR #231, opened 2016-08-28
V2 1

fenix230:


paroj:
updated manually to safe merge commit.

PR #293, opened 2017-01-05
add support for non intrusive introspection to StringInterface

paroj:


paroj:
nowadays we want to get rid of StringInterface as far as possible. Maybe, this will still come in handy there.

PR #361, opened 2017-03-09
Fix codec registration. In Ogre::Root::Root, image codec registration…

fzimmerm:
While porting from v1-9 to v2-0, I ran into a problem where any png file wouldn't be loaded correctly by Ogre. A call to Ogre::Codec::getCodec would throw a OGRE EXCEPTION(5:ItemIdentityException): Can not find codec for 'png' image format.

I isolated the problem by debugging the Ogre::Root::Root constructor: getCodec("png") worked after registering the FreeImage codecs, but would fail after loading the plugins. Moving the image codec registration after loading the plugins solved the issue. In case other people are affected, here's a pull-request.

### DIFF
--- a/OgreMain/src/OgreRoot.cpp
+++ b/OgreMain/src/OgreRoot.cpp
@@ -229,24 +229,6 @@ namespace Ogre {
         ArchiveManager::getSingleton().addArchiveFactory( mEmbeddedZipArchiveFactory );
 #   endif
 
-#if OGRE_NO_DDS_CODEC == 0
-        // Register image codecs
-        DDSCodec::startup();
-#endif
-#if OGRE_NO_FREEIMAGE == 0
-        // Register image codecs
-        FreeImageCodec::startup();
-#endif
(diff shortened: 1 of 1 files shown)


paroj:
> worked after registering the FreeImage codecs, but would fail after loading the plugins.

could you find out why? The proposed fix prevents the plugins to query which codecs are supported by Ogre.

fzimmerm:
I didn't dig far enough into the plugin loading routines to find the root cause. The only plugin we're using is RenderSystem_GL. I'll have another look to see if I can find the line in OgreGLRenderSystem that causes the problem.
As an aside, I'm only testing with the debug build at the moment, so I'm not sure if the issue also appears in release builds.

paroj:
> Could it be that linking a debug build of ogre with a release build of RenderSystem_GL.so 

yes. this can work sometimes, but most of the time you will get strange bugs due to structure size mismatches.

fzimmerm:
I've confirmed that indeed the version mismatch was the problem. Sorry for wasting your time!

PR #365, opened 2017-03-16
Correction of EXP and LOG operations on Quaternions.

dcarlus:
The exponential of a quaternion was incorrect, the results were different from what can be found with Matlab or Octave.

The logarithm computation is improved by using the atan2 form to compute the angle. As mentioned in "Quaternion Computation" by Neil Dantam, it is best for numerical stability.
http://www.neil.dantam.name/note/dantam-quaternion.pdf

### DIFF
--- a/OgreMain/src/OgreQuaternion.cpp
+++ b/OgreMain/src/OgreQuaternion.cpp
@@ -376,10 +376,10 @@ namespace Ogre {
 
         if ( Math::Abs(w) < 1.0 )
         {
-			// According to Neil Dantam, atan2 has the best stability.
-			// http://www.neil.dantam.name/note/dantam-quaternion.pdf
-			Real fNormV = Math::Sqrt(x*x + y*y + z*z);
-			Radian fAngle ( Math::ATan2(fNormV, w) );
+            // According to Neil Dantam, atan2 has the best stability.
+            // http://www.neil.dantam.name/note/dantam-quaternion.pdf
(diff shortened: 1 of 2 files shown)


paroj:
thanks for your contribution.  could you also provide regression tests for this?
[see this for an example how to write them](https://github.com/OGRECave/ogre/blob/master/Tests/OgreMain/src/DualQuaternionTests.cpp#L52-L63)

dcarlus:
I am writing the tests.

While preparing the one on the Quaternion::Exp() method, I've seen something that was not clear in comparison to the formula (see commit https://github.com/OGRECave/ogre/pull/365/commits/3383bd830e19214065c84a09350e2a5eea55a5a0). I am not sure of your expectations about this part of code as it may have be done for a good reason.

paroj:
Real defaults to float in OGRE, therefore EXPECT_DOUBLE_EQ would still fail.

However, you have given the truth values only up to the fifth decimal place. Therefore you should use:
`EXPECT_NEAR(a, b, 1e-6)`

paroj:
>  I am not sure of your expectations about this part of code as it may have be done for a good reason.

that code has been broken for 15 years, so probably not. Also this is just a zero test and `sin(0) == 0`.

dcarlus:
Thank you for your help, I hope it will finally succeed! I am not familiar with googletest.
Indeed, the check is quite the same as testing the angle. It brings more sense to the code though.

PR #409, opened 2017-05-13
Fixed Example Include Dirs

emersont1:
Hi, this fixes the include dirs so that it will include the correct files if it's not being installed

### DIFF
--- a/Samples/2.0/ApiUsage/AnimationTagPoint/CMakeLists.txt
+++ b/Samples/2.0/ApiUsage/AnimationTagPoint/CMakeLists.txt
@@ -12,6 +12,10 @@ macro( add_recursive dir retVal )
 endmacro()
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
+ogre_add_component_include_dir(Hlms/Pbs)
+ogre_add_component_include_dir(Hlms/Unlit)
+ogre_add_component_include_dir(Hlms/Common)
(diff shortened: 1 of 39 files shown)


emersont1:
If it's any more convinient, i'll move it over to bitbucket

darksylinc:
Two things:
1. Yes, when it comes to 2.1 changes, it is best to do them in bitbucket (I work best on bitbucket, pavel works best on github and then he syncs the repos)
2. what is the problem this PR fixes? (please explain in detail)

emersont1:
It's for when building the samples, there are errors with including things like OgreOverlayPrerequisites and some HLMS headers (Example with MSVC, but it occours on linux/G++ too)
```
(ClCompile target) ->
  D:\game\thirdparty\Ogre\ogre\Samples\2.0\Common\include\GraphicsSystem.h(10): fatal error C1083: Cannot open include
file: 'OgreOverlayPrerequisites.h': No such file or directory (compiling source file D:\game\thirdpart\Ogre\ogre\Samples\2.0\Common\src\CameraController.cpp)
```

darksylinc:
What CMake version are you using, which compile settings and how are you compiling it?
Because I cannot get these errors in any of the platforms (Windows w/ MSVC2008-2015, Linux & iOS).

paroj:
to be continued at https://bitbucket.org/sinbad/ogre/pull-requests/759 as matias prefers bitbucket.

PR #417, opened 2017-05-31
FFP_AlphaTest Fixes

SNiLD:


paroj:
have you verified that this indeed works? I think there is a change needed here too:
https://github.com/OGRECave/ogre/blob/master/Components/RTShaderSystem/src/OgreShaderFFPRenderStateBuilder.cpp#L148

paroj:
and here - the case does not match the filename: https://github.com/OGRECave/ogre/blob/master/Components/RTShaderSystem/include/OgreShaderFFPRenderState.h#L129

paroj:
also take a look at https://github.com/erlang/otp/wiki/Writing-good-commit-messages

SNiLD:
I have this in use with OpenGL already and it works.

paroj:
> I have this in use with OpenGL already and it works.

could you provide a code snippet so I can test on my end as well? Just calling `ShaderGenerator::validateMaterial` should be broken because of the "Direct3D11" check.

If possible use the VTest framework as:
https://github.com/OGRECave/ogre/blob/master/Tests/VisualTests/PlayPen/src/PlayPenTests.cpp#L149-L175
so we can directly use it as a regression test.

SNiLD:
I'm sorry but it's part of large project and I don't currently have resources to rip it off of it. Maybe the Ogre samples could be updated for including this feature. I'm not familiar with the VTest framework and I have no idea how to use it.

SNiLD:
Also note that my test case is with OpenGL in Windows so the broken FFP_AlphaTest function name probably works because of that in Windows. Also note that the OpenGL function lib is just made of the HLSL version with the float4 changed to vec4 with the exact same base name.

paroj:
> I'm sorry but it's part of large project and I don't currently have resources to rip it off of it

thats fine, however it will take me some weeks until I can come back to this.

> Also note that my test case is with OpenGL in Windows

GL or GL3+?

SNiLD:
> GL or GL3+?

GL

paroj:
with GL `_setAlphaRejectSettings` sets the FFP glAlphaFunc parameter, which AFAIR overwrites any shader settings. Have you verified that AlphaTest was not working before your changes?
Note that I still want to merge these changes, I just want to clarify in advance which test I should perform.

SNiLD:
The problem is not global alpha rejection, but alpha rejection that you can set for a specific material. In my use case I want to use PerPixelLighting and my custom template sub render state that generates infrared view so I need to remove default templates. Some of my materials use `alpha_rejection` which can not be handled unless you include `FFPAlphaTest` sub render state. Because I have Ogre compiled as DLL, the `FFPAlphaTest` can not be added unless the class is declared with the export macro.

I had to make these changes specifically to get the `alpha_rejection` in the materials to work with RTSS. Only optional change is the `RTSHADER_SYSTEM_BUILD_CORE_SHADERS` macro change which does not make a difference if you compile with both defines defined.

paroj:
`_setAlphaRejectSettings` [is also used for per material settings](https://github.com/OGRECave/ogre/blob/49c51c0754b4556ab4bc0a79dc476804634256c4/OgreMain/src/OgreSceneManager.cpp#L5650).
which in turn calls `glAlphaFunc` which works even if you use GLSL shaders (RTSS). So the FFPAlphaTest SRS should be only required on GL3+ where `glAlphaFunc` is no longer available.

I am bringing this up because I think the FFPAlphaTest SRS is [not actually added on anything but D3D11](https://github.com/OGRECave/ogre/blob/master/Components/RTShaderSystem/src/OgreShaderFFPRenderStateBuilder.cpp#L148).

SNiLD:
Here is piece of code how I add the FFPAlphaTest ( the explanation of this is on  https://github.com/OGRECave/ogre/pull/417#issuecomment-306164596 ):

```cpp
Ogre::RTShader::RenderState* renderState = generator->getRenderState("ir");

// Remove default templates because we use per pixel lighting.
renderState->reset();

Ogre::RTShader::SubRenderState* infraRed = generator->createSubRenderState(Visual::clInfraRedRTShader::Type);
renderState->addTemplateSubRenderState(infraRed);

Ogre::RTShader::SubRenderState* perPixelLightModel = generator->createSubRenderState(Ogre::RTShader::PerPixelLighting::Type);
renderState->addTemplateSubRenderState(perPixelLightModel);

Ogre::RTShader::SubRenderState* alphaTest = generator->createSubRenderState(Ogre::RTShader::FFPAlphaTest::Type);
renderState->addTemplateSubRenderState(alphaTest);

// Invalidate the scheme in order to re-generate all shaders based technique related to this scheme.
generator->invalidateScheme("ir");
```
If I don't do this, I will not get alpha rejection on materials.

paroj:
with the latest changes you should be able to just use `Pass::setAlphaRejectSettings` the render state will be created on demand by `ShaderGenerator::validateMaterial`.

PR #429, opened 2017-06-30
Renderbuffer reuse

paroj:
from: http://www.ogre3d.org/forums/viewtopic.php?f=25&t=92915

# expected
 each Ogre::Texture would have a unique resolve texture, but would share the same MSAA surface.
# issue
 1. existing renderbuffers with the same format to not reused. Instead, a new renderbuffer is created with the same format, which overwrites the element in the map that might have been there prior to the call. 
2. Thus, when the original renderbuffer with the same format is supposed to be released, the new renderbuffer is found instead, and deleted. 
3. With the wrong renderbuffer deleted, the memory error occurs when a texture is deleted which refers to the incorrectly deleted renderbuffer.


paroj:
:+1:

PR #481, opened 2017-08-14
Improvements for macOS

fschopp:
### Fix build issues/inconveniences

Changes:
- Previously, CMAKE_PREFIX_PATH could effectively be ignored, due to the
  way it was extended.
- On Apple platforms, do not add “_d” suffix to libraries for debug
  builds. The pkg-config files do not have them, either.
- CMake variable SDL2_LIBRARY can contain whitespace on macOS, e.g., it
  contains “-framework Cocoa”. This caused problems in dependent
  projects relying on OGREConfig.cmake.
- Make building libraries as macOS frameworks optional (default is yes
  for backward compatibility).
- Make installing CMake Find<package> scripts optional. They can get in
  the way.
- Added .gitignore file, given that there is an official github mirror.
- Fix build failure in dependent projects, caused by recent changes to
  OgreMain/include/OgreFileSystemLayer.h, so that it now contains
  #include “macUtils.h”, even though macUtils.h is deeply nested in the
  include directory.

### [OSX] Fix failed assertion on macOS

Fixed the following failed assertion that would cause debug builds to
crash:

Assertion failed:
(mView == nil || int(mHeight)
  == _getPixelFromPoint([mView frame].size.height)),
function getHeight,
file /Users/.../RenderSystems/GLSupport/src/OSX/OgreOSXCocoaWindow.mm,
line 378.

### [OSX] Add option for NSOpenGLCPSurfaceOrder

Ogre::Root::createRenderWindow now understands key
“NSOpenGLCPSurfaceOrder” on macOS. This is needed for applications that
want to display user controls on top of the OpenGL view.

### [OSX] createRenderWindow(): "border" on macOS

Ogre::Root::createRenderWindow now supports the “border” key also on
macOS.


fschopp:
The [Gazebo](http://gazebosim.org) robot simulator needs support for `NSOpenGLCPSurfaceOrder` added by this pull request. See: https://bitbucket.org/osrf/gazebo/pull-requests/2764

paroj:
:+1:  thanks

fschopp:
Nice. Thanks for merging!

PR #494, opened 2017-08-21
Fix PDB installation on Win32

tkausl:
I've replaced all PDB install hardcoded paths with the `$<TARGET_PDB_FILE:tgt>` cmake generator expression, this seems to be the _cleanest_ solution, however, this _might_ break other systems which do not support PDB files.

The generator expression (https://github.com/thedmd/cmake-hacked/blob/master/Source/cmGeneratorExpressionEvaluator.cxx#L1598) checks for PDB support in the `"CMAKE_" + language + "_LINKER_SUPPORTS_PDB"` variable, unfortunately `get_target_property(LINKER_LANGUAGE, [target], LINKER_LANGUAGE)` always returned me NOTFOUND, so I couldn't check for PDB-Support. A hardcoded `CMAKE_CXX_LINKER_SUPPORTS_PDB` might work, not sure though. (Calling the generator-expression on unsupported targets breaks the build completely, thats why we need to check manually)

Just setting `OGRE_RELWDBG_PATH` to `/relwithdebinfo` and `OGRE_DEBUG_PATH` to `/debug` should have worked too, but this does also mean that the files are installed in `/bin/debug`, not in `/bin`. Any opinions on this?

### DIFF
--- a/CMake/Utils/OgreConfigTargets.cmake
+++ b/CMake/Utils/OgreConfigTargets.cmake
@@ -189,6 +189,16 @@ function(ogre_config_common TARGETNAME)
   ogre_create_vcproj_userfile(${TARGETNAME})
 endfunction(ogre_config_common)
 
+# checks whether the target LIBNAME produces a pdb file
+function(ogre_produces_pdb VARNAME LIBNAME)
+  get_target_property(TYPE ${LIBNAME} TYPE)
+  if (TYPE STREQUAL "SHARED_LIBRARY" OR TYPE STREQUAL "MODULE_LIBRARY" OR TYPE STREQUAL "EXECUTABLE")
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -6,7 +6,7 @@ build:
 
 build_script:
  - cmake -P ci-build.cmake
- - cmake --build .
+ - cmake --build . --config Debug --target INSTALL
```
paroj: please drop the `--target INSTALL` part here as it spams the build logs. otherwise this looks good to me.

tkausl: I've reset `appveyor.yml` and `ci-build.cmake` to the originals, these were just to test while this PR was WIP

PR #502, opened 2017-08-26
Added mappings for Visual Studio 2013, 2015 and 2017

majic79:
Small fix to allow support of Visual Studio 2017 (and earlier) built TBB libraries


OgreMain/include/Threading/OgreDefaultWorkQueueTBB.h
```diff
@@ -39,7 +39,7 @@ namespace Ogre
     {
     public:
         DefaultWorkQueue(const String& name = BLANKSTRING);
-        virtual ~DefaultWorkQueue();
+        virtual ~DefaultWorkQueue(void) noexcept(true);
```
paroj: we still use C++03 by default. Please remove C++11 features or ifdef them.

majic79:
Ah, I didn't realise - I've added a check as requested

PR #507, opened 2017-08-29
check that glx extension functions are found

wjwwood:
I'm trying to use Ogre 1.10.7 in a Linux (Ubuntu 16.04) VM on my macbook pro. My opengl version is 2.1 and GSLS version 1.2. I've been running into segfaults because some of the glx extension functions are not being found (i.e. `glXGetProcAddressARB` is returning `nullptr`. In this pull request I'm improving detection for these functions (so you get a runtime error instead of a segfault), but I'm still not sure why these functions are not loading. I'm wondering if the 1.10 releases have been tested on opengl 2.1 very well and maybe some unavailable opengl 3+ functions are being used in the common gl support part of Ogre.

My setup is admittedly a bit exotic (doing static linking and integrating into a Qt5 QWindow), but I can follow up on why these functions are not loaded in my case in a separate issue if this pr is acceptable as-is. I'm hoping someone has an idea of why these are not available.

Also I was able to use this patch to get past the missing `glXCreateContextAttribsARB` function but I cannot get past the missing `_glXSwapIntervalSGI` function:

```diff
diff --git a/RenderSystems/GLSupport/src/GLX/OgreGLXGLSupport.cpp b/RenderSystems/GLSupport/src/GLX/OgreGLXGLSupport.cpp
index 278221b79..aca4d2ecb 100644
--- a/RenderSystems/GLSupport/src/GLX/OgreGLXGLSupport.cpp
+++ b/RenderSystems/GLSupport/src/GLX/OgreGLXGLSupport.cpp
@@ -865,7 +865,11 @@ namespace Ogre
         while(!glxContext && (context_attribs[1] >= minVersion))
         {
             ctxErrorOccurred = false;
-            glxContext = _glXCreateContextAttribsARB(mGLDisplay, fbConfig, shareList, direct, context_attribs);
+            if (!_glXCreateContextAttribsARB) {
+                glxContext = glXCreateNewContext(mGLDisplay, fbConfig, GLX_RGBA_TYPE, shareList, TRUE);
+            } else {
+                glxContext = _glXCreateContextAttribsARB(mGLDisplay, fbConfig, shareList, direct, context_attribs);
+            }
             // Sync to ensure any errors generated are processed.
             XSync( mGLDisplay, False );
             if ( !ctxErrorOccurred && glxContext )
```

### DIFF
--- a/RenderSystems/GLSupport/src/GLX/OgreGLXGLSupport.cpp
+++ b/RenderSystems/GLSupport/src/GLX/OgreGLXGLSupport.cpp
@@ -856,6 +856,12 @@ namespace Ogre
         PFNGLXCREATECONTEXTATTRIBSARBPROC _glXCreateContextAttribsARB;
         _glXCreateContextAttribsARB = (PFNGLXCREATECONTEXTATTRIBSARBPROC)const_cast<GLXGLSupport*>(this)->getProcAddress("glXCreateContextAttribsARB");
 
+        if (!_glXCreateContextAttribsARB) {
+            std::string msg = "glXCreateContextAttribsARB() function not found";
+            LogManager::getSingleton().logMessage(LML_CRITICAL, msg);
+            throw std::runtime_error(msg);
+        }
+
         while(!glxContext && (context_attribs[1] >= minVersion))
         {
             ctxErrorOccurred = false;
--- a/RenderSystems/GLSupport/src/GLX/OgreGLXGLSupport.cpp
+++ b/RenderSystems/GLSupport/src/GLX/OgreGLXGLSupport.cpp
@@ -856,11 +856,7 @@ namespace Ogre
         PFNGLXCREATECONTEXTATTRIBSARBPROC _glXCreateContextAttribsARB;
         _glXCreateContextAttribsARB = (PFNGLXCREATECONTEXTATTRIBSARBPROC)const_cast<GLXGLSupport*>(this)->getProcAddress("glXCreateContextAttribsARB");
 
-        if (!_glXCreateContextAttribsARB) {
-            std::string msg = "glXCreateContextAttribsARB() function not found";
-            LogManager::getSingleton().logMessage(LML_CRITICAL, msg);
-            throw std::runtime_error(msg);
-        }
+	OgreAssert(_glXCreateContextAttribsARB, "glXCreateContextAttribsARB() function not found");
 
         while(!glxContext && (context_attribs[1] >= minVersion))
         {
--- a/RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
+++ b/RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
@@ -601,8 +601,10 @@ namespace Ogre
             }
             else if( _glXSwapIntervalMESA )
                 _glXSwapIntervalMESA( vsync ? mVSyncInterval : 0 );
-            else
+            else if( _glXSwapIntervalSGI )
                 _glXSwapIntervalSGI( vsync ? mVSyncInterval : 0 );
+            else
+                throw std::runtime_error("no glx swap interval function found");
         }
 
         mContext->endCurrent();
--- a/RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
+++ b/RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
@@ -601,9 +601,10 @@ namespace Ogre
             }
             else if( _glXSwapIntervalMESA )
                 _glXSwapIntervalMESA( vsync ? mVSyncInterval : 0 );
-            else
+            else {
 		OgreAssert(_glXSwapIntervalSGI, "no glx swap interval function found");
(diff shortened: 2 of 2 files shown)


RenderSystems/GLSupport/src/GLX/OgreGLXGLSupport.cpp
```diff
@@ -856,6 +856,12 @@ namespace Ogre
         PFNGLXCREATECONTEXTATTRIBSARBPROC _glXCreateContextAttribsARB;
         _glXCreateContextAttribsARB = (PFNGLXCREATECONTEXTATTRIBSARBPROC)const_cast<GLXGLSupport*>(this)->getProcAddress("glXCreateContextAttribsARB");
 
+        if (!_glXCreateContextAttribsARB) {
+            std::string msg = "glXCreateContextAttribsARB() function not found";
+            LogManager::getSingleton().logMessage(LML_CRITICAL, msg);
+            throw std::runtime_error(msg);
```
paroj: please rather use `OgreAssert(cond, "msg")` which will throw the appropriate exception.

RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
```diff
@@ -601,8 +601,10 @@ namespace Ogre
             }
             else if( _glXSwapIntervalMESA )
                 _glXSwapIntervalMESA( vsync ? mVSyncInterval : 0 );
-            else
+            else if( _glXSwapIntervalSGI )
                 _glXSwapIntervalSGI( vsync ? mVSyncInterval : 0 );
+            else
+                throw std::runtime_error("no glx swap interval function found");
```
paroj: same

wjwwood:
I updated both checks to use `OgreAssert`. Naive question are those checked in release mode?

Also, any idea why those functions would not be found (i.e. end up `nullptr`) other than they might not be available pre opengl 3?

RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
paroj: here you can make the calls conditional as ´!mIsExternalGLControl && vsync`

paroj:
this issue is not related to GL, but to GLX (which is the API that creates the actual GL context, c.f. EGL). Your emulator seems to only offer GLX 1.3 without any extensions which is not what you encounter on actual linux systems. (with MESA or NVidia implementations)

Regarding OgreAssert: https://ogrecave.github.io/ogre/api/1.10/_ogre_exception_8h.html#ab64188d2d6f4a02714a8f02e23a491d2

wjwwood:
@paroj For what it is worth I see the same thing on my macbook running in macOS. At least I'm pretty sure. I can check again tomorrow.

Thanks for the link to the docs.

RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
wjwwood: Do you mean the whole block? I wouldn't have thought that was ok since not calling the function is not the same as calling it with `0` right? What if vsync was on but was then turned off? Maybe that situation wouldn't happen, idk. I'm happy to do what ever you suggest, but can you clarify if you mean the entire block starting with `if (! mIsExternalGLControl )` or a narrower scope, something like in the final else case:

```c++
if (_glXSwapIntervalSGI) {
  _glXSwapIntervalSGI( vsync ? mVSyncInterval : 0 );
} else if (vsync) {
  // throw exception because vsync requested and _glXSwapIntervalSGI not found
}  // implicit else, pass on missing function because no vsync requested
```

RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
```diff
@@ -602,6 +602,7 @@ namespace Ogre
             else if( _glXSwapIntervalMESA )
                 _glXSwapIntervalMESA( vsync ? mVSyncInterval : 0 );
             else
+		OgreAssert(_glXSwapIntervalSGI, "no glx swap interval function found");
```
paroj: braces required in else block

RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
wjwwood: Edited the previous comment to say "not ok" rather than "ok"...

RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
wjwwood: Doh, obviously... thanks.

RenderSystems/GLSupport/src/GLX/OgreGLXWindow.cpp
wjwwood: Sorry reading this again, that sentence makes no sense. What I meant to say was that I did not think it would be acceptable to not call one of the swap interface functions when vsync was false, because I assumed not calling one of those functions and calling one of those functions with argument `0` would have different outcomes.

PR #510, opened 2017-09-04
Corrected an error in Mutex locks

majic79:
Thread Mutexes will not exist if Thread Support is not enabled. Thread support is only enabled when option 1 or 2 is selected. I've expanded this to include option 3 & 4 (per comment text)


paroj:
there is no OGRE_THREAD_SUPPORT=3 in v2-1. Also OGRE_LOCK_MUTEX should evaluate to nothing when there is no thread support. Can you elaborate on the actual error you try to fix here?

majic79:
The main issue is with OgreConvexBody.cpp - the header conditionally declares the msFreePolygonsMutex while the code unconditionally attempts to grab locks on it. It seems more appropriate to make use of the lock on the condition that threading support is available, than it does to unconditionally decalare the mutex.

Whilst there may not be support (currently) for OGRE_THREAD_SUPPORT=3, this news post (admittedly with regard to 1.10) http://www.ogre3d.org/2017/09/02/ogre-1-10-mid-term-report suggests that OGRE_THREAD_SUPPORT=3 may be here to stay. Irrespective of that, the comment following the OGRE_THREAD_SUPPORT redefinitions suggests that values 0 through 4 are valid. As I don't know enough about the development direction, this seemed simplest

paroj:
> The main issue is with OgreConvexBody.cpp - the header conditionally declares the msFreePolygonsMutex 

ah, yeah this is misleading `OGRE_STATIC_MUTEX` also evaluates to nothing without thread support. Removed the misleading condition when introducing thread support  = 3 in 1.10.

> OGRE_THREAD_SUPPORT redefinitions suggests that values 0 through 4 are valid.

OGRE_THREAD_SUPPORT != OGRE_THREAD_PROVIDER

PR #512, opened 2017-09-05
Basic Tutorial 1 - Porting from wiki

bohdankornienko:
I have just transferred content of the wiki manual and modified code a little bit. 
The idea is to adopt whole tutorials course with ApplicationContext instead of BasicApplicaion class. And of course to adopt the idea that all valid tutorials placed in manual pages.


Docs/src/tutorials/basictutorials/code/basictutorial1fullsource.md
```diff
@@ -0,0 +1,151 @@
...
+};
+
+#endif /* SRC_TUTORIALAPPLICATION_HPP_ */
+
+
+```
+
+## Source file
+
+```cpp
```
paroj: could you instead of pasting the c++ source into the doc, reference a compilable source file using the @snippet command?

see:
- https://github.com/OGRECave/ogre/blob/master/Docs/src/tutorials/setup.md
- https://github.com/OGRECave/ogre/blob/master/Samples/Bootstrap/main.cpp

store the sources as `Samples/Tutorials/basic1.cpp` etc., if you can also provide a CMakeLists to compile them.

This way we can ensure that the tutorials are up to date - even if the API changes.

Docs/src/tutorials/basictutorials/code/basictutorial1fullsource.md
```diff
@@ -0,0 +1,151 @@
...
+
+This source file is a part of OGRE
+
+#ifndef SRC_TUTORIALAPPLICATION_HPP_
+#define SRC_TUTORIALAPPLICATION_HPP_
+
+#include <OGRE/Ogre.h>
+#include <OGRE/Bites/OgreApplicationContext.h>
```
paroj: if you correctly use `${OGRE_INCLUDE_DIRS}` (see https://ogrecave.github.io/ogre/api/1.10/setup.html) these should be just `<Ogre.h>`, `<OgreApplicationContext.h>` etc.

Docs/src/tutorials/basictutorials/code/tutorialapplicationcpp.md
```diff
@@ -0,0 +1,131 @@
+# Tutorial application startup source file {#tutorialapplicationcpp}
```
paroj: this just duplicates the Bootstrap sample, so we should drop it

Docs/src/tutorials/basictutorials/basictutorial1.md
```diff
@@ -0,0 +1,347 @@
...
+
+Compile and run your application. Your Ogre head should have a buddy.
+
+![](bt1_added_entity.png)
+
+## More About Entities
+
+The Ogre::Entity class is very extensive. We will now introduce just a few more of its methods that will be useful. The Entity class has setVisible and isVisible methods. If you want an Entity to be hidden, but you still need it later, then you can use this function instead of destroying the Entity and rebuilding it later.
+
+**Note:** Entities do not need to be pooled like they are in some graphics engines. Only one copy of each mesh and texture is ever loaded into memory, so there is not a big savings from trying to minimize the number of Entities.
```
paroj: use @note for a consisten formatting

Docs/src/tutorials/basictutorials/basictutorial1.md
```diff
@@ -0,0 +1,347 @@
...
+
+#### Third-party Plugins
+
+The last major group contains third-party libraries and other general support libraries. Ogre is focused sharply on being a graphics rendering library. This group makes it easy to integrate external libraries to add things like physics, input, and GUI systems. These libraries are used together to form a full game development environment. You might find this piecemeal approach a little strange, but it is a very common design pattern in large software projects. It is harder to comprehend at first, but it is a much more flexible approach when you want to start building more complicated scenes.
+
+The Ogre demos and SDK include some of these third-party libraries. The [Open Input System](http://www.ogre3d.org/tikiwiki/tiki-index.php?page=OIS) can be used to manage input events and distribute them to Ogre. This is contained in OIS.dll or libOIS.so. You can also make use of Cg, which is used by CgProgramManager. This library allows you to produce materials with custom shaders. There are other libraries (not included with Ogre) that offer functionality such as sound and physics.
```
paroj: with 1.10 we have replaced OIS by SDL2

Docs/src/tutorials/basictutorials/basictutorial1.md
```diff
@@ -0,0 +1,347 @@
...
+
+Ogre uses several configuration files (\*.cfg). They control things like which plugins are loaded and where your application will search for resource files. We will briefly introduce you to each of these files. You'll slowly read more about them as you progress through the tutorials as well.
+
+#### Plugin Configuration
+
+**plugins.cfg**
+
+This file tells Ogre which plugins to load. You modify this file when you want to load a different set of plugins. It is often most useful to simply "comment out" lines instead of removing them, because you never know when a stroke of inspiration will mean you want to reload some unused plugins. Here is some sample content:
+
+<pre>
```
paroj: instead of pre you can use the ``` without specifying a language

Docs/src/tutorials/basictutorials/basictutorial1.md
```diff
@@ -0,0 +1,347 @@
...
+
+When you run your application, notice how your Ogre head is facing towards the camera down the positive z-axis. This is a property of the mesh itself and the orientation of the camera. Cameras are covered in a later tutorial. The Ogre head is sitting at the origin of our world, (0, 0, 0). The direction the head is facing by default is a result of which way it was facing when it was originally modeled. You can effectively change this from within Ogre as well, but it will require some knowledge of quaternions, which aren't really covered until the [Intermediate Tutorials](http://www.ogre3d.org/tikiwiki/tiki-index.php?page=Intermediate+Tutorials).
+
+[//]: <> (TODO: Replace link with manual page when time has come)
```
paroj: better put a dummy link like [this] (#) instead of linking outdated pages.

Docs/src/tutorials/basictutorials/code/basictutorial1fullsource.md
bohdankornienko: Making Cmake project for tutorials is good idea! 
The question is what structure of the project should I make?

For the first glance it could be main CMake file and then set of small child cmake projects. But in this case main.cpp will be copy and pasted for every project. 

I see the project structure as this:

<pre>
CMakeLists.txt
+basic1
   - CMakeLists.txt
   - main.cpp
   - BasicTutorial1.cpp
   - BasicTutorial1.hpp
+basic2
   - CMakeLists.txt
   - main.cpp
   - BasicTutorial1.cpp
   - BasicTutorial1.hpp
<pre>

In this case we can have a set of targets which will be compiled all at once.

Docs/src/tutorials/basictutorials/code/tutorialapplicationcpp.md
bohdankornienko: Yes it is. But for sake of study (and Basic Tutorial 1 suggest it) person moves along the declaration of entities and learns things. 
We could not drop it since the person has to start with something.

Docs/src/tutorials/basictutorials/basictutorial1.md
bohdankornienko: Basically we just left stubs until we add some Intermediate tutorials. Right?

Docs/src/tutorials/basictutorials/basictutorial1.md
bohdankornienko: Does it mean that we should remove any notion of OIS in tutorials at all? 

Basic Tutorial 6
http://www.ogre3d.org/tikiwiki/tiki-index.php?page=Basic+Tutorial+6&structure=Tutorials

has full source (which built application from from scratch) and it works on 1.10. In this example OIS is used as input system. So I was expecting to left room for this part as well so that learner can see that Ogre is able to work not only with SDL as an input.

paroj: > Does it mean that we should remove any notion of OIS in tutorials at all?

yes, OIS is dead. If anything I would make an example for Qt. But we should focus on the core parts first..

Docs/src/tutorials/basictutorials/basictutorial1.md
paroj: yes

Docs/src/tutorials/basictutorials/code/basictutorial1fullsource.md
paroj: as the point of the tutorials are the Ogre concepts and not c++ best practices, I would just put everything in one .cpp file (again see Bootstrap sample) as:
```
CMakeLists.txt
- add_executable(BasicTutorial1 BasicTutorial1.cpp)
- ...
- add_executable(BasicTutorialN BasicTutorialN.cpp)

BasicTutorial1.cpp
...
BasicTutorialN.cpp
```

Docs/src/tutorials/basictutorials/code/tutorialapplicationcpp.md
paroj: why not start with bootstrap?

Docs/src/tutorials/basictutorials/code/basictutorial1fullsource.md
bohdankornienko: Yeah. That is what I thought later would be better in this case.
Will do it in this way.

Docs/src/tutorials/basictutorials/basictutorial1.md
bohdankornienko: I see. Then will redo everything in SDL input context.

Docs/src/tutorials/basictutorials/code/tutorialapplicationcpp.md
bohdankornienko: In my opinion it iw better not know about bootstrap while one learning Ogre for first time. It is nice to have it for reference and for someone who knows a bit about engine.
Reading bootstrap code one can be just overwhelmed by its complexity. It is better to introduce concepts step by step. Original tutorials have this notion. And for me as first time learner of it it made positive.

Another point is we leaving room for keeping idea with tutorials as with basic application in case if bootstrap suddenly changed to be not appropriate for tutorials. But it is just subjective.

Samples/Tutorials/BasicTutorial1.hpp
```diff
@@ -0,0 +1,54 @@
...
+THE SOFTWARE
+-------------------------------------------------------------------------*/
+
+//! [fullsource]
+
+#ifndef __BASIC_TUTORIAL_1__
+#define __BASIC_TUTORIAL_1__
+
+#include <Ogre.h>
+#include <Bites/OgreApplicationContext.h>
```
paroj: please use the header without the directory prefix. e.g.
`<OgreApplicationContext.h>` instead of `<Bites/OgreApplicationContext.h>`

Docs/src/tutorials/basictutorials/code/basictutorial1fullsource.md
```diff
@@ -0,0 +1,8 @@
+# Basic Tutorial 1 - Full source code {#basictutorial1fullsource}
```
paroj: do we really need the full source in Doxygen if it is already ready to built in the source tree?

Docs/src/tutorials/basictutorials/basictutorials.md
```diff
@@ -0,0 +1,5 @@
+# Basic Tutorials {#basictutorials}
+
+Group devoted to basic totorials.
```
paroj: typo

Docs/CMakeLists.txt
```diff
@@ -28,10 +28,11 @@ if(DOXYGEN_FOUND)
 	PATTERN "src" EXCLUDE
       )
     endif ()
-    
+
     file(GLOB TUTORIALS ${OGRE_SOURCE_DIR}/Docs/src/tutorials/*.md)
-    
-    set(CMAKE_DOXYGEN_INPUT_LIST 
+    file(GLOB_RECURSE BASIC_TUTORIALS ${OGRE_SOURCE_DIR}/Docs/src/tutorials/basictutorials/*.md)
```
paroj: please use GLOB instead of GLOB_RECURSE here

Docs/src/tutorials/basictutorials/basictutorial1.md
```diff
@@ -0,0 +1,302 @@
...
+
+One final thing to keep in mind about SceneNodes is that their position is **always** relative to their parent SceneNode and each SceneManager creates a root Node to which all other SceneNodes are attached.
+
+### Entity
+
+An Entity is one type of object that you can render in your scene. It is anything that is represented by a 3D mesh. Even terrain objects are very large entities. Lights, Billboards, Particles, and Cameras are examples of scene elements that are not entities. Ogre uses a well-known design pattern that separates renderable objects from information like their location. This means that you don't directly place an Entity into your scene. Instead, you place a SceneNode into your scene, then attach your Entity to that SceneNode. The Entity is then rendered using information taken from the SceneNode.
+
+## Getting started with codding
```
paroj: typo

Docs/src/tutorials/basictutorials/basictutorial1.md
```diff
@@ -0,0 +1,302 @@
+# Basic Tutorial 1 {#basictutorial1}
+
```
paroj: the tutorial is rather lenghty - you should add:
`@tableofcontents`
this will require you to move all headers one level up and add a tag i.e. `## How Ogre Works` -> `# How Ogre Works {#howogreworks}`.

paroj: we should keep the "TUTORIAL INTRODUCTION" text from the wiki that gives an outline.

paroj:
thanks for the update. As this lays the foundation for further tutorials, I added some more comments on the structure.

Docs/src/tutorials/basictutorials/basictutorial1.md
bohdankornienko: Can you clarify please?

Docs/src/tutorials/basictutorials/code/basictutorial1fullsource.md
bohdankornienko: Not sure. What if the person will not have access to sources? Like on windows with SDK.
Personally for me it was good to have link for full source when I was using tutorials. There could be missing something and direct reference to full source may clear things up. 

I can remove it if you say so.

Docs/src/tutorials/basictutorials/basictutorial1.md
paroj: bildschirmfoto vom 2017-09-10 19-45-28

Docs/src/tutorials/basictutorials/code/basictutorial1fullsource.md
paroj: one will most likely read the docs online, where one can easily find the according source on bitbucket/ github. Therefore I would remove it here.

Docs/src/tutorials/basictutorials/basictutorial1.md
bohdankornienko: I see. I got this! 
Will be done.

Docs/src/tutorials/basictutorials/code/basictutorial1fullsource.md
bohdankornienko: I see. I just pushed another change. 
Let me cleanup from full source reference then.
And then I can just refer to Samples/Tutorials/BasicTutorial1.cpp files for full source.

PR #515, opened 2017-09-07
GL*: use depth buffer as stencil buffer pointer for packed formats

fbridault:
This caused the stencil buffer to be unbound when using stencil test for these formats. 

The bug was introduced at commit 369fdbbfe95b3f5b4991c6718b4e08d07b4a35d0. I'm not sure it is the best way to fix it but that looked safer than modifying getBestDepthStencil(). 
Also beware I did the modifications for all render systems but I could only test it on GL3Plus.


paroj:
thanks

PR #558, opened 2017-10-27
Fix V734 warnings from PVS-Studio Static Analyzer

SerVerXPlanet:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. A bug, found using PVS-Studio.
Warning:

- [V734](https://www.viva64.com/en/w/v734/) An excessive check. Examine the conditions containing search for the substrings "int" and "uint". [ogrescripttranslator.cpp 513](https://github.com/OGRECave/ogre/blob/master/OgreMain/src/OgreScriptTranslator.cpp#L513)


paroj:
please convert tabs to spaces

PR #571, opened 2017-11-15
Compiler warning fixes for Visual Studio 2010

SNiLD:


Components/Terrain/src/OgreTerrainLodManager.cpp
```diff
@@ -66,7 +66,14 @@ namespace Ogre
 
     bool TerrainLodManager::isOpen() const
     {
+#ifndef OGRE_HAS_CXX11_EXPLICIT_CONVERSION_OPERATORS
+#   pragma warning ( push )
+#   pragma warning ( disable: 4800 )
+#endif
         return (bool)mDataStream;
```
paroj: I think `mDataStream != 0` would be a better fix instead of using the pragmas and associated compiler checks for this.

SNiLD: I agree. I'll try this.

Components/Terrain/src/OgreTerrainLodManager.cpp
```diff
@@ -66,7 +66,11 @@ namespace Ogre
 
     bool TerrainLodManager::isOpen() const
     {
+#ifndef OGRE_HAS_CXX11_EXPLICIT_CONVERSION_OPERATORS
+        return mDataStream != 0;
```
paroj: sorry, this should have been ´mDataStream.get() != 0;` as SharedPtr lacks a != operator for ptr. Also you can drop the ifdefs as this method will also work with OGRE_HAS_CXX11_EXPLICIT_CONVERSION_OPERATORS

OgreMain/include/OgreResourceManager.h
```diff
@@ -361,6 +361,10 @@ namespace Ogre {
         */
         virtual ResourcePtr getByHandle(ResourceHandle handle);
         
+#ifndef OGRE_HAS_CXX11_EXPLICIT_CONVERSION_OPERATORS
```
paroj: you missed this part

Components/Terrain/src/OgreTerrainLodManager.cpp
SNiLD: Ok, I'll change it. I wonder why it compiled without warnings.

SNiLD: Should I drop the new `OGRE_HAS_CXX11_EXPLICIT_CONVERSION_OPERATORS` define as well since it's no longer needed?

paroj: yes

Issue #437, opened 2017-07-09
GpuProgramParameters::copyMatchingNamedConstantsFrom invalidates GLUniformReferences


paroj:
Segfaults with GL StateCache

ASAN trace:
```=================================================================
==20373==ERROR: AddressSanitizer: heap-use-after-free on address 0x60b000129d08 at pc 0x7fa8edb293af bp 0x7ffd14a60120 sp 0x7ffd14a60110
READ of size 2 at 0x60b000129d08 thread T0
    #0 0x7fa8edb293ae in Ogre::GLSL::GLSLLinkProgram::updateUniforms(Ogre::SharedPtr<Ogre::GpuProgramParameters>, unsigned short, Ogre::GpuProgramType) ../RenderSystems/GL/src/GLSL/src/OgreGLSLLinkProgram.cpp:287
    #1 0x7fa8edb27896 in Ogre::GLSL::GLSLGpuProgram::bindProgramParameters(Ogre::SharedPtr<Ogre::GpuProgramParameters>, unsigned short) ../RenderSystems/GL/src/GLSL/src/OgreGLSLGpuProgram.cpp:123
    #2 0x7fa8edbc13c6 in Ogre::GLRenderSystem::bindGpuProgramParameters(Ogre::GpuProgramType, Ogre::SharedPtr<Ogre::GpuProgramParameters>, unsigned short) ../RenderSystems/GL/src/OgreGLRenderSystem.cpp:3053
    #3 0x7fa8f9f0f7c2 in Ogre::SceneManager::updateGpuProgramParameters(Ogre::Pass const*) ../OgreMain/src/OgreSceneManager.cpp:7367
    #4 0x7fa8f9f0254a in Ogre::SceneManager::_issueRenderOp(Ogre::Renderable*, Ogre::Pass const*) ../OgreMain/src/OgreSceneManager.cpp:7411
    #5 0x7fa8f9f38242 in Ogre::SceneManager::renderSingleObject(Ogre::Renderable*, Ogre::Pass const*, bool, bool, Ogre::HashedVector<Ogre::Light*> const*) ../OgreMain/src/OgreSceneManager.cpp:3552
    #6 0x7fa8f9effcf2 in Ogre::SceneManager::SceneMgrQueuedRenderableVisitor::visit(Ogre::Renderable*) ../OgreMain/src/OgreSceneManager.cpp:2987
[...]

PR #576, opened 2017-11-20
Possible fix for issue #437

SNiLD:
Probably my solution doesn't perfectly fit what you had planned for but this worked in my environment (Ogre 1.10.9).

### DIFF
--- a/RenderSystems/GL/src/GLSL/src/OgreGLSLLinkProgramManager.cpp
+++ b/RenderSystems/GL/src/GLSL/src/OgreGLSLLinkProgramManager.cpp
@@ -104,7 +104,12 @@ namespace Ogre {
     {
         // if there is an active link program then return it
         if (mActiveLinkProgram)
+        {
+            if (!mActiveLinkProgram->isValid())
+                mActiveLinkProgram->activate();
+
             return mActiveLinkProgram;
+        }
 
         // no active link program so find one or make a new one
         // is there an active key?
--- a/RenderSystems/GL/src/GLSL/src/OgreGLSLLinkProgramManager.cpp
+++ b/RenderSystems/GL/src/GLSL/src/OgreGLSLLinkProgramManager.cpp
@@ -104,12 +104,7 @@ namespace Ogre {
     {
         // if there is an active link program then return it
         if (mActiveLinkProgram)
-        {
-            if (!mActiveLinkProgram->isValid())
-                mActiveLinkProgram->activate();
-
             return mActiveLinkProgram;
-        }
 
         // no active link program so find one or make a new one
         // is there an active key?
@@ -400,6 +395,31 @@ namespace Ogre {
             } // end if
         } // end for
 
+    }
+    //---------------------------------------------------------------------
+    void GLSLLinkProgramManager::removeShader(GLSLProgram* gpuProgram)
+    {
+        for (LinkProgramMap::iterator iLinkProgram = mLinkPrograms.begin();
+             iLinkProgram != mLinkPrograms.end();)
+        {
+            if (iLinkProgram->second->getVertexProgram() == gpuProgram ||
+                iLinkProgram->second->getFragmentProgram() == gpuProgram ||
+                iLinkProgram->second->getGeometryProgram() == gpuProgram)
+            {
+                if (mActiveLinkProgram == iLinkProgram->second)
+                {
+                    mActiveLinkProgram = NULL;
+                    // change back to fixed pipeline
+                    glUseProgramObjectARB(0);
+                }
+                delete iLinkProgram->second;
+                iLinkProgram = mLinkPrograms.erase(iLinkProgram);
+            }
+            else
+            {
+                ++iLinkProgram;
+            }
+        }
     }
 }
 }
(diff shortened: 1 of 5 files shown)


paroj:
how about putting `mGLUniformReferences` into the respective {fragment, vertex, geom} programs where they would be cleared on unload()?

I think this idea had some issues, but I do not recall them any more.. :confused:

paroj:
the underlying issue is probably in `GLSLLinkProgramManager::setActiveXXXShader` which only checks the pointer but not the program status..

SNiLD:
Yes. The program definitely has to be re-linked and re-created since the old programs have been destroyed as well.

paroj:
you solution is definitely the right approach, I just think it is at the wrong place.

It basically duplicates this code:
https://github.com/OGRECave/ogre/blob/73928a19d70966a6c6c18624b6a7034ba6b559da/RenderSystems/GL/src/GLSL/src/OgreGLSLLinkProgramManager.cpp#L111-L124

which seems not to be working correctly - so we rather should fix the code above,

SNiLD:
Shader ID stays the same because it is only assigned during the construction of the object. GLHandle however changes depending on the unloading/loading. If the code you referred was changed to use GLHandle instead it would find/create new programs but there would be no way to find the old program and destroy it (which would lead to memory leaks).

paroj:
The code is already broken, as it leaks if you remove the HighLevelGpuProgram instead of just reloading it. I would rather  fix one issue at a time and consider the leaking a separate issue.

SNiLD:
Sounds like HighLevelGpuProgram needs mechanism to notify listeners that it will be removed/loaded/unloaded and GLSLLinkProgramManager needs to lister for these. Doesn't the ResourceManager provide this type of functionality? That would fix both of these issues with one go (memory leak and unload/reload).

paroj:
actually, yes
https://github.com/OGRECave/ogre/blob/6c2c08fb5b609e1f2ee9deb771348ec53fa78394/OgreMain/include/OgreResource.h#L423

SNiLD:
Nice. Except it doesn't provide information when the resource is deleted. However this probably could be done through GLSLGpuProgram destructor. It could notify GLSLLinkProgramManager that the `mGLSLProgram` is no longer valid and GLSLLinkProgramManager should delete the appropriate GLSLLinkProgram.

paroj:
actually we can always delete the program on unload (which also happens on deletion), as the program has to be equally re-created in both cases.

SNiLD:
Maybe GLSLProgram should override the `GLSLShaderCommon::unloadImpl()` and first tell GLSLLinkProgramManager that the resource will be gone and the reset the shared pointer.

paroj:
the problem is that GLSLProgram (shader actually) knows nothing of its parent Program. Also the solution should scale to GL3+ where we have separable programs (a shader can be part of multiple programs).
Therefore I prefer the listener approach. It might be slower but this is neglectable compared to the incurred shader compilation.

SNiLD:
Actually `GLSLGpuProgram::unloadImpl()` should probably notify GLSLLinkProgramManager as it already notifies on bindProgram and unbindProgram. Problem is that once one GLSLGpuProgram is gone, it might affect many GLSLLinkPrograms and all of the should be searched for removal.

SNiLD:
This new method for removal should also take care of the possible memory leaks. The removal method is bit brutal since it's brute force but I'm under the impression that unloads are not that common? This way at least the code stays simple.

Btw. for some reason the diff shows modifications from my previous commit as well although I've undone them.

SNiLD:
I guess same issue persists on GL3+ side so that should include similar fix.

RenderSystems/GL/src/GLSL/src/OgreGLSLLinkProgramManager.cpp
```diff
@@ -396,5 +396,30 @@ namespace Ogre {
...
+                iLinkProgram->second->getGeometryProgram() == gpuProgram)
+            {
+                if (mActiveLinkProgram == iLinkProgram->second)
+                {
+                    mActiveLinkProgram = NULL;
+                    // change back to fixed pipeline
+                    glUseProgramObjectARB(0);
+                }
+                delete iLinkProgram->second;
+                iLinkProgram = mLinkPrograms.erase(iLinkProgram);
```
paroj: https://stackoverflow.com/questions/263945/what-happens-if-you-call-erase-on-a-map-element-while-iterating-from-begin-to

paroj:
fun fact: this is already implemented for GLES2:
https://github.com/OGRECave/ogre/blob/a3d7681b0fad5549138af57746a6cd8fcc9d5fd9/RenderSystems/GLES2/src/GLSLES/src/OgreGLSLESLinkProgramManager.cpp#L145-L163

SNiLD:
Hey, unifying the shader stuff is the right way to go. Should I close this pull request?

SNiLD:
Also kind of related is that `mUniformRefsBuilt` in `GLSLLinkProgram` shadows the `mUniformRefsBuilt` of `GLSLProgramCommon`. Which I had resolved in my commit but don't see in #577 .

paroj:
yes, lets move further discussion to #577

PR #581, opened 2017-11-25
Fixed issue with indices.

erikogenvik:
Depending on the compiler, inserting into a map while reading the size
of in the same line produces different results. So on some machines the
indices would be messed up if a mesh with shared vertex data was used.

See https://blog.jayway.com/2015/09/08/undefined-behaviour-in-c-when-adding-to-map/ for more info.


paroj:
thanks

PR #583, opened 2017-11-26
Docs: TextureManager::createManual() returns existing textures

mimon:


paroj:
sounds more like a bug

mimon:
Oh, ok. I just assumed it wasn't because the "name" param for ResourceManager::createResource() says it has to be unique.

Also, I'm using Metal on macOs.

paroj:
lets rather track this as an issue

PR #584, opened 2017-11-26
safely remove frame listener

bracoduvnjak:
Since all program events are usually handled on FrameStarted, ...ended, etc events, removing frame listener might cause ogre to crash.
For example, let in frameListener set have listeners (classes) A,B,C...
then let say, when listener A is called, listener A deletes class B (listener B). The existing loop will crash ogre since listener B will be called next, but it is already deleted...
I suggest to modify _fireFrameStarted (and similar functions like _fireFrameEnded etc) to take into consideration above described case.

Perhaps this (or something similar) can help:

```cpp
bool Root::_fireFrameStarted(FrameEvent& evt)
{
OgreProfileBeginGroup("Frame", OGREPROF_GENERAL);
_syncAddedRemovedFrameListeners();

// Tell all listeners
for (set<FrameListener*>::type::iterator i = mFrameListeners.begin(); i != mFrameListeners.end(); ++i)
{
//Add this modification
if(mRemovedFrameListeners.size())
{
set<FrameListener*>::type::iterator it = mRemovedFrameListeners.find(*i);
if(it != mRemovedFrameListeners.end())
continue;
}
//end of added block

if (!(*i)->frameStarted(evt))
return false;
}

return true;
}
```


paroj:
you are trying to merge master into v1-9 here. You should instead merge the master (or any other branch of your fork) into OGRECave/ogre master.
Furthermore It seems like you did not add any commits here.

paroj:
also since we already have the delayed listener addition/ deletion in place I would rather extend the documentation to state that listener deletion is only allowed after renderOneFrame returns.

bracoduvnjak:
Sorry about wrong pull request master -> 1.9 (I'm not familiar how to make one correctly)
However, I think that suggested modification is valuable:
1. When application is developed to use renderOneFrame, your suggestion make sense, but it require developers extra work (time) to maintain separate containers of objects that need to be removed. Moreover, applications that use startRendering (like mine) as per your suggestion, would need separate loop (which manages when to remove objects). That loop is very hard and unnecessary to implement. On which event would removal function be called???
2. People might miss documentation specification (deletion only after renderOneFrame returns)
which again creates unnecessary frustration and lost of time.
3. Cost for extra checking for deleted objects is negligible.
Please reconsider.

paroj:
as this PR does not contain any commits, lets continue at #590

PR #587, opened 2017-11-27
OgreTerrainGroup thread safety regarding Terrain creation

SNiLD:
TerrainGroup will use Terrain instance inside thread but allows destroying it while it's being processed in the thread.

### DIFF
--- a/Components/Terrain/src/OgreTerrainGroup.cpp
+++ b/Components/Terrain/src/OgreTerrainGroup.cpp
@@ -709,6 +709,7 @@ namespace Ogre
         if (it == mTerrainPrepareRequests.end())
         {
             freeTerrainSlotInstance(lreq.slot);
+            return;
         }
         else
         {
--- a/Components/Terrain/src/OgreTerrainGroup.cpp
+++ b/Components/Terrain/src/OgreTerrainGroup.cpp
@@ -38,7 +38,6 @@ namespace Ogre
     const uint16 TerrainGroup::WORKQUEUE_LOAD_REQUEST = 1;
     const uint32 TerrainGroup::CHUNK_ID = StreamSerialiser::makeIdentifier("TERG");
     const uint16 TerrainGroup::CHUNK_VERSION = 1;
-    uint TerrainGroup::LoadRequest::loadingTaskNum = 0;
 
     //---------------------------------------------------------------------
     TerrainGroup::TerrainGroup(SceneManager* sm, Terrain::Alignment align, 
@@ -100,7 +99,7 @@ namespace Ogre
         }
 
         // waiting for terrain preparing finished
-        while(LoadRequest::loadingTaskNum>0)
+        while (getNumTerrainPrepareRequests() > 0)
         {
             OGRE_THREAD_SLEEP(50);
             Root::getSingleton().getWorkQueue()->processResponses();
@@ -309,12 +308,11 @@ namespace Ogre
             LoadRequest req;
             req.slot = slot;
             req.origin = this;
-            ++LoadRequest::loadingTaskNum;
-            mTerrainLoadRequests.insert(TerrainLoadRequestMap::value_type(slot->instance, false));
-            Root::getSingleton().getWorkQueue()->addRequest(
-                mWorkQueueChannel, WORKQUEUE_LOAD_REQUEST,
-                Any(req), 0, synchronous);
-
+            WorkQueue::RequestID id =
+                Root::getSingleton().getWorkQueue()->addRequest(
+                    mWorkQueueChannel, WORKQUEUE_LOAD_REQUEST,
+                    Any(req), 0, synchronous);
+            mTerrainPrepareRequests.insert(TerrainPrepareRequestMap::value_type(slot, id));
         }
     }
     //---------------------------------------------------------------------
@@ -364,6 +362,11 @@ namespace Ogre
                 mAutoUpdateLod->autoUpdateLod(slot->instance, synchronous, data);
             }
         }
+    }
+    //---------------------------------------------------------------------
+    size_t TerrainGroup::getNumTerrainPrepareRequests() const
+    {
+        return mTerrainPrepareRequests.size();
     }
     //---------------------------------------------------------------------
     void TerrainGroup::unloadTerrain(long x, long y)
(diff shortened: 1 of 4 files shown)


SNiLD:
This change also kind of makes `LoadRequest::loadingTaskNum` redundant but I don't want to remove it so that backwards compatibility does not break.

paroj:
I am not very familiar with the Terrain component API. Is `TerrainGroup` supposed to be inherited from? If not you can remove `LoadRequest::loadingTaskNum` as it is protected.

Components/Terrain/src/OgreTerrainGroup.cpp
```diff
@@ -825,6 +837,26 @@ namespace Ogre
...
+        if (!slot)
+            return;
+
+        TerrainLoadRequestMap::iterator it = mTerrainLoadRequests.find(slot->instance);
+
+        // Terrain was in load request so we need to schedule the deletion see handleResponse().
+        if (it != mTerrainLoadRequests.end())
+        {
+            // We would like to abort the request here but WorkQueue does not provide proper tools for this
+            // (the data we need will be destroyed if we call abortRequest()).
```
paroj: which data are you referring to here? As far as I can see `abortRequest()` will only delete `ResourceRequest::loadParams`

paroj: You are probably referring to the `TerrainSlot*`. Would calling abortRequest() and immediately deleting it here work?

SNiLD: You're correct that the problem is `TerrainSlot*`. It can't be deleted however since `Terrain*` would also be deleted and that is used in the thread. If `WorkQueue` provided a method to abort request if it was not being processed right now it would be fine. That way we would have tools to abort all other requests except the one that's currently being processed.

SNiLD:
I think @sinbad meant `TerrainGroup` to be inherited from. Many of the important functions like `define*()` are virtual.

Components/Terrain/src/OgreTerrainGroup.cpp
SNiLD: `WorkQueue` has function `abortPendingRequestsByChannel(uint16 channel)` so if there was function like  `abortPendingRequest(RequestID id)` as well it could be done.

paroj:
> I think @sinbad meant TerrainGroup to be inherited from. Many of the important functions like define*() are virtual.


I would not assume that this is intentional. At some point he just made everything virtual (Java style).
A quick look at the samples and Ogitor showed that neither are inheriting from TerrainGroup. I assume you do neither - so feel free to consider everything protected to be actually private.

Components/Terrain/src/OgreTerrainGroup.cpp
paroj: > if there was function like abortPendingRequest(RequestID id) as well it could be done.

this will probably be useful for others as well. So I would prefer this solution over the currently proposed workaround.

SNiLD: I'll make this change but currently I'm a bit busy so it might take a few days.

SNiLD:
This should work now.

Disappointingly I found out that calling `Terrain::prepare()` from background thread is not safe anyways since another terrain might try to access these resources (height data and quadtree) when they're neighbours. Fix to make `Terrain::prepare()` thread safe should be a separate issue though.

SNiLD:
Actually there could be at least semi-remedy to this if TerrainGroup had `disconnectNeighbour()` as well like `connectNeighbour()`. I'll try this as well.

Components/Terrain/src/OgreTerrainGroup.cpp
```diff
@@ -693,25 +696,23 @@ namespace Ogre
...
-            bool isRemoved = it->second;
-            mTerrainLoadRequests.erase(it);
-
-            // Instance was scheduled for removal while it was processed in another thread.
-            if (isRemoved)
-            {
-                freeTerrainSlotInstance(lreq.slot);
-                return;
-            }
+            freeTerrainSlotInstance(lreq.slot);
```
SNiLD: There should be return; after the free here. I'll try if this fixes the neighbour issue.

SNiLD: Ok that did not fix the issue, unfortunately.

SNiLD:
`TerrainGroup::disconnectNeighbour()` does not make sense since `Terrain` will remove itself from neighbours when it's destroyed. The only case where it would be needed is when terrain instance prepare is aborted and is currently being processed in TerrainGroup. However once that function returns it will check for `mTerrainPrepareRequests` and free the slot if it was not found so it is never even connected.

Making `Terrain` thread safe regarding the neighbour access should be separate issue as stated before.

paroj:
:+1:  from my side. Is this ready to merge?

SNiLD:
Yes this is ready.

paroj:
after this merge the "Terrain" sample crashes. Can you reproduce?

SNiLD:
Yes, it crashes for me too. Instances taken from `mTerrainGroup->getTerrainIterator()` seem to be null. Maybe I've missed something. Let me investigate.

Components/Terrain/src/OgreTerrainGroup.cpp
```diff
@@ -309,11 +308,11 @@ namespace Ogre
...
-            ++LoadRequest::loadingTaskNum;
-            Root::getSingleton().getWorkQueue()->addRequest(
-                mWorkQueueChannel, WORKQUEUE_LOAD_REQUEST, 
-                Any(req), 0, synchronous);
-
+            WorkQueue::RequestID id =
+                Root::getSingleton().getWorkQueue()->addRequest(
+                    mWorkQueueChannel, WORKQUEUE_LOAD_REQUEST,
+                    Any(req), 0, synchronous);
+            mTerrainPrepareRequests.insert(TerrainPrepareRequestMap::value_type(slot, id));
```
SNiLD: This is the problem line. When synchronous is true it won't go into background thread and there won't be anything in `mTerrainPrepareRequests` so it will be deleted.

paroj:
inserting in mTerrainPrepareRequests *after* doing a synchronous request is too late. But even when fixing this, the sample crashes on shutdown. Please investigate.

Components/Terrain/src/OgreTerrainGroup.cpp
paroj: local fix:
```cpp
     TerrainPrepareRequestMap::iterator iter =
                mTerrainPrepareRequests.insert(TerrainPrepareRequestMap::value_type(slot, 0)).first;
            iter->second = Root::getSingleton().getWorkQueue()->addRequest(
                mWorkQueueChannel, WORKQUEUE_LOAD_REQUEST, Any(req), 0, synchronous);
```

SNiLD: You can't insert to that iterator at that point anymore since it will be invalid at that point. This is how I'd fix it:
```
            std::pair<TerrainPrepareRequestMap::iterator, bool> ret = mTerrainPrepareRequests.insert(TerrainPrepareRequestMap::value_type(slot, 0));
            assert(ret.second == false);
            WorkQueue::RequestID id =
                Root::getSingleton().getWorkQueue()->addRequest(
                    mWorkQueueChannel, WORKQUEUE_LOAD_REQUEST,
                    Any(req), 0, synchronous);
            if (!synchronous)
                ret.first->second = id;
```

SNiLD: Ok this fix stops `Terrain` sample from crashing at start.

SNiLD:
What is the proper way to merge that change as well? I pushed it to the same branch that you've already merged.

Components/Terrain/src/OgreTerrainGroup.cpp
paroj: ah.. yes.. I figured. seems to be properly fixed this way

paroj:
you have to create a new pull-request using the same branch.

PR #597, opened 2017-12-07
Android Build SampleBrowser library using CMake instead of android makefiles

kPYKx7Ddw4n1aIKZ:
This pull request is related to https://ogre3d.atlassian.net/browse/OGRE-568.
It adds the option `OGRE_BUILD_ANDROID_JNI_LIBS_CMAKE` which enables the building of the SampleBrowser library on android using CMake instead of android makefiles.

However this does not completely solve the referenced issue.

I tested this on a physical android device, where it worked.
Besides, it was the only way to build OGRE using the newest NDK with the Clang compiler.

### DIFF
--- a/CMake/Utils/AndroidMacros.cmake
+++ b/CMake/Utils/AndroidMacros.cmake
@@ -8,23 +8,25 @@
 #-------------------------------------------------------------------
 
 
-if(NOT TARGET cpufeatures)
-    add_library(cpufeatures STATIC ${ANDROID_NDK}/sources/android/cpufeatures/cpu-features.c)
-    target_link_libraries(cpufeatures dl)
-endif(NOT TARGET cpufeatures)
+if(OGRE_BUILD_ANDROID_JNI_LIBS_CMAKE)
+    if(NOT TARGET cpufeatures)
+        add_library(cpufeatures STATIC ${ANDROID_NDK}/sources/android/cpufeatures/cpu-features.c)
+        target_link_libraries(cpufeatures dl)
+    endif(NOT TARGET cpufeatures)
 
-if(NOT TARGET native_app_glue)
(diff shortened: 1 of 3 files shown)


CMake/Utils/AndroidMacros.cmake
```diff
@@ -8,6 +8,26 @@
 #-------------------------------------------------------------------
 
 
+if(OGRE_BUILD_ANDROID_JNI_LIBS_CMAKE)
+    if(NOT TARGET cpufeatures)
+        add_library(cpufeatures STATIC ${ANDROID_NDK}/sources/android/cpufeatures/cpu-features.c)
+        target_link_libraries(cpufeatures dl)
+    endif(NOT TARGET cpufeatures)
+
+    if(NOT TARGET native_app_glue)
```
paroj: the condition when the native_app_glue is built is hard to follow, please rather copy the according parts from here:
https://github.com/taka-no-me/android-cmake/blob/master/AndroidNdkModules.cmake

CMakeLists.txt
```diff
@@ -269,6 +269,7 @@ elseif (ANDROID)
   set(OGRE_PLATFORM OGRE_PLATFORM_ANDROID)
   set(OGRE_CONFIG_ENABLE_VIEWPORT_ORIENTATIONMODE FALSE CACHE BOOL "Disable viewport orientation Android" FORCE)
   option(OGRE_BUILD_ANDROID_JNI_SAMPLE "Enable JNI Sample" FALSE)
+  option(OGRE_BUILD_ANDROID_JNI_LIBS_CMAKE "Build native android JNI libs using CMake instead of android makefiles [BETA]" FALSE)
```
paroj: no need to make it optional. We do not need two ways to compile the browser.

paroj:
:+1: looks good for me. Just need to wait for #595 to get rid of the _d suffix in debug builds.

PR #602, opened 2017-12-11
GL3 plus shadow buffer fix

digimatic:
Propagate the parameter correctly to the base class. 
I don't know why it was enforce to off?
Seems to work now anyhow for our use case.


AllianceDarkSylinc:
This is a sort of "don't touch, don't know" situation.

The case of GL3+ is that this is some legacy code carried over from 1.x.
This setting was mostly because of device resets (although there is a second useful reason: direct CPU access without requiring to lock).

Unlike D3D9 there are no device resets, so it is very likely the reason it was forced off. Although if may just have been an oversight years ago. By turning it on you're just doubling memory usage since the GL driver has to keep a copy too.
The paths using the shadow copy are not as well tested as the path without it.

And because it's legacy, I just didn't devote time to it.

Metal is a different thing, since I wrote it from scratch. The v1 HardwareVertexBuffer & family was very difficult to write for Metal because Metal has explicit syncs, and no notion of buffer discards. HardwareVertexBuffer uses buffers with a 2000-era mentality on Metal, a very modern API.

I had to write a class that would emulate buffer renaming (basically what the D3D9 & D3D11 drivers do behind our backs) and work as fast as possible, while keeping everything in sync. And fall off to very slow paths (stall everything) if we cannot reliably keep all copies in sync.
Having MetalHardwareVertexBuffer & family work fine with this code was difficult, and having to also worry about the CPU shadow copy being kept in sync made things worse.

I wrote the Metal code with shadow buffer copies in mind. However due to time contraints I could not test it at all. Testing would have to be thorough.
Because Metal was first written for iOS, shadow buffer didn't make sense anyway: the CPU and GPU share the same RAM (thus no sync problems or transfer worries) and doubling memory RAM is prohibitively expensive in that platform.
This is not true for macOS though.

So... there it is. Why it's forced false on Metal, and why we haven't worried about GL3+.

paroj:
on 1.10 applying this fixes bboxes for hardware skeletal animation. But maybe it just masks some other bug this way..

paroj:
no, one should generally respect the shadowBuffer setting:
https://github.com/OGRECave/ogre/blob/2d97fa67340911f309a9edd213a09677c54bd6ef/OgreMain/src/OgreMesh.cpp#L1013-L1018

AllianceDarkSylinc:
I think I know what that "masked bug" could be.

You can see in:
https://bitbucket.org/sinbad/ogre/src/23a6fd3844fa0b3319c4d6637ff1fbceccf40568/RenderSystems/GL3Plus/src/OgreGL3PlusHardwareBuffer.cpp?at=default&fileviewer=file-view-default#OgreGL3PlusHardwareBuffer.cpp-69

That when a buffer was created mUsage & HardwareBuffer::HBU_WRITE_ONLY, we will never set the GL_MAP_READ_BIT.
It is quite possible the buffer was created with HBU_WRITE_ONLY and then as LockOption, HBL_READ_ONLY was passed. This is an invalid combination.

In older drivers this would work because the drivers would return an area of memory that happened to have the contents we were seeking. But newer drivers return a different region, so reading from there will result in garbage.

In Ogre 2.1 we assert when this happens; and our recommendation is to create the buffer without the HBU_WRITE_ONLY flag. Since you can't map for reading something you promised was write-only.

AllianceDarkSylinc:
We replied at roughly the same time

AllianceDarkSylinc:
My biggest concern is what is the default? Default should be to not keep shadow copies.

paroj:
> My biggest concern is what is the default? Default should be to not keep shadow copies.

yeah.. you guessed corretly: default is creating shadow copies..

paroj:
> In Ogre 2.1 we assert when this happens; and our recommendation is to create the buffer without the HBU_WRITE_ONLY flag. Since you can't map for reading something you promised was write-only.

this is wrong though. See glBufferData spec:

> usage is a hint to the GL implementation as to how a buffer object's data store will be accessed. This enables the GL implementation to make more intelligent decisions that may significantly impact buffer object performance. **It does not, however, constrain the actual usage of the data store.**

darksylinc:
Yeah I know, but this can force GL drivers on a slow path (GPU can no longer assume it's private memory).

Lock a mesh even once for reading -> potentially get stuck with a slow mesh forever.

Better make this explicit.

paroj:
closing due to conflicts and as this probably should go via bitbucket

PR #622, opened 2018-01-05
[OGRE-592] RTShaderSystem compile error when using double precision

niparx:
https://ogre3d.atlassian.net/browse/OGRE-592

### DIFF
--- a/Components/RTShaderSystem/src/OgreShaderFFPTexturing.cpp
+++ b/Components/RTShaderSystem/src/OgreShaderFFPTexturing.cpp
@@ -655,7 +655,11 @@ void FFPTexturing::addPSArgumentInvocations(Function* psMain,
         }
         else
         {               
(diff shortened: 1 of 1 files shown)


Components/RTShaderSystem/src/OgreShaderFFPTexturing.cpp
```diff
@@ -655,7 +655,11 @@ void FFPTexturing::addPSArgumentInvocations(Function* psMain,
         }
         else
         {               
-            curFuncInvocation->pushOperand(ParameterFactory::createConstParam(Vector4(colourValue.ptr())), Operand::OPS_IN);
+			curFuncInvocation->pushOperand(ParameterFactory::createConstParam(Vector4((Real)colourValue.r,
```
paroj: please change all tabs to 4 spaces.

PR #641, opened 2018-01-24
Fixes assert in PBS setTexture to allow setting null texture

digimatic:
Fixes the assert to allow setting PbsDatablock texture to the null texture.


digimatic:
Off-topic question: Is it ok to submit MR thru this gihub mirror?

darksylinc:
Hi! Thanks!

Normally paroj uses github for 1.x, I use bitbucket for 2.x; I should ask paroj what's the procedure for bringing commits from github to bitbucket mirrors

Since this is a minor bug, I just fixed it myself. I forgot about it until 5 minutes ago when I was struck by this bug and then D'oh! I remembered about this ticket. Sorry for the delay.

Thanks for the report!

darksylinc:
(Fixed in https://bitbucket.org/sinbad/ogre/commits/c0f0c7ad1ee3db853faef329e92034befb8e124b )

PR #656, opened 2018-02-03
[OSX] Fix crash when embedding OgreOSXCocoaWindow in external window

### LINKED ISSUE #653: When used in combination with QWidget::createWindowContainer(), CocoaWindow::windowMovedOrResized() crashes
Environment
```
OS: macOS 10.13.3
Xcode 9.2 clang-900.0.39.2
OGRE: 1.10.11
Qt: 5.10
````

Hi

I have created a QWindow - based OgreWindow to use OGRE with Qt.
https://github.com/buresu/OgreMockup/blob/master/src/window/OgreWindow.hpp
https://github.com/buresu/OgreMockup/blob/master/src/window/OgreWindow.cpp

It works well on a standalone basis.

But, I embeded it in QWidget, it crashes with CocoaWindow::windowMovedOrResized() when resizing the window.

The code of embedding in QWidget
```
QWidget *widget = new QWidget();

QHBoxLayout *layout = new QHBoxLayout(widget);

OgreWindow *view = new OgreWindow;
QWidget *container = QWidget::createWindowContainer(view, widget);

layout->addWidget(container);
widget->setLayout(layout);

widget->show();
```

This problem did not occur on Windows.

This problem does not always occur.
If it succeeds at application startup, it does not crash afterwards.

I don't know if there is a relationship, but occasionally there is output like the following when crashing.
```
[__NSMallocBlock__ frame]: unrecognized selector sent to instance 0x7f9b2d522470
```

Is there any solution?

Thanks

buresu:
fix #653

### DIFF
--- a/RenderSystems/GLSupport/src/OSX/OgreOSXCocoaWindow.mm
+++ b/RenderSystems/GLSupport/src/OSX/OgreOSXCocoaWindow.mm
@@ -505,12 +505,16 @@ namespace Ogre {
         if(mIsFullScreen)
             return;
 
-		NSRect frame = [mWindow frame];
+        NSRect frame = mIsExternal ? [[mView window] frame] : [mWindow frame];
         NSRect screenFrame = [[NSScreen mainScreen] visibleFrame];
 		frame.origin.x = leftPt;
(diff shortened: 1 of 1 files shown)


paroj:
thanks

PR #675, opened 2018-02-24
WIP: TextureUnitState performance optimization by avoiding vector…

TheNicker:


OgreMain/src/OgreTextureUnitState.cpp
```diff
@@ -292,7 +302,7 @@ namespace Ogre {
         if (ct == CONTENT_SHADOW || ct == CONTENT_COMPOSITOR)
         {
             // Clear out texture frames, not applicable
-            mFrames.clear();
+            setNumFrames(0);
```
paroj: even here `mFrames.size() == mNumFrames`, so I still think we can drop `mNumFrames`

paroj: I am trying to just remove mFrames, as it duplicates information with mFramePtrs.

paroj: needs more thought, see #677

paroj:
no reply from author

PR #676, opened 2018-02-25
Changed: Direct3D9 performance optimizations.

TheNicker:


paroj:
for reference: try to be more descriptive in the commit message. e.g. it is evident that a commit changes stuff, so drop "Changed".

More info: https://chris.beams.io/posts/git-commit/

PR #682, opened 2018-02-28
Fixes metal pvrtc support

digimatic:
This PR do essentially three things (aiming to get pvrtc to work on Metal/iOS):
 - Adds support for PVRTC pixel format in the KTX-loader (named ETCCodec). KTX is a general container (similar to DDS) and it can be useful to be able to use that not only on Android.
 - Adds support for loading KTX-texture with mip maps.
 - Loosens the PixelFormat functions to be able to work on at least a full slice when pixel format is compressed. Otherwise Metal renderer fails to load pvrtc compressed textures.

### DIFF
--- a/OgreMain/src/OgreETCCodec.cpp
+++ b/OgreMain/src/OgreETCCodec.cpp
@@ -328,6 +328,18 @@ namespace Ogre {
         case 33779: // DXT 5
             imgData->format = PF_DXT5;
             break;
+         case 0x8c00: // COMPRESSED_RGB_PVRTC_4BPPV1_IMG
+            imgData->format = PF_PVRTC_RGB4;
+            break;
+        case 0x8c01: // COMPRESSED_RGB_PVRTC_2BPPV1_IMG
+            imgData->format = PF_PVRTC_RGB2;
+            break;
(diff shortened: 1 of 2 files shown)


paroj:
why did you close this? the changes seem sensible

digimatic:
Sorry, I was debugging a problem and thought it was in this patch , but wasn't. Reopening.

The two additions in ETCCodec I have had for many years in Ogre 1.9/1.10 so it shall work. I can understand if you don't want to expand with other pixel format (like pvrtc), but in my opinion it shall be expanded with a lot more formats like float rgb and more. If go for that I think perhaps the file shall be renamed OgreKTXCodec instead (similar to OgreDDSCodec).

AllianceDarkSylinc:
1. The PR looks OK
1. Holy ####! The ETC codec is a KTX codec? I never truly paid attention to it. I thought it just loaded some obscure android ETC compression format. It should certainly be renamed.

How do we do this Pavel? Accept the PR and I copy the patch to bitbucket? You manually merge it to bitbucket?

paroj:
I will merge this after I sync v2-1 with bitbucket, then things will work semi-automatically

digimatic:
I am a little incorrect. Actually ETCCodec is a class that registers both a .KTX and PKM codec to the system.

PR #693, opened 2018-03-15
Take scale into account when culling objects.

erikogenvik:
This fixes a bug where entities that were scaled would get incorrectly
culled from the scene. Since getMaxScaleCoef() is called more
frequently the mMaxScaleLocal is used as cache when there's a parent
scene node.

To see the bug, on could create a scene with a single instanced entity. Change the scale of the entity by some magnitude and move the camera around. As soon as the unscaled version of the instanced entity would be outside of the frustrum the entity will be hidden.


paroj:
thanks

PR #694, opened 2018-03-15
CMake: fix static build problem

buresu:
Fixed a problem where `FindFreetype.cmake` could not be found during installation.


paroj:
thanks

buresu:
thank you :)

PR #697, opened 2018-03-19
Add warning for improper overlay declaration

### LINKED ISSUE #696: No warning for '{' in overlay name definition
This might be a beginner's mistake, but when writing my own application, I came across the following:
When defining an overlay in a .overlay script such as 
```
RandomOverlay {
...
```
no warnings are shown. However, the name of the overlay will, of course, be set to `RandomOverlay {`.

This is mentioned slightly in the [ogre manual](https://ogrecave.github.io/ogre/api/1.10/Overlay-Scripts.html):
> Every overlay in the script must be given a name, which is the line before the first opening ’{’.

Of course, the correct behaviour would be achieved with
```
RandomOverlay
{
...
```
but there is, at parsing time, no indication that anything is wrong.

marko213:
Make overlay definition and opening (with '{') on the same line work.
Resolves #696.

### DIFF
--- a/Components/Overlay/src/OgreOverlayManager.cpp
+++ b/Components/Overlay/src/OgreOverlayManager.cpp
@@ -243,10 +243,18 @@ namespace Ogre {
                             // chop off the 'particle_system ' needed by new compilers
                             line = line.substr(8);
                         }
+                        if (line[line.length() - 1] == '{') {
+                            // Open the overlay on the same line
+                            int i;
+                            for (i = line.length() - 2;
+                                line[i] == ' ' || line[i] == '\t'; --i);
+                            line = line.substr(0, i + 1);
+                        } else {
+                            // Skip to and over next {
+                            skipToNextOpenBrace(stream);
+                        }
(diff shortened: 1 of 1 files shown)


Components/Overlay/src/OgreOverlayManager.cpp
```diff
@@ -237,6 +237,12 @@ namespace Ogre {
                     }
                     else
                     {
+                        // Check that the { is not on the same line
+		        if (line[line.length() - 1] == '{')
+                        {
+                            LogManager::getSingleton().logWarning(
```
paroj: looking at the following code, I think we can just fix the issue instead of warning about it here.
You just need to strip "{" from the name and skip the `skipToNextOpenBrace` call.

marko213: Assuming that no one actually wants such a name, yes. Also, the possible whitespace between the name and '{' should probably be removed, if there is any (unless that is already handled in Overlay creation).

paroj:
thanks

PR #704, opened 2018-03-21
Add errors for missing braces in script files.

### LINKED ISSUE #701: Parsing script files with missing end bracket(s) fails silently
When a script file (tested with .material so far, but they should have the same parser) with at least one missing end bracket ('}') is parsed, no errors are generated; however the corresponding script is not added to the list of scripts.

Example script:
```cpp
// PanelWhite.material

material PanelWhite {
    receive_shadows off
    technique {
        pass PanelWhite {
            ambient 0.9 0.9 0.9 1.0
            diffuse 1.0 1.0 1.0 1.0
//	}
    }
}
```

marko213:
fixes #701

### DIFF
--- a/OgreMain/src/OgreScriptLexer.cpp
+++ b/OgreMain/src/OgreScriptLexer.cpp
@@ -39,7 +39,7 @@ namespace Ogre{
         char c = 0, lastc = 0;
 
         String lexeme;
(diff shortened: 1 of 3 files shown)


paroj:
thanks

Issue #668, opened 2018-02-11
API defects


paroj:
tracking of public API defects that are to widespread to be fixes at this point
* `Renderable::getWorldTransforms`: uses Matrix4 instead of Affine3
* `Renderable::preRender`: is non const, but is called from a const codepath

PR #708, opened 2018-03-25
Add interface for adding scene loaders

### LINKED ISSUE #566: Introduce API for scene loading
something where DotSceneLoader and x3ogre could register. Should first be tested in Bites.

similar to Codec API:
```cpp
SceneLoader::getLoader("scene")->load("example.scene", "General", scnMgr);
```

marko213:
Add an API / interface for external scene loaders as suggested in #566.

This might need some documentation.

fixes #566


marko213:
Quick and rough test without actual file loading:
[SceneLoaders.zip](https://github.com/OGRECave/ogre/files/1845572/SceneLoaders.zip)

OgreMain/include/OgreSceneLoader.h
```diff
@@ -0,0 +1,18 @@
...
+#include "OgreSceneManager.h"
+
+namespace Ogre {
+    
+    class _OgreExport SceneLoader
+    {
+    public:
+        virtual ~SceneLoader() {};
+        virtual bool load(const String& filename, const String& groupname, SceneNode *rn);
+        virtual bool load(const String& filename, const String& groupname, SceneManager *sm);
```
paroj: SceneNodes know their creator, so this one is not needed

paroj: Also I would prefer exceptions for error handling here

paroj: This function should be pure virtual as not overriding it makes no sense. (making this class header only)

paroj: should take DataStream instead of a filename for flexibility. See `ScriptLoader::parseScript`. Also adapt the comments from there.

OgreMain/include/OgreSceneLoaderManager.h
```diff
@@ -0,0 +1,26 @@
...
+#include "OgreSceneLoader.h"
+
+namespace Ogre {
+    
+    class _OgreExport SceneLoaderManager : public Singleton<SceneLoaderManager>
+    {
+    public:
+        SceneLoaderManager();
+        virtual ~SceneLoaderManager();
+        void registerSceneLoader(const String& name, SceneLoader *sl);
```
paroj: name is misleading in this context. should be rather "extension"/ "pattern"

OgreMain/include/OgreSceneLoaderManager.h
```diff
@@ -0,0 +1,26 @@
...
+        SceneLoader *getByName(const String& name);
```
paroj: I think it makes more sense to provide a high-level "load" here that picks the right sceneloader or throws otherwise.

marko213: Still, what about checking available loaders at runtime?

marko213: Will create seperate method for checks.

OgreMain/include/OgreSceneLoaderManager.h
```diff
@@ -0,0 +1,50 @@
...
+        /** Load a scene from a SceneLoader
+        @remarks This is usually equivalent to calling
+            SceneLoader::load(stream, groupName, rootNode)
+        @param loaderName The name of the SceneLoader that will be used for loading.
+        @param stream Weak reference to a data stream which is the source of the scene
+        @param groupName The name of a resource group which should be used if any resources
+            are created during the parse of this script.
+        @param rootNode The root node for the scene being created
+        */
+        void load(const String& loaderName, DataStreamPtr& stream, const String& groupName, SceneNode *rootNode);
```
paroj: instead of explicitly requiring a loader name, you can use `stream->getName()` to inspect the file extension and automatically select the loader based on that

marko213: Should there still be a check that all extensions supported by the SceneLoader being added have not yet been registered by another SceneLoader and that the extensions don't repeat in the same SceneLoader? It seems unnecessary now that we don't query by name. (So there can be overlaps and such, but the first one registered gets priority.)

paroj: yes, people might register common extensions like "xml" that could map to different incompatible file formats.

marko213: Will warn unless only one format is specified.

paroj: adding an overload `void load(const String& filename, ..` that uses `Root::openFileStream` internally would be convenient.

OgreMain/include/OgreSceneLoaderManager.h
```diff
@@ -0,0 +1,65 @@
...
+    public:
+        SceneLoaderManager();
+        virtual ~SceneLoaderManager();
+        /** Register a new SceneLoader
+            @param name The name for the SceneLoader for lookup.
+            @param ext A StringVector containing the supported extensions for the
+                SceneLoader being registered.
+            @param sl Pointer to the SceneLoader instance.
+        */
+        void registerSceneLoader(const String& name, StringVectorPtr ext, SceneLoader *sl);
```
paroj: please use `const StringVector&` instead of StringVectorPtr. See #668.

Otherwise the API looks good to me :+1:

OgreMain/include/OgreSceneLoaderManager.h
```diff
@@ -0,0 +1,65 @@
...
+    protected:
+        /// Struct for storing data about SceneLoaders internally
+        struct SceneLoaderInfo
+        {
+            SceneLoaderInfo(SceneLoader *l, StringVectorPtr ext);
+            SceneLoader* loader;
+            StringVectorPtr supportedExt;
+        };
+        /// Map from scene loader names to SceneLoaderInfo
+        typedef map<String, SceneLoaderInfo>::type SceneLoaderMap;
```
paroj: just std::map with latest master

paroj:
feel free to remove the WIP in the title in case you are done

PR #711, opened 2018-03-26
GLSupport: fix layout qualifier parsing

fbridault:
This brings two fixes for `extractLayoutQualifiers()`.

## 1. Handle case with interpolation qualifier

When using both layout location and interpolation qualifiers, I use to write it this way:
```
flat layout(location = 1) out vec4 outColor;
```
However I discovered that this syntax is not supported by the MacOs GLSL compiler. It is hard to find an official documentation to know what is the correct syntax (nothing [here](https://www.khronos.org/opengl/wiki/Layout_Qualifier_(GLSL)) for instance), but I found that this way works on all the platforms I tested so far (Linux, Windows, MacOs):
```
layout(location = 1) flat out vec4 outColor;
```
However, `GLSLProgramCommon::extractLayoutQualifiers(void)` fails to parse this correctly. I propose so to detect the case when a interpolation qualifiers is used and skip it appropriately since it does not provide any information we need.

## 2. Fix the increment of curPos when skipping 'out' qualifiers

Working on the previous fix, I discovered that the loop was broken with 'out' qualifiers. The keyword `continue` was used, but without calling the loop increment `currPos = shaderSource.find("layout", currPos);`. So a line with an `out` qualifier was parsed many times before being really skipped, and sometimes this led to raise an assertion.
I propose to avoid the use of continue which led to this issue and invert the `if` condition.

### DIFF
--- a/RenderSystems/GLSupport/src/GLSL/OgreGLSLProgramCommon.cpp
+++ b/RenderSystems/GLSupport/src/GLSL/OgreGLSLProgramCommon.cpp
@@ -108,28 +108,36 @@ void GLSLProgramCommon::extractLayoutQualifiers(void)
             // It should contain 3 parts, i.e. "attribute vec4 vertex".
             break;
         }
-        if (parts[0] == "out")
+        size_t attrStart = 0;
+        if (parts.size() == 4)
         {
-            // This is an output attribute, skip it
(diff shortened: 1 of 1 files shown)


paroj:
thanks

PR #718, opened 2018-03-31
use std::unique_ptr get member function for null pointer check

uvguy:
Fixes https://github.com/OGRECave/ogre-procedural/issues/148

### DIFF
--- a/OgreMain/include/OgrePass.h
+++ b/OgreMain/include/OgrePass.h
@@ -274,25 +274,25 @@ namespace Ogre {
                                                  mTessellationHullProgramUsage || mTessellationDomainProgramUsage || mComputeProgramUsage; }
 
         /// Returns true if this pass uses a programmable vertex pipeline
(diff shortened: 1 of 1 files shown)


paroj:
have you tried -std=c++11 already? The issue is not the ogre code, but that NULL != nullptr

OgreMain/include/OgrePass.h
```diff
@@ -274,25 +274,25 @@ namespace Ogre {
                                                  mTessellationHullProgramUsage || mTessellationDomainProgramUsage || mComputeProgramUsage; }
 
         /// Returns true if this pass uses a programmable vertex pipeline
-        bool hasVertexProgram(void) const { return mVertexProgramUsage != NULL; }
+        bool hasVertexProgram(void) const { return mVertexProgramUsage.get() != nullptr; }
```
paroj: please investigate why NULL != nullptr for you

Issue #726, opened 2018-04-09
revise RenderWindow::setFullscreen API


paroj:
currently it causes conflicts with external windows and it lacks an EGL/ wayland implementation.
Probably just deprecate it.

PR #727, opened 2018-04-09
GLSupport: do not attempt to set external windows full-screen

### LINKED ISSUE #729: TextureUnitState::ensurePrepared/ensureLoaded bug
#### System Information
- Ogre Version:: Master
- Operating System / Platform:  windows
- RenderSystem:  OpenGL

#### Detailled description
TextureUnitState::ensurePrepared set back to the texture ptr the formats and mTextureSrcMipmaps.
But mTextureSrcMipmaps is int and can be -1 since void setNumMipmaps(uint32 num) is unsigned int.
This result as a very big number in mipmaps and crash.

In my case the problem is shown when I use manual textures and apply it after on the material texture unit. Also my original texture ptr is overwritten by the unset texture unit members.
        tex->setTextureType(mTextureType);
        tex->setFormat(mDesiredFormat);
        tex->setGamma(mGamma);
        tex->setNumMipmaps(mTextureSrcMipmaps);
        tex->setHardwareGammaEnabled(mHwGamma);
        tex->setTreatLuminanceAsAlpha(mIsAlpha);

paroj:
fixes #729


arkeon7:
*GLSupport: do not attempt to set external windows full-screen
Any reason to that ? I was using it with my external window ^^

paroj:
this 
https://forums.ogre3d.org/viewtopic.php?f=2&t=94178#p541354

and this: https://github.com/paroj/ogre/blob/3b7e5eee9eb4033ba7f56851478a746b254249bd/RenderSystems/GLSupport/src/win32/OgreWin32Window.cpp#L258

does `Win32Window::setFullscreen` works for you in the scenario described above?

paroj:
I think the API should work everywhere or not be provide at all. Currently setFullscreen would only work on windows under certain circumstances (non external window or non SDL external window) - on GLX it does not work with external windows at all.
Of course we could try to fix those issues, but catching up with SDL (or glfw) in this regard is just wasted time IMO. Therefore I would rather deprecate setFullscreen. #726

paroj:
you are right, so there still should an API to signal that a full-screen change happened.

PR #761, opened 2018-05-14
Update freetype download website.

luchu1993:
Update freetype download website.


paroj:
why?

luchu1993:
It may be a problem with my network environment.. When I use http, I can't download freetype. Instead, I can use https to download it.  :)

PR #798, opened 2018-06-27
GL3Plus: report glLinkProgram failures

fbridault:
This helps a lot to debug GLSL program link failures.

### DIFF
--- a/RenderSystems/GL3Plus/src/GLSL/OgreGLSLSeparableProgram.cpp
+++ b/RenderSystems/GL3Plus/src/GLSL/OgreGLSLSeparableProgram.cpp
@@ -183,6 +183,11 @@ namespace Ogre
                     OGRE_CHECK_GL_ERROR(glLinkProgram(programHandle));
                     OGRE_CHECK_GL_ERROR(glGetProgramiv(programHandle, GL_LINK_STATUS, &linkStatus));
 
+                    if( !linkStatus )
+                    {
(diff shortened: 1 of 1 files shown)


RenderSystems/GL3Plus/src/GLSL/OgreGLSLSeparableProgram.cpp
```diff
@@ -183,6 +183,11 @@ namespace Ogre
                     OGRE_CHECK_GL_ERROR(glLinkProgram(programHandle));
                     OGRE_CHECK_GL_ERROR(glGetProgramiv(programHandle, GL_LINK_STATUS, &linkStatus));
 
+                    if( !linkStatus )
+                    {
+                        logObjectInfo("Error linking program " + programName, programHandle);
```
paroj: how is this different to the log output at
https://github.com/OGRECave/ogre/pull/798/files#diff-db75d5be65077eb20646713be0ca3952R199

fbridault: I was pretty sure I didn't get any log before... but hell yeah, looking at the code this is really similar. And I ran again without this PR and of course I have the log... Sorry I don't how I missed that, maybe because I enabled ENABLE_GL_CHECK and missed it in the flood...
I hope you did not spend too much time on this. Sorry I'll close the PR.

PR #808, opened 2018-07-10
OgreSceneManager: add missing include for compilation with ninja

PamplemousseMR:
## What does this MR do?

Hi,
I work on Windows 10 64 bits and I'm using `ninja` as a generator for CMake. 
My compiler is from Visual Studio: `cl` version 19.00.24215.1.
It appears that in the file `OgreSceneManager` l:442, there is a `ManualObject` but the include is missing.
This MR allows fixing that.

## How to test it?

Generate with `ninja` and compile with `cl`.


paroj:
https://github.com/OGRECave/ogre/blob/v1.11.1/OgreMain/include/OgrePrerequisites.h#L76-L78

this is on purpose, you should add the include in your code.

PR #809, opened 2018-07-12
OgreMatrix4.h: add ctor that takes pointer to 16 floats

Ybalrid:
We talked a bit on gitter about this, would you take this one in? (I'm proposing the same change to the v2-1 branch over at bitbucket)

### DIFF
--- a/Tests/OgreMain/src/VectorTests.cpp
+++ b/Tests/OgreMain/src/VectorTests.cpp
@@ -93,7 +93,7 @@ TEST(VectorTests,Vector4Scaler)
     v1 -= 4;
     EXPECT_EQ(v1, Vector4(-6,-6,-6,-6));
 }
-
+//--------------------------------------------------------------------------
 TEST(VectorTests,PerpendicularVector)
 {
     Vector3 v1(1, 0, 0);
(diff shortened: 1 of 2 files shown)


Ybalrid:
for cross reference: https://bitbucket.org/sinbad/ogre/pull-requests/824

paroj:
could you also provide a [test](https://github.com/OGRECave/ogre/blob/master/Tests/OgreMain/src/VectorTests.cpp#L36)?

Ybalrid:
Mh, Well, there are currently no tests for the Matrix classes. Should I create a MatrixTest.cpp file, or add a test in the one you linked (Vector)? 

This ctor just loads in some data, so what would the tests consist of? I load a matrix, then I try to get values from it with `operator[]` and I assert the equality between the numbers? This is basically testing that `memcpy` actually copy memory, but we can do that ^^"

paroj:
You can put the tests into VectorTests.cpp. I will rename it to VectorMath at some point..

>  I load a matrix, then I try to get values from it with operator[] and I assert the equality between the numbers? 

yes

> This is basically testing that memcpy actually copy memory, but we can do that ^^"

additionally this would check that the array is interpreted as column-major.

Ybalrid:
So, I just need to add a TEST() block and this will be automatically picked-up by the testing system in the CI?

Tests/OgreMain/src/VectorTests.cpp
```diff
@@ -108,4 +109,45 @@ TEST(VectorTests,PerpendicularVector)
     EXPECT_EQ((v3 + v1).dotProduct((v3 + v1).perpendicular()), (Real)0);
 }
 
+TEST(VectorTest, Matrix4ArrayLoading)
+{
+    Real arr0[16] = { 1, 0, 0, 0,
+                      0, 1, 0, 0,
+                      0, 0, 1, 0
```
paroj: missing `,`

paroj:
yes

Tests/OgreMain/src/VectorTests.cpp
Ybalrid: and there's a missing "=" later too. Should be fixed.

Ybalrid:
Some good news incoming:

```
[ RUN      ] VectorTest.Matrix4ArrayLoading
[       OK ] VectorTest.Matrix4ArrayLoading (0 ms)
[----------] 1 test from VectorTest (1 ms total)
[----------] 10 tests from ZipArchiveTests
```

Still waiting for the Travis CI to finish 😉

paroj:
:+1:

PR #811, opened 2018-07-13
Main: Add a colour mask setting to material passes.

gKevin:
All render systems already supported colour masks but no option was provided to configure it in material scripts. This PR provides a `colour_mask` setting for passes.

### DIFF
--- a/OgreMain/include/OgrePass.h
+++ b/OgreMain/include/OgrePass.h
@@ -139,7 +139,10 @@ namespace Ogre {
         /// Determines if we should use separate blending operations for color and alpha channels
         bool mSeparateBlendOperation : 1;
         /// Colour buffer settings
-        bool mColourWrite : 1;
+        bool mColourWriteR : 1;
+        bool mColourWriteG : 1;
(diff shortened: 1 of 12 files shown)


OgreMain/include/OgrePass.h
```diff
@@ -168,6 +168,9 @@ namespace Ogre {
         bool mPointAttenuationEnabled : 1;
         mutable bool mContentTypeLookupBuilt : 1;
 
+        /// Channels enabled for writing.
+        bool mColourMask[4];
```
paroj: instead of introducing a new property, which duplicates `mColourWrite`, you should rather extend `mColourWrite` as
`mColourWriteR`, `mColourWriteG` .. Please use the `: 1;` syntax instead of arrays here to keep the size of `Pass` small.

This will avoid inconsistencies like e.g. `mColourWrite == 0 && mColourMask[2] == 1` on the conceptual level.

Consequently extend `colour_write` in scripts with an extended 4 channel format, where the current format would set all 4 channels at once.

PR #813, opened 2018-07-16
Update v2-1 Doxygen site

Ybalrid:
Hi,

So, I think I talked about this once or twice on the gitter channel, but the Doxygen documentation uploaded here on the OgreCave github pages is **way** out of date (like, may 2017 out of date). I am aware of a number of API changes that have gone by since then. So I decided to update it myself :wink:

This is just the normal output of `make OgreDoc` put into the /api/2.1 folder for the website. 

I'll be more than happy to keep this thing up to date via pull requests if you are okay with that. :wink:


paroj:
I squashed your update into the initial 2.1 upload to keep the gh-pages branch small

Ybalrid:
@paroj that's a good thing to do. There's no need to keep track of the diff of thousands of html/javascript files that where auto-generated by a tool! ;-)

PR #824, opened 2018-07-24
Add symlinks to fix broken links on forums and post about 2.1 docs

Ybalrid:
See these message on the forums https://forums.ogre3d.org/viewtopic.php?f=25&t=92874&start=25#p542183

Signed-off by: Arthur Brainville (Ybalrid) <ybalrid@ybalrid.info>

### DIFF
--- a/api/2.1/SettingUpOgre.html
+++ b/api/2.1/SettingUpOgre.html
@@ -1,0 +1,1 @@
+_setting_up_ogre.html
--- a/api/2.1/UsingOgreInYourApp.html
+++ b/api/2.1/UsingOgreInYourApp.html
@@ -1,0 +1,1 @@
+_using_ogre_in_your_app.html


paroj:
there are many more pages with broken links then the two mentioned on the forums. Also this is not a sustainable solution as these two manual symlinks are among dozens of autogenerated and autonamed files. Therefore we should just get over it and update the references.

darksylinc:
I will have to disagree on this one.

It's true there's a lot of autogen-ed files; but these ones were written _by hand_ and are part of the first experience with Ogre.
We have forum posts, posts in other websites, wikis, emails with links to these URLs.

We need stable links at least for those files, rather than something that could change with every doxygen release.

paroj:
the hope is that the links will keep stable from now on.

But as said I also have issues with this specific fix:
- it should include all subpages [here](https://ogrecave.github.io/ogre/api/2.1/manual.html) and [here](https://ogrecave.github.io/ogre/api/2.1/_setting_up_ogre.html)
- doxygen names some intermediate files by their hashes. Therefore when updating the docs I purge the directory and only add the freshly generated files. The fix should survive this.

So probably we rather need a script that does the symlinking.

PR #829, opened 2018-07-31
Samples/PNTriangles: fixed exception in light setup

only-a-ptr:
Light was being parented to 2 scene nodes, causing OGRE to throw Exception::ERR_INVALIDPARAMS every time the sample was starting.


paroj:
thanks

PR #830, opened 2018-08-03
Main: move the adjacency flag to render operation where it belongs

paroj:
instead of the geometry shader - for e.g. rendering adjacency buffers
without an geometry shader attached.

patch based on
https://forums.ogre3d.org/viewtopic.php?f=4&t=94417#p542277


paroj:
@darksylinc this is relevant for 2.x as well

Issue #857, opened 2018-09-04
emscripten_set_canvas_size is deprecated. 


Apivan:
emscripten_set_canvas_size, which is used at https://github.com/OGRECave/ogre/search?q=emscripten_set_canvas_size%2C&unscoped_q=emscripten_set_canvas_size%2C
, is deprecated. 

Please see: https://groups.google.com/forum/#!topic/emscripten-discuss/h5Gz0gQrziQ

I would say that this library is probably the best feature-rich webassembly compatible graphic library available right now. It produced the smallest size with high scalability compared to other libraries.

I wish you guys keep focusing on webassembly. Thanks.

paroj:
at this we should also use the "parentWindowHandle" mechanism to specify which canvas should be rendered to.

Can you provide an according pull-request?

apivan2:
Sorry for late reply. I am quite busy at the moment.
This is a proof-of-concept patch.  I will make a proper pull request soon.

```

diff --git a/RenderSystems/GLSupport/include/EGL/Emscripten/OgreEmscriptenEGLWindow.h b/RenderSystems/GLSupport/include/EGL/Emscripten/OgreEmscriptenEGLWindow.h
index f0a485295..104157354 100644
--- a/RenderSystems/GLSupport/include/EGL/Emscripten/OgreEmscriptenEGLWindow.h
+++ b/RenderSystems/GLSupport/include/EGL/Emscripten/OgreEmscriptenEGLWindow.h
@@ -54,6 +54,7 @@ namespace Ogre {
         int mCSAA;
         int mOldWidth;
         int mOldHeight;
+        std::string mCanvasName;
         
     protected:
         virtual void getLeftAndTopFromNativeWindow(int & left, int & top, uint width, uint height);
diff --git a/RenderSystems/GLSupport/src/EGL/Emscripten/OgreEmscriptenEGLWindow.cpp b/RenderSystems/GLSupport/src/EGL/Emscripten/OgreEmscriptenEGLWindow.cpp
index 5be80263c..5afc4f25e 100644
--- a/RenderSystems/GLSupport/src/EGL/Emscripten/OgreEmscriptenEGLWindow.cpp
+++ b/RenderSystems/GLSupport/src/EGL/Emscripten/OgreEmscriptenEGLWindow.cpp
@@ -89,7 +89,7 @@ namespace Ogre {
         mWidth = width;
         mHeight = height;
         
-        emscripten_set_canvas_size(mWidth, mHeight);
[...]

PR #865, opened 2018-09-15
use emscripten_set_canvas_element_size

Apivan:
https://github.com/OGRECave/ogre/issues/857

### DIFF
--- a/Components/Bites/src/OgreApplicationContext.cpp
+++ b/Components/Bites/src/OgreApplicationContext.cpp
@@ -379,6 +379,8 @@ NativeWindowPair ApplicationContext::createWindow(const Ogre::String& name, Ogre
 #if OGRE_PLATFORM == OGRE_PLATFORM_EMSCRIPTEN
     SDL_GL_CreateContext(ret.native);
     miscParams["currentGLContext"] = "true";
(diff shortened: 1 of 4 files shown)


RenderSystems/GLES2/src/OgreGLES2RenderSystem.cpp
```diff
@@ -181,6 +181,15 @@ namespace Ogre {
 
         initConfigOptions();
 
+#if OGRE_PLATFORM == OGRE_PLATFORM_EMSCRIPTEN
```
paroj: this part is not needed as `parentWindowHandle` should only be set by code

Components/Bites/src/OgreApplicationContext.cpp
```diff
@@ -379,6 +379,8 @@ NativeWindowPair ApplicationContext::createWindow(const Ogre::String& name, Ogre
 #if OGRE_PLATFORM == OGRE_PLATFORM_EMSCRIPTEN
     SDL_GL_CreateContext(ret.native);
     miscParams["currentGLContext"] = "true";
+    miscParams["parentWindowHandle"] = ropts["parentWindowHandle"].currentValue;
```
paroj: drop this part: ropts should not contain parentWindowHandle and we should respect whatever is in miscParams

PR #877, opened 2018-09-29
SWIG_ADD_MODULE is deprecated and replaced by SWIG_ADD_LIBRARY

geho2:


paroj:
unfortunately SWIG_ADD_LIBRARY is only available with CMake 3.8, while we only require 3.3. Need to re-evaluate whether bumping the requirements makes sense for the 1.12 release.

paroj:
will break building snaps, so no

PR #906, opened 2018-10-17
Fixed OgreTargets to take different configurations into account

SNiLD:


CMake/InstallResources.cmake
```diff
@@ -259,4 +259,7 @@ install(FILES
    ${OGRE_BINARY_DIR}/cmake/OGREConfigVersion.cmake
    DESTINATION ${OGRE_CMAKE_DIR}
 )
-install(EXPORT OgreTargets DESTINATION ${OGRE_CMAKE_DIR})
```
paroj: please leave this as is, for this..

CMake/Utils/OgreConfigTargets.cmake
```diff
@@ -100,57 +100,63 @@ function(ogre_install_target TARGETNAME SUFFIX EXPORT)
...
-		FRAMEWORK DESTINATION "${OGRE_LIB_DIRECTORY}${OGRE_DEBUG_PATH}/Debug" CONFIGURATIONS Debug
-	  )
+	  install(TARGETS ${TARGETNAME} EXPORT OgreTargetsRelease
+		CONFIGURATIONS Release None ""
+		BUNDLE DESTINATION "bin${OGRE_RELEASE_PATH}"
+		RUNTIME DESTINATION "bin${OGRE_RELEASE_PATH}"
+		LIBRARY DESTINATION "${OGRE_LIB_DIRECTORY}${OGRE_LIB_RELEASE_PATH}${SUFFIX}"
+		ARCHIVE DESTINATION "${OGRE_LIB_DIRECTORY}${OGRE_LIB_RELEASE_PATH}${SUFFIX}"
+		FRAMEWORK DESTINATION "${OGRE_LIB_DIRECTORY}${OGRE_RELEASE_PATH}/Release")
+	  install(TARGETS ${TARGETNAME} EXPORT OgreTargetsRelWithDebInfo
```
paroj: remove this and the following "install" lines, to keep the same structure as before, but having the CONFIGURATIONS at the correct place. You can use a single install statement for multiple configurations.

CMake/Utils/OgreConfigTargets.cmake
```diff
@@ -100,57 +100,63 @@ function(ogre_install_target TARGETNAME SUFFIX EXPORT)
...
-		ARCHIVE DESTINATION "${OGRE_LIB_DIRECTORY}${OGRE_LIB_DEBUG_PATH}${SUFFIX}" CONFIGURATIONS Debug
-		FRAMEWORK DESTINATION "${OGRE_LIB_DIRECTORY}${OGRE_DEBUG_PATH}/Debug" CONFIGURATIONS Debug
-	  )
+		CONFIGURATIONS Release None ""
+		BUNDLE DESTINATION "bin${OGRE_RELEASE_PATH}" 
+		RUNTIME DESTINATION "bin${OGRE_RELEASE_PATH}"
+		LIBRARY DESTINATION "${OGRE_LIB_DIRECTORY}${OGRE_LIB_RELEASE_PATH}${SUFFIX}"
+		ARCHIVE DESTINATION "${OGRE_LIB_DIRECTORY}${OGRE_LIB_RELEASE_PATH}${SUFFIX}"
+		FRAMEWORK DESTINATION "${OGRE_LIB_DIRECTORY}${OGRE_RELEASE_PATH}/Release")
+	  install(TARGETS ${TARGETNAME}
```
paroj: as above

CMake/Utils/OgreConfigTargets.cmake
SNiLD: Unfortunately that does not work. When you build Release in that case you end up with Debug paths. When you build Debug you end up with almost empty file.

PR #922, opened 2018-11-01
Update OgrePageManager.cpp

wxyu247:
mCameraList removeListener when the PageManager is destructed


paroj:
which issue is this supposed to solve?

wxyu247:
In the Ogitor, the PageManager is deleted first, and then the Camera is deleted, causing the Listener in the Camera to become a wild pointer.

PR #932, opened 2018-11-07
Fixed the inconsistency use of CMAKE_BINARY_DIR and CMAKE_SOURCE_DIR …

choiip:
…in some cases. Instead, they are replaced by OGRE_BINARY_DIR and OGRE_SOURCE_DIR correspondingly.

### DIFF
--- a/Components/Bites/include/OgreAdvancedRenderControls.h
+++ b/Components/Bites/include/OgreAdvancedRenderControls.h
@@ -34,7 +34,7 @@ THE SOFTWARE.
 #include "OgreCamera.h"
 
 #ifdef OGRE_BUILD_COMPONENT_RTSHADERSYSTEM
-#include <OgreRTShaderSystem.h>
+#include "OgreRTShaderSystem.h"
 #endif
 
 namespace OgreBites {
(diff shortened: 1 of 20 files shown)


choiip:
According to the official documentation https://cmake.org/cmake/help/v3.0/command/project.html, the variables OGRE_* is set when project(OGRE) is made. In Ogre, the directory containing the root CMakeLists.txt is the "SOURCE_DIR", which is the same as CMAKE_SOURCE_DIR if the cmake command is run with this as working directory. In other word, CMAKE_SOURCE_DIR == OGRE_SOURCE_DIR and CMAKE_BINARY_DIR == OGRE_SOURCE_DIR.

However, in my situation, I use add_subdirectory(ogre) in another CMakeLists.txt. The mentioned variables are not equal correspondingly. I review all of the related code and make the changes for this reason.

Another reason to make the changes is consistency. Take a look at Dependencies.cmake. The ZZIPlib is downloaded at  ${OGRE_BINARY_DIR} but then the cmake command is run at  ${CMAKE_BINARY_DIR}. This seems to be inconsistent when it is comparing to the 'building SDL2' code snippet.

paroj:
thanks for the explanations. Still I would opt for using `PROJECT_*_DIR` variants as I think they are more common (at least to me).

paroj:
messed up the rename. continued at #937

PR #944, opened 2018-11-14
Docs: point out render order if DisableSort is used

mimon:
Based on what dark_sylink wrote in https://forums.ogre3d.org/viewtopic.php?f=2&t=94628#p543341


paroj:
v2 PR should be on bitbucket

PR #955, opened 2018-11-22
Docs: Resource Management - group information on a single manual page

paroj:
instead of spreading it across multiple API calls


akien-mga:
Thanks a lot, that's very useful!

PR #967, opened 2018-11-29
GLSupport: remove dependency on legacy libGL.so

xuzhongxing:


paroj:
does this change the semantics of the function?

also could you instead use `getProcAddress("glXGetFBConfigFromVisualSGIX");` to remove the dependency?

PR #1008, opened 2018-12-18
Added option for building legacy light lists

bpaberg:
Also setting light list auto-param in HlmsLowLevel material.

This optionally re-enables lighting of legacy materials, now implemented
with HlmsLowLevel.

Note that the legacy light lists cost a lot of performance and is not
used by real hlms materials.


darksylinc:
Added your patch to Bitbucket's mirror https://bitbucket.org/sinbad/ogre/commits/a8abc76f6200bd8f84a492623b57d6814fc59079

It should make its way to this repo soon when Pavel syncs the mirrors (I work on 2.x, usually on the mercurial repo, Pavel works 1.x, usually on the git repo).

Thank you for your contribution!

PR #1013, opened 2018-12-23
RTSS: add fresnel term to lighting

### LINKED ISSUE #1138: Blinn Phong physical BRDF
- must follow energy conservation/ fresnel rules
- should still work with blinn phong values as defined in current scripts.
- needs some value conversion
- http://simonstechblog.blogspot.com/2011/12/microfacet-brdf.html

paroj:
fixes #1138


paroj:
obsolete

PR #1028, opened 2019-01-03
Correction on IOS configuration

arkeon7:
Correct a mistake in cmake scripts


paroj:
why do you think this is a mistake?

arkeon7:
Because it was not working last time I build on IOS, and this change corrected it.
also there is a old port about it https://forums.ogre3d.org/viewtopic.php?t=76067

paroj:
that report was about ANDROID and led to the code as it is now. AFAIK xcode is supposed to replace  the `$(` stuff. If you grep for it, we have it all over the CMake files.

arkeon7:
Ok weird last time I've updated xcode it lead to an error. I don't remember well when I did this change, maybe xcode just finally took it...

PR #1032, opened 2019-01-07
 Python: organize all modules in parent Ogre package 

paroj:


paroj:
@robbmcleod what do you think?

robbmcleod:
As I mentioned on the forum, I would lower-case the module names. It is the expected style:

https://www.python.org/dev/peps/pep-0008/#package-and-module-names

I would probably keep `_Ogre` as `ogre` rather than `main`. Generally to reduce typing (and dereferences, those are not free in Python), I would import the modules as follows:

    from ogre3d import ogre, bites, rtshader

Typing `main.getSceneManager()` for example is a little weird.

And the reason why I keep naming the parent package `ogre3d` is because `ogre` is already taken on PyPi:

https://pypi.org/project/OGRe/

paroj:
I see, lets go for `ogre3d` then.

The idea here was to hide `Main` such that `import Ogre` would transparently import `Ogre.Main`. The remaining components would have to be imported separately.

I did not import other components inside `__init__.py` as they are optional (might not be present). Furthermore there are external modules as https://github.com/OGRECave/ogre-imgui which would install into the top level ogre module.

your offer in the forums to take care of python packaging is very welcome as I do not have much experience with this - especially on windows.

paroj:
I just found out that the pip package name does not have to be the same as the python package name:
https://pypi.org/project/opencv-python/

so lets keep Ogre for compatibility

PR #1033, opened 2019-01-07
Add "bufferScale" option to scale Android 3D buffer

arkeon7:


arkeon7:
it seems to fail on OgreScriptCompiler.cpp now, but I didn't change anything on it :/

paroj:
the whitespace check has some false positives if master moves ahead of your fork. We can merge it as is.

PR #1036, opened 2019-01-09
Terrain: force checkQuadIntersection inline and make it private

paroj:


ulteq:
Measured runtime of calling `Ogre::TerrainGroup::saveAllTerrains` on this [terrain](http://www.austingratzer.com/rigs/downloads.php?do=file&id=154):

| Master | OGRE_FORCE_INLINE | Manual inline |
|:---------:|:---------:|:---------:|
| 6.590 s | 6.271 s | 5.990 s

gcc version 8.2.1 with `-O2` optimization enabled.

PR #1040, opened 2019-01-10
fix getRotationTo returning nan quaternion for zero vectors, return I…

Altren:
…DENTITY instead


paroj:
I would prefer the faster check of `b != 2*a`

Altren:
When both a == 0 and b == 0 (at least one zero vector) your check would fail

Altren:
Both branches could produce zero axis there

paroj:
I dont like computing the axis just to throw it away afterwards when we could detect the degenerate case ahead.
We are squeezing for 5% elsewhere https://github.com/OGRECave/ogre/pull/1036, and this is potentially hot code.

paroj:
```cpp
if (Math::RealEqual(b, 2 * a) || a == 0)
    return Quaternion::IDENTITY;
```
before the axis computation should do.

Also [please add a test here](https://github.com/OGRECave/ogre/blob/master/Tests/OgreMain/src/QuaternionTests.cpp#L34) to validate the new bahavior.

Altren:
Git it, will check this version and add unit tests for the edge cases

PR #1041, opened 2019-01-10
fix negative particle size when scale affector decrease size

Altren:


paroj:
how is this supposed to happen? this should hold: mScaleAdj >= 0 && timeElapsed >= 0

Altren:
I do have negative mScaleAdj in my particles
```
	affector Scaler
	{
		rate	-130
	}
```And this doesn't seem to be wrong, it decreases particle dimensions

PR #1042, opened 2019-01-11
WIP: Structure Python bindings into Python package

robbmcleod:
Nowhere near complete yet, I just want a place to discuss how to go about integrating the CMake build process, the Appveyor/TravisCI integration services, and Python packages.

- Copy the structure in `Components/Python` into the build directory provided 
  by CMake.
- Have CMake set environment variables that can be passed into `setup.py`.
- Ensure that `setup.py` has all the files it needs, including possibly 
  default media files?


Components/Python/BUILD_WIN.txt
```diff
@@ -0,0 +1,208 @@
...
+`OGRE_BUILD_COMPONENT_PYTHON` is checked. 
+
+IMPORTANT: at present (2018-Dec-30), you will also have to turn off
+`OGRE_BUILD_COMPONENT_CSHARP` bindings in the configuration step, as they are 
+broken.
+
+`Configure` again, then `Generate` and finally `Open Project` in MSVC2017. 
+Then you can build the entire solution, which takes about five minutes.
+
+Co-locating the Components
```
paroj: if you build the install target all of them are in `build/sdk/bin`

Components/Python/BUILD_WIN.txt
```diff
@@ -0,0 +1,208 @@
+Python for Ogre on Windows
```
paroj: probably this can ultimately be folded into BuildingOgre.md

appveyor.yml
```diff
@@ -2,34 +2,67 @@ platform: x64
...
+
+init:
+  - "ECHO %PYTHON_VERSION% %MINICONDA%"
+
+install:
+ - "set PATH=%MINICONDA%;%MINICONDA%\\Scripts;%PATH%"
+ - "python --version"
+ - conda update -q conda
+ - conda info -a
+ - conda create -q -n env_ogre python=%PYTHON_VERSION% numpy freetype freeimage swig
```
paroj: freeimage is optional. I would probably rather go without it by default

appveyor.yml
```diff
@@ -2,34 +2,67 @@ platform: x64
 configuration: Release
 image: 
  - Visual Studio 2017
- - Visual Studio 2013
+#  - Visual Studio 2013
+
+# Do we want to have a Matrix? 32-bit and 64-bit?
```
paroj: 64bit only should be enough. We only provide a 64bit SDK and nobody complained so far

Samples/Python/ex_pyside2.py
```diff
@@ -0,0 +1,250 @@
...
+    A work-in-progress example of how to embed Ogre into a PySide2/Qt5 application.
+    """
+
+    def __init__(self, parent: qw.QWidget=None):
+        qw.QOpenGLWidget.__init__(self, parent=parent)
+        self._root = Ogre.Root()
+        
+        # Ensure we are using the basic OpenGL renderer
+        for renderer in self._root.getAvailableRenderers():
+            if renderer.getName() == 'OpenGL Rendering Subsystem':
```
paroj: "OpenGL" in renderer.getName()

robbmcleod:
Is there any nice way to build Appveyor interactively, instead of making one tiny incremental commit after another? I tried searching online but didn't find anything.

paroj:
this maybe? 
https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

paroj:
we now have working python bindings in master via #1069

robbmcleod:
Ok, managed to login via RDP, waiting to see the result of the build for the 2017 build.

Will probably have to figure out how to disable the Python build for MSVC2013, if continued support of that compiler is important, as I believe there is no Python version that supports it:

https://wiki.python.org/moin/WindowsCompilers

robbmcleod:
The RDP session dies as soon as the build errors, so not very useful...

I'm back to building locally, trying to use the `chocolatey` version of `swig`, which seems to be working a bit better than the `conda` version. In which case we should consider building with the C:\Python37 install, and use `pip` to install `numpy` instead.

robbmcleod:
A local environment can mimic Appveyor by setting:

    set APPVEYOR=1
    set APPVEYOR_BUILD_WORKER_IMAGE="Visual Studio 2017"

The `ci-build.cmake` script seems to search for SWIG in different places. Rather than specifying paths it's probably better to just call `find_package(SWIG 3.0.8)` although this needs to be checked on Linux as well. On my system, the `.py` files end up in `ogre\build\Components\Python` and the rest of the library in `ogre\build\lib\Release` including the `.pyd` files, so some copying to a central location, along with the `ogre\Components\Python` package itself will have to be done within the `ogre\Components\Python\CMakeLists.txt` script I suspect.

paroj:
> so some copying to a central location

this happens if you build the INSTALL target. Although the locations are currently derived from where Linux would expect them.

robbmcleod:
I am building the INSTALL target. I don't know where it installs though. Looking at the generated Visual Studio projects there's no commands to copy files.

Is it supposed to be this line?

    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${OGRE_LIB_DIRECTORY}")

paroj:
these lines:
https://github.com/OGRECave/ogre/blob/master/Components/Python/CMakeLists.txt#L25-L26

on windows stuff ends up in build/sdk/ by default

paroj:
closing as there is now a conda-forge project: https://github.com/conda-forge/ogre-feedstock

PR #1045, opened 2019-01-12
return IDENTITY in getRotationTo for zero-sized vectors

Altren:


paroj:
you can also force-push to an existing PR without opening a new one. The whitespace errors were due to master being ahead of your fork (i.e. unrelated). However you can fix them by rebasing on master.

PR #1046, opened 2019-01-12
move OgreUTFString to Overlay component

Altren:
also get rid of no longer needed OGRE_UNICODE_SUPPORT: since it was set to 0 only for the gcc3 on mingw, which is no longer supported (no C++11)

### DIFF
--- a/OgreMain/include/OgrePlatform.h
+++ b/OgreMain/include/OgrePlatform.h
@@ -210,41 +210,7 @@ namespace Ogre {
 #       define OGRE_DEBUG_MODE 0
 #   endif
 
-// Disable unicode support on MingW for GCC 3, poorly supported in stdlibc++
-// STLPORT fixes this though so allow if found
-// MinGW C++ Toolkit supports unicode and sets the define __MINGW32_TOOLBOX_UNICODE__ in _mingw.h
-// GCC 4 is also fine
-#if defined(__MINGW32__)
-# if OGRE_COMP_VER < 400
-#  if !defined(_STLPORT_VERSION)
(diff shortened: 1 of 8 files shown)


OgreMain/include/OgrePlatform.h
```diff
@@ -210,40 +210,13 @@ namespace Ogre {
...
-# endif
-#else
-#  define OGRE_UNICODE_SUPPORT 1
-#endif
-
 #endif // OGRE_PLATFORM == OGRE_PLATFORM_WIN32 || OGRE_PLATFORM == OGRE_PLATFORM_WINRT
 
 //----------------------------------------------------------------------------
 // Linux/Apple/iOS/Android/NaCl/Emscripten Settings
 #if OGRE_PLATFORM == OGRE_PLATFORM_LINUX || OGRE_PLATFORM == OGRE_PLATFORM_APPLE || OGRE_PLATFORM == OGRE_PLATFORM_APPLE_IOS || \
```
paroj: this if is empty now

Altren:
I kept it on purpose, for the sake of complete list of platforms there. I can add somethethig like explicit "// empty" there. Do you think that it should be removed?

paroj:
yeah.. its no different then any other dead code. E.g. there is no NaCl anymore..

OgreMain/include/OgrePlatform.h
Altren: fixed

paroj:
@Altren https://gitter.im/OGRECave/ogre?at=5c3cb71e20b78635b621747c

PR #1047, opened 2019-01-13
fix FreeImageCodec and STBICodec public for static builds usage

Altren:


paroj:
while not strictly necessary if you access the class through `Plugin*`, it is certainly cleaner

PR #1079, opened 2019-01-31
allow compiling shaders directly so apps can implement background asset streaming, provide auto updating camera world position data source when camera relative rendering is enabled.

creadmefford:
i use this to compile dx9 shaders on background threads.

### DIFF
--- a/OgreMain/include/OgreHighLevelGpuProgram.h
+++ b/OgreMain/include/OgreHighLevelGpuProgram.h
@@ -73,8 +73,13 @@ namespace Ogre {
         /// Preprocessor options
         String mPreprocessorDefines;
 
-        /// Internal load high-level portion if not loaded
+    public:
+
+        // Allow compiling shaders directly so apps can implement background asset streaming.
         virtual void loadHighLevel(void);
+
+    protected:
+
         /// Internal unload high-level portion if loaded
         virtual void unloadHighLevel(void);
         /** Internal load implementation, loads just the high-level portion, enough to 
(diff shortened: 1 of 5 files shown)


OgreMain/include/OgreHighLevelGpuProgram.h
```diff
@@ -73,8 +73,13 @@ namespace Ogre {
         /// Preprocessor options
         String mPreprocessorDefines;
 
-        /// Internal load high-level portion if not loaded
+    public:
+
+        // Allow compiling shaders directly so apps can implement background asset streaming.
```
paroj: this method is not really meant for public use as it is not properly locked.
Would a simple `load()` work as well or is `createLowLevelImpl` problematic?

alternatively use `createParameters` to trigger `loadHighLevel` only.

creadmefford: while the game is playing, i'm loading the next chunk and compiling it's DX9 shaders on worker threads off the main thread.  i don't need locking since only one thread is compiling a particular shader.  once the compiling is done i only use it on the main rendering thread.

if i call Load(), it uses other DX9 APIs that are not safe off the main thread.  if we could split Load() into two: Load() and Prepare() like other resources (meshes, textures etc) that could work.  

Prepare() - essentially call LoadHighLevel() which is safe to call on a worker thread, sort of like mesh prepare() or texture prepare(). 
Load() - do the remaining work which is specific to the render thread.

PR #1081, opened 2019-02-01
Fix the CMAKE_MAKE_PROGRAM variable is not set in 3rd party libraries

choiip:
I found that if the 'ninja' program is not install in the default path (e.g. /usr/bin/ninja), the 3rd party libraries  build will be fail and output the error "CMAKE_MAKE_PROGRAM" is not set.


choiip:
It seems that the ZZip source isn't available at that short moment. Any method to retrigger the CI process?

PR #1082, opened 2019-02-01
provide an auto updating camera world position that is available when camera relative rendering is enabled.

creadmefford:
some shaders need the true camera position instead of the camera relative position when camera relative rendering is enabled.

### DIFF
--- a/OgreMain/src/OgreSceneManager.cpp
+++ b/OgreMain/src/OgreSceneManager.cpp
@@ -815,8 +815,11 @@ void SceneManager::_destroySceneNode(SceneNodeList::iterator i)
     if(!(*i)->getName().empty())
         mNamedNodes.erase((*i)->getName());
     OGRE_DELETE *i;
-    std::swap(*i, mSceneNodes.back());
-    (*i)->mGlobalIndex = i - mSceneNodes.begin();
+    if (std::next(i) != mSceneNodes.end())
+    {
+       std::swap(*i, mSceneNodes.back());
+       (*i)->mGlobalIndex = i - mSceneNodes.begin();
+    }
     mSceneNodes.pop_back();
 }
 //---------------------------------------------------------------------


OgreMain/src/OgreAutoParamDataSource.cpp
```diff
@@ -463,6 +463,23 @@ namespace Ogre {
...
     //-----------------------------------------------------------------------------
+    const Vector4& AutoParamDataSource::getCameraPositionWorldSpace(void) const
+    {
+        if (mCameraPositionDirty)
+        {
+            Vector3 vec3 = mCurrentCamera->getDerivedPosition();
+            // We don't include the offset for camera relative rendering.
+            // This is intended for use when camera relative rendering is enabled
+            //   and the true world position is needed by a shader.
+            mCameraPosition[0] = vec3[0];
```
paroj: this will overwrite the relative camera position. Please follow `getCameraPositionObjectSpace`

paroj: just took a closer look at the file. Maybe it is even better to expose `ACT_CAMERA_RELATIVE_OFFSET` instead. We already have `mCameraRelativePosition` and it is used for various other `ACT_*` types.
This way one does not have to create `*_WORLD_SPACE` variants of those.

paroj:
note that you do not have to open a new pull request - force pushing your to your branch is sufficient.

OgreMain/src/OgreAutoParamDataSource.cpp
creadmefford: that sounds great.  i like the idea of giving it a distinct name.  so rename ACT_CAMERA_POSITION_WORLD_SPACE to ACT_CAMERA_RELATIVE_POSITION (and coresponding functions, etc) and return mCameraRelativePosition.  (or did you have something else in mind with "OFFSET"?).

creadmefford:
> note that you do not have to open a new pull request - force pushing your to your branch is sufficient.

thanks for the tip... first time using github.

OgreMain/src/OgreAutoParamDataSource.cpp
paroj: ACT_CAMERA_RELATIVE_POSITION sounds fine

creadmefford:
> note that you do not have to open a new pull request - force pushing your to your branch is sufficient.
do you normally have a single pull request with multiple unrelated commits?  thanks again!

paroj:
messed up the force push - continued at #1084

typically you would create a separate PR per change. This would allow me to squash all fixup commits into one. (this is what I attempted to do manually here)

Also this speeds up merging as there might be objections to the unrelated commit while the original one is good.

creadmefford:
ok thanks for the info.  will do that next time.

PR #1102, opened 2019-02-19
Fixes for Emscripten/Webassembly sample to work again

gabrielcuvillier:
Hello,
I did a couple of minor fixes to have Ogre build again with Emscripten / WebAssembly. It was somehow broken since some time.

So the existing demo is now working again out of the box, with latest Ogre version. See here: 
http://wasm.continuation-labs.com/ogredemo

Changes:
- CMakeList.txt: 
-- disabled SSE support on Emscripten. SSE is not supported on WASM (just a warning fix)
-- disabled Java bindings on Emscripten, as it breaks cmake process.
- Samples/Emscripten/media/resources.cfg: Added missing ShadowVolume resources
- Samples/Emscripten/CMakeList.txt:
-- replaced add_definition(...) with set(CMAKE_EXE_LINKER_FLAGS ...), as these options have to be passed at link time to work
-- updated path to SdkTrays.zip
-- updated paths for RTShaderLib resources
-- added path for ShadowVolume resources
- Samples/Emscripten/main.cpp: display exception results with printf directly (there is issues with emscripten_get_callstack and emscripten_run_script)
- Dependencies.cmake: Disabled "find_package(SDL2)" on Emscripten. Not needed as we use the SDL2 port provided (eg. "-s SDL2")
- BuildingOgre.md: updated the Build HowTo with the correct way to build the project with latest versions of Emscripten

NB: the demo is much bigger than the earlier one published. Maybe because there is a new modules/plugins being compiled ? I don't know exactly. CMake report is the following:

Building components:
  + Bites
  + HLMS [BETA]
  + MeshLodGenerator
  + Overlay
  + Property
  + RTShader System
  + RTShader System Core Shaders
  + RTShader System Extensions Shaders
Building plugins:
  + Octree scene manager
  + STBI codec (generic)
  + Particle FX
Building rendersystems:
  + OpenGL ES2/ ES3
Building executables:
  + Samples
Building core features:
  + Mesh Lod
  + DDS image codec (.dds)
  + PVRTC image codec (.pvr)
  + ETC image codec (.pkm, .ktx) 
  + ZIP archives

Feel free to merge!


gabrielcuvillier:
EDIT: there is some issue with tabs/whitespace formatting. Sorry about this.

paroj:
thanks for the PR! please replace the unchanged Context.cpp by the upstream version as:
`git checkout origin/master Samples/Emscripten/Context.cpp`

also run `git diff --check master` to fix the remaining whitespace issues locally.

gabrielcuvillier:
Apparently both .cpp files used CRLF instead of LF as elsewhere in the project. I restored this, but github diff still complains. I don't know why.
Also, main.cpp mixed tabs and whitespaces (4 space). I fixed this to use only whitespaces

gabrielcuvillier:
Context.cpp have been modified, because emscripten_get_callstack was not working properly

PR #1117, opened 2019-03-03
Alonfaraj fixes

alonfaraj:


OgreMain/src/OgreScriptTranslator.cpp
```diff
@@ -3841,7 +3841,7 @@ namespace Ogre{
         // Set the custom parameters
         for(const auto& p : customParameters)
         {
-            if(prog->isSupported() && !prog->setParameter(p.first->name, p.second))
```
paroj: we need this so parameters of unsupported programs will not generate false positives. Think of having GLSL and HLSL shaders in of material file.

Also `mSyntaxCode` is initialised by all programs to sane values on construction. If it is `""` for you, the error must be somewhere else.

alonfaraj: ok, thanks, I will check it again.

The second commit is not related to this one, I will wait for your review.

paroj: the second commit is fine, however there is no easy way to merge individual commits of a PR.

alonfaraj: my bad, it's the my first PR, sorry for that.
I will make a new one.

paroj: if you know your way around git, it is enough to force-push to the existing branch, without creating a new PR.

paroj:
closing this for now

PR #1130, opened 2019-03-18
Devel

paroj:


TheNicker:
> Main: TextureUnitState - drop useless binding types
Is that means that binding textures would be for all the active sharers in the pipeline ?

paroj:
yes, this was always the case anyway. The only exception was (and still is) D3D9

TheNicker:
The code is useless because it's not used as planned. If I recall correctly this has been presented in GSOC 2013 or around that time.
Changes from that time left many half baked implementations mostly in the Direct3D11 render system like the BindingType.
Another step towards making it a whole solution was presented by me in 2015 as shown here: 
here: https://bitbucket.org/LiorL/ogre-rtss/commits/118873e30b70f0e77b21439ccd81a01f5d3d2620?at=default

My point is that I think it's better to have a half baked solution even if it's not complete or it's ugly than to remove every piece of code that is not perfect. In this case, follow through the original author's intent  rather than dismiss it.

paroj:
> My point is that I think it's better to have a half baked solution even if it's not complete 

I disagree. All dead code should be removed and not clutter the actually used code-paths. Some projects even go as far as removing code where no unit tests exist.

> Another step towards making it a whole solution was presented by me in 2015

What is the problem you are trying to solve there? What is wrong with making all Texture Units available to all pipeline stage as is currently the case?

paroj:
probably this: https://anteru.net/blog/2013/porting-from-directx11-to-opengl-4-2-textures-samplers/
(I come from an OpenGL background)

PR #1159, opened 2019-04-07
D3D9: Small fix for FFP params

kPYKx7Ddw4n1aIKZ:
This initializes a HRESULT value that may not be initialized.
If this value is not changed, it may fail on this [lines](https://github.com/OGRECave/ogre/blob/25f5808a8e27c6ae511ebec1c98b104e22840fd7/RenderSystems/Direct3D9/src/OgreD3D9RenderSystem.cpp#L256-L257).


paroj:
thanks!

PR #1183, opened 2019-05-10
Fix resource location priority

EmilyMansfield:
[The documentation](https://ogrecave.github.io/ogre/api/latest/_resource-_management.html#Resource-Location) states that resource locations added first are preferred over those added later. Currently this is not the case because `ResourceGroup::addToIndex` overwrites the stored `Archive` for any file that has already been found; if two files with the same name exist in different resource locations, then the file in the location added *last* will take priority. 

This fixes the ordering and adds a test to check the correct behaviour. A `DummyArchive` class is used to save modifying `resources.cfg` by adding another resource location.

I've tried to follow the existing coding conventions as closely as possible, but if anything's wrong I'll gladly make changes.

### DIFF
--- a/Tests/OgreMain/src/ResourceLocationPriorityTest.cpp
+++ b/Tests/OgreMain/src/ResourceLocationPriorityTest.cpp
@@ -1,0 +1,59 @@
+/*
+-----------------------------------------------------------------------------
+This source file is part of OGRE
+    (Object-oriented Graphics Rendering Engine)
+For the latest info, see http://www.ogre3d.org/
+
+Copyright (c) 2000-2019 Torus Knot Software Ltd
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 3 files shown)


Tests/OgreMain/src/ResourceLocationPriorityTest.cpp
```diff
@@ -0,0 +1,59 @@
...
+*/
+
+#include <gtest/gtest.h>
+
+#include "ResourceLocationPriorityTest.h"
+#include "RootWithoutRenderSystemFixture.h"
+
+#include "OgreArchiveManager.h"
+
+TEST(ResourceGroupLocationTest, ResourceLocationPriority)
```
paroj: :+1:

paroj:
build failures seem unrelated. However I need to investigate them first.

PR #1185, opened 2019-05-10
Update OgreD3D9RenderSystem.cpp

jvanlabeke:
fixes incorrect lighting when the ambiant value of D3DLIGHT9 is left unitialised.


paroj:
thanks

PR #1188, opened 2019-05-14
Fix GLSL install directory in GL3Plus

EmilyMansfield:
CMake was installing the *contents* of the GLSL folder instead of the folder itself, so would complain about a nonexistent GLSL folder in the user's CMake configuration step. This only seems to be an issue if the user explicitly links against `RenderSystem_GL3Plus`, which as I understand is necessary when linking OGRE statically and using the GL3+ RenderSystem.

None of the other RenderSystems appear to share this problem.


paroj:
> None of the other RenderSystems appear to share this problem.

yes, the other RenderSystems properly hide their GLSL part. Therefore instead of fully exposing it with GL3Plus, I would rather hide it there as well.

As far as I can see `OgreGL3PlusRenderSystem.h` is the only external header, including `OgreGLSLShader.h`.
Therefore I would rather convert that to a forward declaration and stop installing the GLSL folder at all.

RenderSystems/GL3Plus/CMakeLists.txt
```diff
@@ -13,24 +13,22 @@ file(GLOB HEADER_FILES "${CMAKE_CURRENT_SOURCE_DIR}/include/*.h")
...
-file(GLOB GLSL_HEADERS "${CMAKE_CURRENT_SOURCE_DIR}/include/GLSL/*.h")
-file(GLOB GLSL_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/src/GLSL/*.cpp")
-
-set(HEADER_FILES ${HEADER_FILES} ${GLSL_HEADERS})
+file(GLOB GLSL_HEADERS "${CMAKE_CURRENT_SOURCE_DIR}/src/GLSL/include/*.h")
+file(GLOB GLSL_SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/src/GLSL/src/*.cpp")
 
 add_definitions(${OGRE_VISIBILITY_FLAGS})
 
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src/GLSL/include)
```
paroj: please keep this at the target_include_directories call below as:
```sh
    "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/GLSL/include>"
```

EmilyMansfield: My mistake, should it be `PRIVATE` too?

paroj: yes, please

paroj:
thanks :+1:

PR #1192, opened 2019-05-16
add Hygon dhyana support to enable CPU feature detection

hygonsoc:
Add Hygon Dhyana CPU Vendor ID detection("HygonGenuine") to enable Hygon CPU Feature Flag.


paroj:
what is the log output before/ after this patch?

I think it should just work even without the patch if the cpuid instruction is supported.

paroj:
closing as per discussion in https://github.com/magnumripper/JohnTheRipper/pull/3983

PR #1205, opened 2019-05-26
Fix SceneManager inconsistencies with `virtual`

EmilyMansfield:
The main problem is that `SceneManager::createLight()` is `virtual` but `SceneManager::createMovableObject()` is not, so any overridden behaviour in `createLight()` will not be run if the user instead calls `createMovableObject("Light")`. This seems like a bug, or at the very least is surprising behaviour. The functions `hasLight()`, `getLight()` etc. have the same issue.

Fixing this by adding `virtual` to `createMovableObject()` and so on then makes things inconsistent with `BillboardSet`, `ParticleSystem`, and others, so those functions have been made `virtual` too.

Finally, some functions such as `findShadowCastersForLight()` and  `_populateLightList()` have comments suggesting that they can be overridden in subclasses, but they are not marked `virtual`.

### DIFF
--- a/OgreMain/include/OgreSceneManager.h
+++ b/OgreMain/include/OgreSceneManager.h
@@ -1041,7 +1041,7 @@ namespace Ogre {
             Custom scene managers are encouraged to override this method to add optimisations, 
             and to add their own custom shadow casters (perhaps for world geometry)
         */
-        const ShadowCasterList& findShadowCastersForLight(const Light* light,
+        virtual const ShadowCasterList& findShadowCastersForLight(const Light* light,
             const Camera* camera);
         /** Render a group in the ordinary way */
         void renderBasicQueueGroupObjects(RenderQueueGroup* pGroup,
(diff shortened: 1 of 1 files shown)


paroj:
the type name of any derived `Light` will not be `"Light"` but e.g `"PCZLight"` due to the `MovableObjectFactory` indirection. Hence `createMovableObject` does not have to be virtual. 

The `virtual` was deliberately removed in cf09575fb277a1ed6170efec42396383b0901fba as it was generally over-used (Java-style). Therefore I would rather update the comments accordingly. Unless you come up with a use case where virtual is indeed needed.

EmilyMansfield:
> the type name of any derived Light will not be "Light" but e.g "PCZLight" due to the MovableObjectFactory indirection.

That's true, I had missed this. In my case I was keeping track of additional information when `Light`s were added and removed but clearly it's more idiomatic to subclass `Light` and add the information directly there instead of storing it in the `SceneManager`.

Sorry for the trouble, I'll close this pull request and open a new one for the fixed comments.

Issue #1099, opened 2019-02-18
Fails to create a mesh from a manual object due to materials in different groups


jvanlabeke:
System Information
- Ogre Version: 1.11.5
 - Operating System / Platform: : Windows 10 x64
- RenderSystem: DX9 / opengl

We use a manual object to create an mesh. This mesh uses materials in differerent groups.

Basically we configure the manual object with the following pseudo-code:

Begin( materialA, groupA)
..
End()

Begin(materialB, groupB)
...
End()

When we convert the manual object to a mesh by calling convertToMesh(), the code try to setup the mesh material using the group given as a parameter to convertToMesh() ... 
which defeats the purpose of giving groups with the Begin() methods.

Please look up at line 787, in file OgreManualObject.cpp:

            sm->setMaterialName(sec->getMaterialName(), groupName);

The material group name stored in the ManualObjectSection sec is not used.

I think the code should be:

            sm->setMaterialName(sec->getMaterialName(), sec->getMaterialGroup());

Whet do you think?
Regards

paroj:
sounds good. can you provide a pull-request?

jvanlabeke:
I can try to write a pull request (would be my first).

But ... after further investigations, the groupName is not forwarded to the subMesh.
And then later the submesh doesnt forward the groupName to the subEntity when we create the entity in a scene.

cf OgreSubMesh.cpp @ line 54:
```
void SubMesh::setMaterialName( const String& name, const String& groupName /* = ResourceGroupManager::AUTODETECT_RESOURCE_GROUP_NAME */)
    {
        mMaterialName = name;
        // TOFIX? here i think we should add a member mGroupName
        mMatInitialised = true;
    }
```

Then OgreEntity.cpp @ line 1474:
```
    void Entity::buildSubEntityList(MeshPtr& mesh, SubEntityList* sublist)
    {
        // Create SubEntities
        size_t i, numSubMeshes;

        numSubMeshes = mesh->getNumSubMeshes();
        for (i = 0; i < numSubMeshes; ++i)
        {
            SubMesh* subMesh = mesh->getSubMesh(i);
            SubEntity* subEnt = OGRE_NEW SubEntity(this, subMesh);
            if (subMesh->isMatInitialised())
                // TOFIX? here, the group from the material submesh is not used
                subEnt->setMaterialName(subMesh->getMaterialName(), mesh->getGroup()); 
            sublist->push_back(subEnt);
        }
    }
```

[...]

PR #1210, opened 2019-05-29
Add MaterialPtr overloads to ManualObject

### LINKED ISSUE #1209: Terrain: _renderCompositeMap does not use the material's group
#### System Information
- Ogre Version: 1.12.1

#### Detailed description
`TerrainMaterialGenerator::_renderCompositeMap` takes a `MaterialPtr` but only passes its name to the composite map plane, not its group, so unless the material is in the default group the material won't be found when the composite map is baked.

This can happen if the user has subclassed `TerrainMaterialGenerator` and is generating textures in their own resource group, or even if they're using the provided `TerrainMaterialGeneratorA` but the `Terrain` itself isn't in the default group.

#### Proposed Fix(es)
This is easy to fix by just passing the material group in the appropriate places in `_renderCompositeMap`, but since the composite map plane is a `ManualObject` there's an opportunity here to make some changes to that. In particular, `ManualObject` and `ManualObjectSection` work with material name and group pairs, but could have overloads added that take `MaterialPtr` instead, like `SubMesh` now has after #1103. This would save some material lookups and make it harder to introduce bugs like the above in the future. #1099 could also be closed simultaneously.

Is there a preference for either of these fixes? I'm happy to submit a pull request for either.

EmilyMansfield:
This fixes #1209 and #1099 and also has a little bit of formatting cleanup.

### DIFF
--- a/Components/Terrain/src/OgreTerrainMaterialGenerator.cpp
+++ b/Components/Terrain/src/OgreTerrainMaterialGenerator.cpp
@@ -115,7 +115,7 @@ namespace Ogre
 
             // set up scene
             mCompositeMapPlane = mCompositeMapSM->createManualObject();
-            mCompositeMapPlane->begin(mat->getName());
+            mCompositeMapPlane->begin(mat);
             mCompositeMapPlane->position(-halfCamDist, halfCamDist, 0);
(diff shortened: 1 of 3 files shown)


paroj:
:+1:

Issue #1191, opened 2019-05-15
D3D11: Mixed types in getCustomAttribute() method


kPYKx7Ddw4n1aIKZ:
#### System Information
- Ogre Version: 1.12
- Operating System / Platform: Windows 10
- RenderSystem: D3D11

#### Detailled description
I recently noticed that some values returned by the getCustomAttribute() method have mixed types or do not exist at all.
- BUFFER
This is mostly of the type `D3D11HardwarePixelBuffer` and is used this way.
But it is only implemented for `D3D11RenderTexture` which can cause crashes at [this position](https://github.com/OGRECave/ogre/blob/d4b31c0586dd23e2fc4d48b087e8303205caa45c/RenderSystems/Direct3D11/src/OgreD3D11RenderSystem.cpp#L1203) with `D3D11MultiRenderTarget` for example.
- DDBACKBUFFER
This is treated like 2 different types: [ID3D11Texture2D**](https://github.com/OGRECave/ogre/blob/d4b31c0586dd23e2fc4d48b087e8303205caa45c/RenderSystems/Direct3D11/src/OgreD3D11MultiRenderTarget.cpp#L115) and [HardwarePixelBuffer**](https://github.com/OGRECave/ogre/blob/d4b31c0586dd23e2fc4d48b087e8303205caa45c/RenderSystems/Direct3D11/src/OgreD3D11Texture.cpp#L604) but is only used as [ID3D11Texture2D**](https://github.com/OGRECave/ogre/blob/d4b31c0586dd23e2fc4d48b087e8303205caa45c/RenderSystems/Direct3D11/src/OgreD3D11DepthBuffer.cpp#L75).

I can't figure out why this was done by the `getCustomAttribute()` method at all because it returns render system specific values (ID3D11Texture2D for example).

paroj:
the lack of proper internal interfaces basically. getCustomAttribute serves two purposes:
1. it checks if the given pointer is of type `D3D11RenderTexture` (other classes return NULL)
2. it allows exposing arbitrary data over an [any-like](https://en.cppreference.com/w/cpp/utility/any) interface

Nr 2 was the original use-case and is still valid  & [document here](https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_render_target.html#a6c32524efacf6da599d5daccf3d9e166).
[...]

PR #1212, opened 2019-05-30
D3D11: Replaced the use of getCustomAttribute() methods by new D3D11RenderTarget interface

kPYKx7Ddw4n1aIKZ:
I created the `D3D11RenderTarget` class/interface and inherited some classes from it.
However, I dont know if this is the best way.
Also, the related calls to the `getCustomAttribute()` methods have not been replaced yet.

Related to #1191.


paroj:
you should also replace all internal calls to `getCustomAttribute` to verify the new API

kPYKx7Ddw4n1aIKZ:
I replaced all internal calls to `getCustomAttribute()`.
This should solve #1191.

PR #1239, opened 2019-08-06
Fix error linking (set build pugixml as shared lib)

baytuch:


RenderSystems/GLSupport/CMakeLists.txt
```diff
@@ -94,7 +94,7 @@ if(OGRE_CONFIG_ENABLE_GL_STATE_CACHE_SUPPORT)
     PROPERTIES COMPILE_DEFINITIONS OGRE_ENABLE_STATE_CACHE)
 endif()
 
-add_library(OgreGLSupport STATIC 
+add_library(OgreGLSupport SHARED
```
paroj: this unrelated change breaks all builds, please revert

Lastick: I did not check the fix in windows, but I saw a compiler bot report. Apparently, this is indeed so. But I was embarrassed when I compiled into Linux that this library was static, when like the others were - shared. I will definitely study the issue in more detail.

baytuch: I did not check the fix in windows, but I saw a compiler bot report. Apparently, this is indeed so. But I was embarrassed when I compiled into Linux that this library was static, when like the others were - shared. I will definitely study the issue in more detail.

paroj: it is intentionally static, so the rendersystem plugins are self-contained. Allowing it shared would require more CMake changes for proper GL/ EGL dependency handling.

paroj:
should be fixed in master

PR #1250, opened 2019-08-26
GLSupport: implement the 'currentGLContext' parameter for macOS cocoa windows 

gKevin:
The `currentGLContext` parameter is defined in the [`Ogre::Root::createRenderWindow()` documentation](https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_root.html#a537b7d1d0937f799cfe4936f6b672620) but was not implemented for macOS (it works fine on Windows and Linux).

I think it would be more convenient to have this option on all platforms so here's the implementation.

I hope you'll find it useful.
Cheers.


paroj:
Thanks

PR #1269, opened 2019-09-18
Remove OSX include which is removed in 1.12

ProfFan:
This directory does no long exist and makes the compiler complain about missing directory.


paroj:
Thank

PR #1270, opened 2019-09-20
GL3+: fix leak when a program object is created but one already exists

gKevin:
I was having a problem when a monolithic GLSL program was re-linked.
The program was re-created through `glCreateProgram` leaking the old one.

I found this error when trying to set transform feedback varyings with [this function](https://www.lgdv.tf.fau.de/publicationen/adaptive-temporal-sampling-for-volumetric-path-tracing-of-medical-data/). `glTransformFeedbackVaryings` was called on the old programId and had therefore no effect.

### DIFF
--- a/RenderSystems/GL3Plus/src/GLSL/src/OgreGLSLMonolithicProgram.cpp
+++ b/RenderSystems/GL3Plus/src/GLSL/src/OgreGLSLMonolithicProgram.cpp
@@ -65,7 +65,10 @@ namespace Ogre {
         {
             uint32 hash = getCombinedHash();
 
-            OGRE_CHECK_GL_ERROR(mGLProgramHandle = glCreateProgram());
+            if(mGLProgramHandle == 0)
+            {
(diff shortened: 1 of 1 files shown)


RenderSystems/GL3Plus/src/GLSL/src/OgreGLSLMonolithicProgram.cpp
```diff
@@ -65,7 +65,10 @@ namespace Ogre {
         {
             uint32 hash = getCombinedHash();
 
-            OGRE_CHECK_GL_ERROR(mGLProgramHandle = glCreateProgram());
+            if(!glIsProgram(mGLProgramHandle))
```
paroj: it should be legit to just check `mGLProgramHandle == 0` here instead of calling glIsProgram

gKevin: It works indeed 👍

paroj:
thanks

PR #1274, opened 2019-09-21
Provide default ctor/copy ctors where possible.

erikogenvik:
This doesn't change anything, but it appeases gcc 9.2.1 which otherwise
absolutely spams the console with warnings about mixing implicit and
explicit copy constructors.


erikogenvik:
This made the assumption that C++11 was used. I'll provide another.

paroj:
C++11 is used, however we still support VS2013 which does not know constexpr:
https://devblogs.microsoft.com/cppblog/c111417-features-in-vs-2015-rc/

also you do not have to create a new PR, force-pushing to an existing one is enough.

PR #1276, opened 2019-09-22
Provide copy and assign ctors where needed.

erikogenvik:
This doesn't change anything, but it appeases gcc 9.2.1 which otherwise
absolutely spams the console with warnings about mixing implicit and
explicit copy constructors.

This implementation should work with C++98.


paroj:
thanks

PR #1291, opened 2019-10-04
Reverse depth buffer implementation for OpenGL3+

SNiLD:
RenderSystem has new function setReverseDepthBuffer for controlling if this is enabled. This function must be called early to take effect.
glClipControl was not found in old gl3w so it had to be updated as well.
GLSL shaders will have REVERSE_Z_BUFFER define defined when reverse depth buffer is in use.
 * Implementations that used gl_Position.z for calculating fog for example need to change their implementation and this define can be used for that.

Normally OpenGL has it's depth range from -1 to 1 but this flag will change it to be from 1 to 0.

### DIFF
--- a/RenderSystems/GLSupport/src/GLSL/OgreGLSLShaderCommon.cpp
+++ b/RenderSystems/GLSupport/src/GLSL/OgreGLSLShaderCommon.cpp
@@ -62,7 +62,14 @@ namespace Ogre {
         if(getLanguage() == "glsles")
             cpp.Define("GL_ES", 5, 1);
 
+        RenderSystem* renderSystem = Root::getSingleton().getRenderSystem();
+        if (renderSystem && renderSystem->isReverseDepthBufferEnabled())
+        {
+            cpp.Define("OGRE_REVERSED_Z", 15, 1);
+        }
+
         String defines = mPreprocessorDefines;
+
         for(const auto& def : parseDefines(defines))
         {
             cpp.Define(def.first, strlen(def.first), def.second, strlen(def.second));
--- a/OgreMain/include/OgreMath.h
+++ b/OgreMain/include/OgreMath.h
@@ -736,7 +736,7 @@ namespace Ogre
            q = - (far + near) / (far - near)
            qn = - 2 * (far * near) / (far - near)
          */
-        static Matrix4 makePerspectiveMatrix(Real left, Real right, Real bottom, Real top, Real zNear, Real zFar);
+        static Matrix4 makePerspectiveMatrix(Real left, Real right, Real bottom, Real top, Real zNear, Real zFar, bool isReverseDepth = false);
 
         /** Get the radius of the origin-centered bounding sphere from the bounding box. */
         static Real boundingRadiusFromAABB(const AxisAlignedBox& aabb);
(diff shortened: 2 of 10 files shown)


paroj:
the whitespace failure is a false positive. However the gl3w update is superficial:
https://github.com/OGRECave/ogre/blob/master/RenderSystems/GL3Plus/include/GL/gl3w.h#L1379

RenderSystems/GLSupport/src/GLSL/OgreGLSLShaderCommon.cpp
```diff
@@ -63,6 +63,17 @@ namespace Ogre {
...
         String defines = mPreprocessorDefines;
+
+        RenderSystem* renderSystem = Root::getSingleton().getRenderSystem();
+
+        if (renderSystem && renderSystem->isReverseDepthBufferEnabled())
+        {
+            if (!defines.empty())
+                defines += ";";
+
+            defines += "REVERSE_Z_BUFFER";
```
paroj: please prefix this with `OGRE_` to avoid conflicts with user defines

paroj:
the only thing missing for this to work is to return `GL_DEPTH_COMPONENT32F` in `GL3PlusRenderSystem::_createDepthBufferFor` if `isReverseDepthBufferEnabled()`, right?

SNiLD:
Seems like that to me. But how to make that happen in `GL3PlusRenderSystem::_createDepthBufferFor` in a proper way? (I have to go and can't make the changes you suggested until monday)

paroj:
for now you can just assume that GL_DEPTH_COMPONENT32F is supported if isReverseDepthBufferEnabled is true.

SNiLD:
Using `depthFormat = GL_DEPTH_COMPONENT32F;` directly there (if reverseDepthBufferEnabled) results in black screen (this did work in 1.11.6).

paroj:
this is strange - nothing significant changed there. Maybe defining `ENABLE_GL_CHECK` will provide some insight what is going wrong. Did you additionally set stencilFormat = 0?

SNiLD:
Ok the blank screen got solved. `GLDepthBufferCommon::isCompatible` uses `_getDepthStencilFormatFor` directly for choosing depth format so I changed `GL3PlusRenderSystem::_createDepthBufferFor` to use it as well. Maybe the GLPlugin should be refactored to respect this format as well for consistency (so that there is only one path everything related uses instead of multiple).

I also noticed logical bug here https://github.com/OGRECave/ogre/blob/master/RenderSystems/GLSupport/src/OgreGLDepthBufferCommon.cpp#L83. If depth buffer and stencil buffer are in the same buffer, stencil format is not necessary GL_NONE, but `GL_STENCIL_INDEX8` when depth is `GL_DEPTH_COMPONENT24` for example.

There also seems to be weird timing issue that the depthFormat does not get set properly if the application starts too fast. If I run it in debugger and set some breakpoints (for example to `OgreGL3PlusDepthBuffer.cpp:42`) and spend some time before continuing from there the depth buffer will work correctly. However if I run the application from debugger without those breakpoints or if I run it without debugger it will have wrong depth buffer format.

paroj:
I slightly cleaned up this for merging and would do so as soon as the CI passes.

While this is probably not complete, it is a good foundation to work on.

Also I would need to set up a small demo to investigate the remaining issues.

PR #1298, opened 2019-10-08
Main: make reverse-Z a ConfigOption and check compatibility in GL3Plus

paroj:
this ensures it is enabled early on and that it cannot be enabled on
RenderSystems not explicitly supporting it.

### DIFF
--- a/OgreMain/include/OgreRenderSystem.h
+++ b/OgreMain/include/OgreRenderSystem.h
@@ -435,26 +435,12 @@ namespace Ogre
         void setDepthBufferFor( RenderTarget *renderTarget );
 
         /**
-         Set reverse depth buffer enabled or disabled (default disabled).
-
-         @note You must call this function EARLY (before creating windows)!
-               You can't change this value after that.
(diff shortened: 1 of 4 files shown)


paroj:
@SNiLD

SNiLD:
This sound like useful way to enable it.

SNiLD:
I tried this and there's a problem. `initConfigOptions()` initializes `mOptions` but there is no way to modify it. `RenderSystem::getRenderWindowDescription()` will initialize `miscParams` which can be used for configuration and will be passed to `GL3PlusRenderSystem::_createRenderWindow` as `const NameValuePairList *miscParams`. So I think instead of `auto it = mOptions.find("Reversed Z-Buffer");` there should be `miscParams->find("Reversed Z-Buffer");`

Also this is first key that does have spaces. Other values are mapped to shorter values without spaces in `RenderSystem::getRenderWindowDescription()`. So maybe it's better to clean the config option there?

paroj:
`RenderSystem::setConfigOption`. You can set it in the ogre.cfg as well.

PR #1343, opened 2019-11-22
map Dear ImGui to SDL_TEXTINPUT for correct text character processing.

creadmefford:
see: https://forums.ogre3d.org/viewtopic.php?f=2&t=95504


paroj:
will merge manually

PR #1357, opened 2019-12-03
GLSupport: add the 'externalGLControl' option to Cocoa render windows

gKevin:
The option was missing for macOS but described in the [`Ogre::Root::createRenderWindow`](https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_root.html#a537b7d1d0937f799cfe4936f6b672620) documentation.


RenderSystems/GLSupport/src/OSX/OgreOSXCocoaWindow.mm
```diff
@@ -128,6 +128,12 @@ - (BOOL)acceptsFirstResponder
 
         ***Key: "currentGLContext" Description: use an externally created OpenGL context (must be current)
          Values: true, false Default: false
+
```
paroj: no need to duplicate the documentation hre

gKevin: Ok, I removed it.

PR #1413, opened 2020-01-10
Void Pointer and casting of MovableObject

JanGeykenORGADATA:
The changes are for the C# wrapper.
They allow the use of System.IntPtr instead of void* marshalling of Swig and it is possible to create Vertex Buffers with Code like that:
```
var vertexBuffer = HardwareBufferManager.getSingleton().createVertexBuffer(offset, vertexCount, HardwareBuffer.Usage.HBU_STATIC_WRITE_ONLY);
vertexBuffer.writeData(0, vertexBuffer.getSizeInBytes(), new IntPtr(verticesPtr), true);
```
The second change makes the implementation of this tutorial possible, there I needed to cast MovableObject to Entity
http://wiki.ogre3d.org/Raycasting+to+the+polygon+level

### DIFF
--- a/OgreMain/include/Ogre.i
+++ b/OgreMain/include/Ogre.i
@@ -87,6 +87,18 @@ JNIEnv* OgreJNIGetEnv() {
 }
 #endif
 
+#ifdef SWIGCSHARP
+  %apply void *VOID_INT_PTR {void *}
+
+  %typemap(csvarout) void * %{
+    get {
+	  global::System.IntPtr cPtr = $imcall;
+	  if (OgrePINVOKE.SWIGPendingException.Pending) throw OgrePINVOKE.SWIGPendingException.Retrieve();
+	  return cPtr;
+	}
+  %}
+#endif
+
 // convert c++ exceptions to language native exceptions
 %exception {
     try {
@@ -497,7 +509,7 @@ SHARED_PTR(Material);
 %include "OgreMaterialManager.h"
 %include "OgreRenderable.h"
 %include "OgreShadowCaster.h"
-//#ifdef SWIGCSHARP
+#ifdef SWIGCSHARP
 %typemap(cscode) Ogre::MovableObject %{
   public bool tryCastEntity(out Entity entity)
   {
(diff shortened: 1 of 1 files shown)


OgreMain/include/Ogre.i
```diff
@@ -547,6 +575,29 @@ SHARED_PTR(Material);
...
+		static void* SWIG_CastMovableObjectToEntity(void * jarg1)
+		{
+          void * jresult ;
+          Ogre::MovableObject *arg1 = (Ogre::MovableObject *) 0 ;
+          Ogre::Entity *result = 0 ;
+
+          arg1 = (Ogre::MovableObject *)jarg1;
+          {
+            try {
+              result = (Ogre::Entity *)arg1;
```
paroj: this will never throw. You probably wanted to do that:
```cpp
result = dynamic_cast<Ogre::Entity*>(arg1);
if(!result)
{
    SWIG_CSharpException ...
}
```

OgreMain/include/Ogre.i
```diff
@@ -497,6 +509,22 @@ SHARED_PTR(Material);
 %include "OgreMaterialManager.h"
 %include "OgreRenderable.h"
 %include "OgreShadowCaster.h"
+#ifdef SWIGCSHARP
+%typemap(cscode) Ogre::MovableObject %{
+  public bool tryCastEntity(out Entity entity)
+  {
+	entity = null;
+  	if (this.getMovableType() != "Entity")
```
paroj: assuming the check on C++ side is working, this is not needed. Also you should either go with the return value or with exception for error handling.

paroj:
I modified your downcast code to be C++ only (thus also usable by other language bindings).
Can you verify whether this also works for you?

JanGeykenORGADATA:
What SWIG version are you using?
I tried to compile it with SWIG 4.0.0 and 4.0.1 and only get "Syntax error in input(1)" in line 517, the closing %} of the %extend.

paroj:
this is not compile tested, as I am on the go

JanGeykenORGADATA:
It now translates to 
`SWIGINTERN Ogre::Entity *Ogre_MovableObject_castEntity(Ogre::MovableObject *self){
    return dynamic_cast<Ogre::Entity*>(self);
 }` 
in OgreCSHARP_wrap.cxx.

Thank you

PR #1430, opened 2020-01-24
Devel

paroj:


SNiLD:
Thanks for notifying! I have to remember to check if that affects the lights when we update. Looking at the code it shouldn't have any effect since the composite map has separate scene manager and we haven't moved our root scene node from the origin.

paroj:
the direction is not affected by the translation. The bug is:
- the direction is set in node local coordinates by default. That is `this = this * targetDir`
- which happens every time `_renderCompositeMap` is called

basically the default behavior is completely nuts. I was hesitant changing due to backwards compatibility, but it will definitely change in 1.13 (likely to TS_PARENT).

PR #1439, opened 2020-02-02
Fixed possible crash in the addImpl() method of the ResourceManager class

kPYKx7Ddw4n1aIKZ:
The iterator of a map returned by the find() method was used without checking it for validity.
This caused a crash.

### DIFF
--- a/OgreMain/src/OgreResourceManager.cpp
+++ b/OgreMain/src/OgreResourceManager.cpp
@@ -138,8 +138,8 @@ namespace Ogre {
             }
             else
             {
-                ResourceWithGroupMap::iterator itGroup = mResourcesWithGroup.find(res->getGroup());
-                result = itGroup->second.emplace(res->getName(), res);
+                auto resgroup = mResourcesWithGroup.emplace(res->getGroup(), ResourceMap()).first;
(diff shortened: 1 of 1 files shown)


paroj:
actually isResourceGroupInGlobalPool will throw if the group does not exists, but we can add this for consistency

paroj:
> This caused a crash.

so you actually triggered one? Can you elaborate on how?

kPYKx7Ddw4n1aIKZ:
@paroj Yes, I did trigger a crash in Ogitor.
I think it is caused by a custom [resource loading listener](https://github.com/OGRECave/ogitor/blob/cf2b7ce9af2b4ea0e60ab33319281624c315f6a4/Ogitor/src/OgitorsRootUtilityFunctions.cpp#L1732-L1748) which [removes an element](https://github.com/OGRECave/ogre/blob/ff12c183065241736346fd78bd97eb80c8668365/OgreMain/src/OgreResourceManager.cpp#L187-L190) from the `mResourcesWithGroup` map causing the crash.
To reproduce it should be enough to build and run Ogitor with the latest Ogre version and load the `SampleOFS` project.

PR #1454, opened 2020-02-10
CMake/Utils/PrecompiledHeader.cmake: Add -c argument to build precompiled headers

ffontaine:
Add "-c" argument when building precompiled headers to fix build with
RELRO.

More information on a similar issue with domoticz can be found here:
https://patchwork.ozlabs.org/patch/1187328:

"The problem AFAICS is that if no -c or similar option is given, GCC
decides what needs to be done based on the rest of the arguments. If the
rest of the arguments include a -Wl,... option, it decides that linking
needs to be done. If the rest of the arguments are just header files, it
decides to create a precompiled header."

Fixes:
 - http://autobuild.buildroot.org/results/8fabf8d270b9257c3a9db6a2f17f1c08ec9428d3

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


paroj:
Thanks

PR #1455, opened 2020-02-10
Checks for strtol_l functions

ffontaine:
strtol_l, strtoul_l, strtoll_l and strtoull_l are not always available
(for example on musl) so check for them and define a fallback if needed
to avoid the following build failure:

```
/home/buildroot/autobuild/instance-1/output-1/build/ogre-1.12.0/OgreMain/src/OgreStringConverter.cpp: In static member function 'static bool Ogre::StringConverter::parse(const String&, Ogre::int32&)':
/home/buildroot/autobuild/instance-1/output-1/build/ogre-1.12.0/OgreMain/src/OgreStringConverter.cpp:253:22: error: 'strtol_l' was not declared in this scope
         ret = (int32)strtol_l(val.c_str(), &end, 0, _numLocale);
                      ^~~~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/107cffe41081ce46441dec8699d6ad0f152bc152

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/OgreMain/include/OgreString.h
+++ b/OgreMain/include/OgreString.h
@@ -46,8 +46,11 @@ THE SOFTWARE.
 #   define strnicmp strncasecmp
 #endif
 
+#if OGRE_PLATFORM == OGRE_PLATFORM_ANDROID || OGRE_PLATFORM == OGRE_PLATFORM_EMSCRIPTEN || \
+	(OGRE_PLATFORM == OGRE_PLATFORM_LINUX && OGRE_NO_LOCALE_STRCONVERT == 1)
 #if OGRE_PLATFORM == OGRE_PLATFORM_ANDROID || OGRE_PLATFORM == OGRE_PLATFORM_EMSCRIPTEN
 #   define locale_t int
+#endif
 #   define strtod_l(ptr, end, l) strtod(ptr, end)
 #   define strtoul_l(ptr, end, base, l) strtoul(ptr, end, base)
 #   define strtol_l(ptr, end, base, l) strtol(ptr, end, base)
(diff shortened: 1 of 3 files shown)


OgreMain/include/OgreString.h
```diff
@@ -55,6 +55,24 @@ THE SOFTWARE.
 #   define strtoll_l(ptr, end, base, l) strtoll(ptr, end, base)
 #endif
 
+#if OGRE_PLATFORM == OGRE_PLATFORM_LINUX
+#if !defined(HAVE_STRTOUL_L)
```
paroj: this will break for projects consuming OGRE, as we do not make sure `HAVE_STRTO*` is defined for them. Also macros in the public header must be namespaced i.e. start with `OGRE_*`.

CMakeLists.txt
```diff
@@ -99,6 +99,25 @@ else ()
   set(OGRE_PLATFORM_X64 FALSE)
 endif ()
 
+# determine if strtoul_l, strtol_l, strtoull_l and strtoll_l are supported
+include(CheckFunctionExists)
+CHECK_FUNCTION_EXISTS(strtoul_l HAVE_STRTOUL_L)
```
paroj: the result of the test must be stored in `OgreBuildSettings.h.in` to propagate to consuming projects.
Also, I would prefer musl to use the `#if OGRE_PLATFORM == OGRE_PLATFORM_ANDROID ...` fallback branch.
Here, you simplify the test to `#if OGRE_NO_LOCALE_STRCONVERT` and determine that variable as you currently do.

ffontaine: Android fallback can't be used "as is" because this will raise a build failure due to locale_t being set to int

OgreMain/include/OgreString.h
ffontaine: OK done

paroj:
:+1:

PR #1478, opened 2020-02-29
GpuSharedParameters - Fixed for offset calculation

### LINKED ISSUE #1476: Shared params & Uniform buffer
Hi,

I'm trying to use the gpu shared parameters in combination with Uniform buffers, but there is an issue with the GpuSharedParameters::_upload() implementation.

The function iterates through the map of params and calculates the offset like this:

```
// in bytes
            size_t length = param.arraySize * param.elementSize * 4;
            mHardwareBuffer->writeData(offset, length, dataPtr);
            offset += length;
```

However, the map is ordered alphabetically, and there is a mismatch between the offset calculated this way and the actual offset in the uniform buffer in the gpu.

Example:
Here is my shared_param declaration

```
shared_params FogParams
{
	shared_param_named sunDirection float4 0 -1 0 0
	shared_param_named sunColor 	float4 1 1 1 0
	shared_param_named sunPower 	float 1
	
	shared_param_named BetaRay float3 		0.0020687620  0.0034983605  0.0072542001
	shared_param_named BetaMie float3 		0.00036545776 0.00047315995 0.00066936179
}

```

and the uniform block (glsl):

```
layout(shared) uniform FogParams
{
	vec4 sunDirection;
	vec4 sunColor;
	float sunPower;
	
	vec3 BetaRay;
	vec3 BetaMie;
	vec4 DensityAlt;
}

```

But when uploading params, the order is

>BetaMie
>BetaRay
>DensityAlt
>...
>(etc)

Therefore, the values aren't uploaded at the right place.

Maybe the GpuConstantDefinitionMap should use an unordered map, or the offset should be calculated using the physical index instead.

LMCrashy:
fixes #1476

### DIFF
--- a/OgreMain/include/OgreGpuProgramParams.h
+++ b/OgreMain/include/OgreGpuProgramParams.h
@@ -451,6 +451,9 @@ namespace Ogre {
         /// Version number of the definitions in this buffer.
         unsigned long mVersion;
 
+		/// Accumulated offset used to calculate uniform location.
+		unsigned int mOffset;
+
         bool mDirty;
 
     public:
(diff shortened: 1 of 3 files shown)


paroj:
:+1: thanks

LMCrashy:
You're welcome.

PR #1505, opened 2020-03-26
Fixes compilation error C2338 when building with double precision

### LINKED ISSUE #1502: error C2338 when building with double precision
#### System Information
- Ogre Version: 1.12.5
- Operating System / Platform: Windows / x64
- Compiler: MSVC 2015

#### Detailed description
Build fails when compiling OgreInstanceBatch.cpp with double precision.

#### Build log
```
E:\Devel\Ogre\v1.12.5-master\ogre-master\OgreMain\src\OgreInstanceBatch.cpp(534): warning C4244: 'initializing': conversion from 'const Ogre::Real' to 'float', possible loss of data
e:\devel\ogre\v1.12.5-master\ogre-master\ogremain\include\OgreVector.h(61): error C2338: must have at least 4 dimensions
```

gmanol:
This seems to fix the issue. I'm not sure why though, any thoughts?

fixes #1502

### DIFF
--- a/OgreMain/include/OgreVector.h
+++ b/OgreMain/include/OgreVector.h
@@ -301,7 +301,7 @@ namespace Ogre
         you interpret the values.
     */
     template<int dims, typename T>
-    class _OgreExport Vector : public VectorBase<dims, T>
+    class Vector : public VectorBase<dims, T>
     {
     public:
         using VectorBase<dims, T>::ptr;


paroj:
you likely needed a clean build, which this change forced

gmanol:
I spend quite some time building and rebuilding the project. I'm not an expert with template programming, so if you have some time please try building with double precision (with and without the patch). This is weird, but I'm sure there is a logical explanation.

paroj:
rebuilding the project without the patch worked fine, thats why I closed the issue.

__declspec(dllexport) is technically not needed, but has some connotation regarding obtaining a function pointer. Therefore I would rather handle this consistently in all Ogre functions.

also try setting OGRE_ENABLE_PRECOMPILED_HEADERS=OFF in CMake. It seems those are not properly updated on configuration change.

gmanol:
Still having the same problem. May I ask about your build environment? MSVC version, CMake version and CMake configuration options mainly. Thank you.

paroj:
only did a quick test with MinGW on Linux so far. Will trigger a CI build with double config to verify

gmanol:
Nice, let's see.

PR #1508, opened 2020-03-27
CI: double test

paroj:


paroj:
this would require workarounds for VS2013 which would make the hard to read Vector code even worse, so I went with #1505 instead

PR #1517, opened 2020-03-31
Main: Swig - extend Matrix3, Matrix4 and Affine3 by non member operator overloads

AndreKamplingORGADATA:
The non member operator overloads were not wrapped by Swig.
In order to to that the %extend directive must be used in the Swig interface file.


paroj:
:+1: thanks

AndreKamplingORGADATA:
Just for my understanding, why are these operators are implemented as non member operator overloads?

paroj:
resolving circlular dependancy of Matrix and Vector in e.g `operator*(const Matrix4& m, const Vector4& v)`.

Also it is the only way to define `operator*(Real s, const Vector4& v)`

PR #1524, opened 2020-04-04
CMake: fix linking against libatomic

paroj:


simonschmeisser:
See https://buildd.debian.org/status/package.php?p=ogre-1.12&suite=experimental for reference

PR #1525, opened 2020-04-05
added macro recursion feature to GLSLPreprocessor

Miss-Yoshino:
GLSL Code before preprocessing:
```
#define RIGHT(left,right) (right)
#define LEFT(left,right) (left)
#define USE(a,b,F) F(a,b)
USE(A,B,LEFT)
USE(RIGHT(A,C),B,LEFT)
USE(A,B,RIGHT(left,LEFT))
```

Current output:
```
()(A, B)
()((C), B)
(())(A, B)
```

Fixed and correct output:
```
(A)
((C))
(LEFT)(A,B)
```

Added some tests.
Everything (concerning this) was checked against the output from `gcc -E [file]` (on gcc (Arch Linux 9.3.0-1) 9.3.0).

Questions, suggestions and/or comments?
(I am calling this a “feature”, but I am also surprised that I have not found any issue relating to this here at all…)


paroj:
👍 thanks

PR #1530, opened 2020-04-10
Remove FindSDL2.cmake as SDL2 provides a sdl2-config.cmake since 2.0.4

simonschmeisser:
Apparently Debian messed up packaging SDL2 and has the headers in architecture dependent subfolders. This would not be an issue if ogre would use the sdl2-config.cmake provided by sdl2 since version 2.0.4 instead of guessing based on a list of possible locations

this is a test PR to see if your CI tells me if that also works on Windows and Mac

there is another option to do
```
FindPackage(SDL2 CONFIG)
if(NOT SDL2_FOUND)
  FindPackage(SDL2)
endif()

```
or to add `/usr/include/${CMAKE_LIBRARY_ARCHITECTURE}/SDL2` to the list


paroj:
On debian we only have `SDL2_INCLUDE_DIRS`, while current upstream only provides the `SDL2::SDL2` target.

welcome to the wonders of cross-platform development.

simonschmeisser:
Let's see if I'm allowed to simply set both ...

paroj:
also build this locally as our CI builds without SDL on Ubuntu.

simonschmeisser:
May I ask why?

paroj:
To cover a wide range of build configurations. Building *with* SDL2 is done on my machine :wink:

simonschmeisser:
Since I don't have xenial either and my machine currently builds the debian image I added this test commit, can remove it later or while squashing on merge

CMake/Dependencies.cmake
```diff
@@ -291,8 +291,15 @@ if(NOT ANDROID AND NOT EMSCRIPTEN)
   macro_log_feature(SDL2_FOUND "SDL2" "Simple DirectMedia Library needed for input handling in samples" "https://www.libsdl.org/" FALSE "" "")
   if(SDL2_FOUND AND WIN32 AND NOT SDL2_BINARY)
     # fix linking static SDL2 on windows
-    set(SDL2_LIBRARY ${SDL2_LIBRARY} winmm.lib imm32.lib version.lib)
+    set(SDL2_LIBRARIES ${SDL2_LIBRARIES} winmm.lib imm32.lib version.lib)
   endif()
+  if(SDL2_FOUND AND NOT TARGET SDL2::SDL2)
+    add_library(SDL2::SDL2 INTERFACE IMPORTED)
```
paroj, line 297: now you can get rid of SDL2_INCLUDE_DIRS and SDL2_LIBRARIES elsewhere altogether

simonschmeisser:
It fails with something about `git diff` on linux, are you running `clang-format` or similar?

paroj:
it runs `git diff --check master` and finds issues in the reverse diff as master moved forward. If you rebase on master (while dropping the TESTING commit), the issue will disappear.

Components/Csharp/CMakeLists.txt
```diff
@@ -40,10 +40,6 @@ if(OGRE_BUILD_COMPONENT_OVERLAY)
 endif()
 
 if(OGRE_BUILD_COMPONENT_BITES)
-	if(SDL2_FOUND)
-		include_directories(${SDL2_INCLUDE_DIRS})
```
simonschmeisser: do you know if this gets build in any/all of the CI builds? If so I could also rely on the transitive target for Java below. If not it might be broken ...

paroj: the MSVC and Ubuntu Builders have all bindings enabled

paroj:
:+1: thanks

PR #1544, opened 2020-04-18
Fix build on FreeBSD.

arrowd:


paroj:
will fix the build failure due to updates in zzip first

paroj:
:+1: thanks

PR #1553, opened 2020-04-24
Fixed badly calculated matrices alignement

LMCrashy:
Related to issue https://github.com/OGRECave/ogre/issues/1476

Matrices params size was not properly handled when calculating the aligned offset.
This new fix should handle every case.

Tested it against this Uniform block:
```

layout(std140) uniform TestAlignment
{
	mat3 _mat33 ; 	//0
	mat4x3 _mat43 ; //48
	float _float ; 	//112
	mat3x4 _mat34 ; //128
	vec3 _vec3;	//176
	vec2 _vec2; 	//192
	mat4 _mat44; 	//208
	float _float1; 	//272
	float _float2; 	//276
	mat3 _mat33_2; 	//288
	mat2 _mat22 ;	//336
};
```

### DIFF
--- a/OgreMain/include/OgreGpuProgramParams.h
+++ b/OgreMain/include/OgreGpuProgramParams.h
@@ -454,7 +454,7 @@ namespace Ogre {
         unsigned long mVersion;
 
 		/// Accumulated offset used to calculate uniform location.
-		unsigned int mOffset;
+		size_t mOffset;
 
         bool mDirty;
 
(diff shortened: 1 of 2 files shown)


paroj:
```
OgreGpuProgramParams.cpp:407:34: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]

         if (mOffset + align_size > nextAlignedOffset)
```

LMCrashy:
Ah damn'.
We could use "unsigned int" for align_size and nextalignementOffset, or use size_t for everything, as def.elementSize and def.logicalIndex are size_t.

paroj:
>use size_t for everything, as def.elementSize and def.logicalIndex are size_t.

:+1:

PR #1561, opened 2020-04-29
Fix OSX platform libs

Tobias-Fischer:
We failed to build ogre in https://github.com/conda-forge/ogre-feedstock/pull/24 after https://github.com/OGRECave/ogre/commit/397e24cfc4ec96f35cab41c99a05d0458af7e1a0 was merged (see https://github.com/conda-forge/ogre-feedstock/pull/24).

This patch fixes the build (see https://github.com/conda-forge/ogre-feedstock/pull/25). It would be great if it could be merged.

Many thanks,
Tobias


paroj:
thanks. This makes sense. Any idea why our CI did not catch this? Our CI builds everything as framework while yours builds plain libs?

Tobias-Fischer:
Thanks for merging this so quickly! Indeed, we build plain libs which could be the difference ..

PR #1565, opened 2020-05-01
Use the right variable to determine if Qt was found

### LINKED ISSUE #1559: CMake: Wrong variable for Qt in the feature summary
https://github.com/OGRECave/ogre/blob/6d93fafc814ccd2527a6fbf8934cfdc2e7106db1/CMake/Dependencies.cmake#L301

I think that `PYTHONLIBS_FOUND ` is wrong here.

kPYKx7Ddw4n1aIKZ:
Tiny fix #1559.


paroj:
thanks. this got completely lost with me

PR #1592, opened 2020-05-28
Only include <sys/sysctl.h> for iOS

jwakely:
This file includes `<sys/sysctl.h>` if the CPU is ARM, which breaks compilation with the upcoming glibc 2.32 release because that header has been removed. The header only seems to be needed for `sysctlbyname`, which is only used on iOS. Change the `#ifdef` to avoid including it when not needed.

This fixes a build failure in Fedora rawhide: https://bugzilla.redhat.com/show_bug.cgi?id=1841324


paroj:
:+1: thanks - now if only fedora could update from the 7 year old version. that would be great.

PR #1598, opened 2020-06-05
Fix crash in ApplicationContextSDL based applications that lose focus…

creadmefford:
… either from window minimization or by session lock (i.e. user locks desktop or machine goes to sleep).

The application still needs to override the handlers to avoid redrawing ImGui.  Example:

virtual void windowFocusGained (Ogre::RenderWindow* rw)
{
    mFocus = true;
}

virtual void windowFocusLost (Ogre::RenderWindow* rw)
{
    mFocus = false;
}

bool app::frameStarted (const Ogre::FrameEvent& evt)
{
    OgreBites::ApplicationContext::frameStarted (evt);
    if (mFocus)
    {
        Ogre::ImGuiOverlay::NewFrame (evt);
    }
}

### DIFF
--- a/Components/Bites/include/OgreApplicationContextBase.h
+++ b/Components/Bites/include/OgreApplicationContextBase.h
@@ -129,6 +129,8 @@ namespace OgreBites
         virtual bool windowClosing(Ogre::RenderWindow* rw) { return true; }
         virtual void windowClosed(Ogre::RenderWindow* rw) {}
         virtual void windowFocusChange(Ogre::RenderWindow* rw) {}
+        virtual void windowFocusGained (Ogre::RenderWindow* rw) {}
+        virtual void windowFocusLost (Ogre::RenderWindow* rw) {}
 
         /**
          * inspect the event and call one of the corresponding functions on the registered InputListener
--- a/Components/Bites/src/OgreApplicationContextSDL.cpp
+++ b/Components/Bites/src/OgreApplicationContextSDL.cpp
@@ -106,6 +106,18 @@ void ApplicationContextSDL::shutdown()
     }
 }
 
+Ogre::RenderWindow* _findWindow (const std::vector<NativeWindowPair>& windows, Uint32 windowID)
+{
+    for (auto& window : windows)
+    {
+        if (windowID != SDL_GetWindowID (window.native))
+            continue;
+        Ogre::RenderWindow* win = window.render;
+        return win;
+    }
+    return nullptr;
+}
+
 void ApplicationContextSDL::pollEvents()
 {
     if(mWindows.empty())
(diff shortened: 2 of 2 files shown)


Components/Bites/src/OgreApplicationContextSDL.cpp
```diff
@@ -123,17 +135,23 @@ void ApplicationContextSDL::pollEvents()
...
-
-            for(WindowList::iterator it = mWindows.begin(); it != mWindows.end(); ++it)
+            if (event.window.event == SDL_WINDOWEVENT_RESIZED)
             {
-                if(event.window.windowID != SDL_GetWindowID(it->native))
-                    continue;
-
-                Ogre::RenderWindow* win = it->render;
-                win->windowMovedOrResized();
-                windowResized(win);
```
creadmefford, line 136: @paroj  -- _findWindow() assumes we only care about ONE match in mWindow

Components/Bites/include/OgreApplicationContextBase.h
```diff
@@ -129,6 +129,8 @@ namespace OgreBites
         virtual bool windowClosing(Ogre::RenderWindow* rw) { return true; }
         virtual void windowClosed(Ogre::RenderWindow* rw) {}
         virtual void windowFocusChange(Ogre::RenderWindow* rw) {}
```
creadmefford, line 131: not sure what this function is useful for? (and how do you know if it's focus lost or gained?

creadmefford: another approach would be: windowFocusChange (bool gained, ...) instead of 2 virtuals.

paroj:
there are 2 issues with this:
1. The API is dictated by `WindowEventListener`, which uses `windowFocusChange` and indeed lacks a way to retrieve the actual window focus state. Most likely one is supposed to use `RenderWindow::isActive()`
2.  `ImGui::EndFrame()` being called actually only depends on `isActive()` via the `SceneManager::fireRenderQueueStarted` event, which is only skipped if the `RenderTarget` is not active. [See here](https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_root.html#aeaad25268c0ad8007ce8961a32242c77).

with my last fix attempt the only way a mismatch can still happen is this iff `isActive()` yields different results between `renderOneFrame` and `frameStarted`. Digging into the `isActive()`implementation on Windows, indeed shows that the status is queried again with each call: https://github.com/OGRECave/ogre/blob/b0308adc61a8534cb14c69a6d9f0a824e40c7480/RenderSystems/Direct3D9/src/OgreD3D9RenderWindow.cpp#L690

so due to bad timing this can happen. Therefore, the correct fix is to use a event that is only fired after `isActive()` instead of `frameStarted`.

creadmefford:
> Therefore, the correct fix is to use a event that is only fired after `isActive()` instead of `frameStarted`.

it looks like "windowsFocusChange" is from the old OIS days.  i don't see any of these virtual functions in OgreApplicationContextBase.h (windowClosing, windowClosed, windowFocusChange, etc) being called anywhere except "windowResized" inside pollEvents().

are these the events you are referring to?

or are you saying there is a better way to check the window handle (HWND) inside the render window?

paroj:
see #1599

creadmefford:
> see #1599

you're amazing!

paroj:
fixed with #1599

PR #1665, opened 2020-08-11
Improve config-path resolution

fegorsch:
To allow loading config files (e.g. plugins.cfg, resources.cfg) from
arbitrary locations. Up to now it was only possible to load them from
- the CWD,
- the writable path (i.e. somewhere in HOME),
- the location Ogre was originally installed to.

Also resolve relative paths in `resources.cfg` relative to the config-file location.

This commit is especially useful, if you need to start your
application from anywhere else than the location of the config-files.

### DIFF
--- a/Components/Bites/src/OgreApplicationContextBase.cpp
+++ b/Components/Bites/src/OgreApplicationContextBase.cpp
@@ -28,6 +28,13 @@ ApplicationContextBase::ApplicationContextBase(const Ogre::String& appName)
 {
     mAppName = appName;
     mFSLayer = new Ogre::FileSystemLayer(mAppName);
+
+    if (char* val = getenv("OGRE_CONFIG_DIR"))
+    {
+        Ogre::String configDir = Ogre::StringUtil::standardisePath(val);
+        mFSLayer->setConfigPaths({ configDir });
+    }
+
     mRoot = NULL;
     mOverlaySystem = NULL;
     mFirstRun = true;
--- a/Components/Bites/src/OgreApplicationContextBase.cpp
+++ b/Components/Bites/src/OgreApplicationContextBase.cpp
@@ -402,11 +402,12 @@ void ApplicationContextBase::locateResources()
     auto& rgm = Ogre::ResourceGroupManager::getSingleton();
     // load resource paths from config file
     Ogre::ConfigFile cf;
+    Ogre::String resourcesPath = mFSLayer->getConfigFilePath("resources.cfg");
 #if OGRE_PLATFORM == OGRE_PLATFORM_ANDROID
     Ogre::Archive* apk = Ogre::ArchiveManager::getSingleton().load("", "APKFileSystem", true);
-    cf.load(apk->open(mFSLayer->getConfigFilePath("resources.cfg")));
+    cf.load(apk->open(resourcesPath));
 #else
-    Ogre::String resourcesPath = mFSLayer->getConfigFilePath("resources.cfg");
+
     if (Ogre::FileSystemLayer::fileExists(resourcesPath) || OGRE_PLATFORM == OGRE_PLATFORM_EMSCRIPTEN)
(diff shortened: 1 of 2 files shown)


OgreMain/src/WIN32/OgreFileSystemLayer.cpp
```diff
@@ -122,6 +122,17 @@ namespace Ogre
         }
 #endif
 
+        if (char* val = getenv("OGRE_CONFIG_PATHS"))
+        {
+#if OGRE_PLATFORM == OGRE_PLATFORM_WIN32 || OGRE_PLATFORM == OGRE_PLATFORM_WINRT
+            Ogre::StringVector configPaths = Ogre::StringUtil::split(val, ";");
+#else
+            Ogre::StringVector configPaths = Ogre::StringUtil::split(val, ":");
```
paroj, line 130: this file is only compiled on windows

OgreMain/src/WIN32/OgreFileSystemLayer.cpp
```diff
@@ -122,6 +122,17 @@ namespace Ogre
         }
 #endif
 
+        if (char* val = getenv("OGRE_CONFIG_PATHS"))
```
paroj, line 125: what is the use-case of specifying multiple paths here?

OgreMain/src/WIN32/OgreFileSystemLayer.cpp
paroj: to make this cross-platform, you can add after this line:
https://github.com/OGRECave/ogre/blob/a2b9c144e653fd15ebbef0bcb6e37ec8c140cd15/Components/Bites/src/OgreApplicationContextBase.cpp#L30

and call `setConfigPaths`. It will *overwrite* the paths though. Which is consistent with OGRE_PLUGIN_DIR on the other hand.

OgreMain/src/WIN32/OgreFileSystemLayer.cpp
paroj, line 125: also, you should mention OGRE_CONFIG_PATHS somewhere in the docs.

fegorsch: I just wanted to be consistent with `mConfigPaths`. Don't care, though.

paroj: my idea was that if you choose to override the default path resolution, you have a specific path in mind.

Use `OGRE_CONFIG_DIR` then - like the CMake variable:
https://ogrecave.github.io/ogre/api/latest/setup.html#cmake

paroj:
please rebase on master to get rid of the CI failure

paroj:
```
Components/Bites/src/OgreApplicationContextBase.cpp:410: trailing whitespace.

+    
```

Components/Bites/src/OgreApplicationContextBase.cpp
```diff
@@ -28,6 +28,13 @@ ApplicationContextBase::ApplicationContextBase(const Ogre::String& appName)
 {
     mAppName = appName;
     mFSLayer = new Ogre::FileSystemLayer(mAppName);
+
+    if (char* val = getenv("OGRE_CONFIG_DIR"))
+    {
+        Ogre::String configDir = Ogre::StringUtil::standardisePath(val);
+        mFSLayer->setConfigPaths({ configDir });
```
paroj, line 35:
```suggestion
        mFSLayer->setConfigPaths({ configDir });
        LogManager::getSingleton().logMessage(
                "setting ConfigFile Path from OGRE_CONFIG_DIR environment variable");
```

fegorsch: doesn't work, because the singleton is still NULL here.

PR #1674, opened 2020-08-17
Extend Swig Wrapper

AndreKamplingORGADATA:
Extend PixelBox by method returning data as void pointer resulting in C# as IntPtr for direct access.
Accessing the public `data` field results in `SWIGTYPE_p_unsigned_char` which is not usable on C# side.


paroj:
looks like you should wrap this in `#ifdef SWIGCSHARP`

PR #1698, opened 2020-09-04
Fixes issue preventing use of latest emscripten versions

### LINKED ISSUE #1289: Full-screen switching broken with latest emscripten SDK
last verified working version is v1.38.24. Probably related to https://github.com/emscripten-core/emscripten/pull/7977

The workaround is to use emscripten not newer than the version specified above
```
./emsdk install sdk-fastcomp-1.38.24-64bit
```

note: to use DISABLE_DEPRECATED_FIND_EVENT_TARGET_BEHAVIOR=1, at least 1.38.35 is needed due to a fix in SDL2.

also related:
- https://github.com/emscripten-core/emscripten/issues/8590
- https://github.com/emscripten-ports/SDL2/issues/8

### LINKED ISSUE #1686: remove BINARYEN_TRAP_MODE=clamp in Samples/Emscripten/CMakeLists.txt
#### System Information
- Ogre Version: 1.12.8
- Operating System / OSX 10.15.5
- RenderSystem OpenGL
- Emscripten: 2.0.1

#### Detailled description
According to [emscripten-core](https://github.com/emscripten-core/emscripten/issues/10419#)
 BINARYEN_TRAP_MODE is not supported by the LLVM wasm backend

> I looks like you are specifying BINARYEN_TRAP_MODE in your link flags. This flag doesn't mean anything for the wasm backend, so you should remove it.

> You could remove BINARYEN_TRAP_MODE=clamp in Samples/Emscripten/CMakeLists.txt
. BINARYEN_TRAP_MODE was removed in EMSctripten 1.39.0.

#### Ogre.log
74240010-5bcd0b00-4cd9-11ea-81b7-9d232305de93

### LINKED ISSUE #1697: Update to the latest version of Emscripten with upstream LLVM backend
#### System Information
- Ogre Version: 1.12.8 
- Operating System / Platform: OSX
- RenderSystem: GLES2RenderSystem

#### Detailled description
Latest version of emscripten has several fixes including handling of fullscreen see  [changelog](https://github.com/emscripten-core/emscripten/blob/master/ChangeLog.md) not to mention it now uses LLVM backend. [emscripten-llvm-wasm](https://v8.dev/blog/emscripten-llvm-wasm) which has several benefits. I am not familiar with the code base so I am unsure what it entails add support for the newer versions of emscripten. 

#### Browser Console.
Looking at the console logs of [EmscriptenDemo](https://ogrecave.github.io/ogre/emscripten/) as compared to the demo compiled using 2.0.1 it seems as if the problem may be on the RenderSystem base class using emsscripten version 2.0.1
Working Demo:
```
GLES2RenderSystem::_createRenderWindow "OGRE Emscripten Sample", 800x600 windowed  miscParams: FSAA=2 currentGLContext=true displayFrequency=N/A gamma=No vsync=Yes 
(index):1237 GL_VERSION = 2.0.0.0
(index):1237 GL_VENDOR = WebKit
(index):1237 GL_RENDERER = WebKit WebGL
(index):1237 
```

Failing Demo with latest emscripten:
```
RenderSystem::_createRenderWindow "OGRE Emscripten Sample", 800x600 windowed  miscParams: FSAA=0 displayFrequency=N/A gamma=No vsync=Yes 
EmscriptenSample.html:1 exception thrown: SyntaxError: Failed to execute 'querySelector' on 'Document': '0' is not a valid selector.,Error: Failed to execute 'querySelector' on 'Document': '0' is not a valid selector.
    at findEventTarget (http://localhost:8000/EmscriptenSample.js:1:152757)
    at __registerWebGlEventCallback (http://localhost:8000/EmscriptenSample.js:1:212626)
    at _emscripten_set_webglcontextlost_callback_on_thread (http://localhost:8000/EmscriptenSample.js:1:212927)
    at http://localhost:8000/EmscriptenSample.wasm:wasm-function[6519]:0x1e6f2a
    at http://localhost:8000/EmscriptenSample.wasm:wasm-function[6515]:0x1e6ce9
    at http://localhost:8000/EmscriptenSample.wasm:wasm-function[6329]:0x1db4bb
    at http://localhost:8000/EmscriptenSample.wasm:wasm-function[3365]:0xc4236
    at http://localhost:8000/EmscriptenSample.wasm:wasm-function[613]:0x1fa4b
    at http://localhost:8000/EmscriptenSample.wasm:wasm-function[435]:0x11dd7
    at http://localhost:8000/EmscriptenSample.wasm:wasm-function[311]:0xa52f
printErr @ EmscriptenSample.html:1
```

guster32:
fixes #1697 
fixes #1686
fixes #1289
These changes enable the use the latest versions of emscripten. Latest version of emscripten has several fixes including handling of fullscreen see [changelog](https://github.com/emscripten-core/emscripten/blob/master/ChangeLog.md) not to mention it now uses LLVM backend.[ emscripten-llvm-wasm](https://v8.dev/blog/emscripten-llvm-wasm) which has several benefits.

### DIFF
--- a/Samples/Emscripten/CMakeLists.txt
+++ b/Samples/Emscripten/CMakeLists.txt
@@ -1,7 +1,7 @@
 set(MEDIA_DIR ${CMAKE_BINARY_DIR}/bin/media/)
 
 set(CMAKE_EXECUTABLE_SUFFIX ".html")
-set(CMAKE_EXE_LINKER_FLAGS "--preload-file ${MEDIA_DIR}/@. -s BINARYEN_TRAP_MODE=clamp -s EXPORTED_FUNCTIONS=\"['_passAssetAsArrayBuffer', '_clearScene', '_main']\" -s USE_SDL=2")
+set(CMAKE_EXE_LINKER_FLAGS "--preload-file ${MEDIA_DIR}/@. -s EXPORTED_FUNCTIONS=\"['_passAssetAsArrayBuffer', '_clearScene', '_main']\" -s USE_SDL=2")
 
 add_definitions(-s USE_SDL=2)
 
--- a/Samples/Emscripten/CMakeLists.txt
+++ b/Samples/Emscripten/CMakeLists.txt
@@ -1,7 +1,7 @@
 set(MEDIA_DIR ${CMAKE_BINARY_DIR}/bin/media/)
 
 set(CMAKE_EXECUTABLE_SUFFIX ".html")
-set(CMAKE_EXE_LINKER_FLAGS "--preload-file ${MEDIA_DIR}/@. -s EXPORTED_FUNCTIONS=\"['_passAssetAsArrayBuffer', '_clearScene', '_main']\" -s USE_SDL=2")
+set(CMAKE_EXE_LINKER_FLAGS "--preload-file ${MEDIA_DIR}/@. -s MIN_WEBGL_VERSION=2 -s MAX_WEBGL_VERSION=2 -s EXPORTED_FUNCTIONS=\"[ '_passAssetAsArrayBuffer', '_clearScene', '_main']\" -s USE_SDL=2")
 
 add_definitions(-s USE_SDL=2)
 
--- a/Samples/Emscripten/CMakeLists.txt
+++ b/Samples/Emscripten/CMakeLists.txt
@@ -1,7 +1,7 @@
 set(MEDIA_DIR ${CMAKE_BINARY_DIR}/bin/media/)
 
 set(CMAKE_EXECUTABLE_SUFFIX ".html")
-set(CMAKE_EXE_LINKER_FLAGS "--preload-file ${MEDIA_DIR}/@. -s MIN_WEBGL_VERSION=2 -s MAX_WEBGL_VERSION=2 -s EXPORTED_FUNCTIONS=\"[ '_passAssetAsArrayBuffer', '_clearScene', '_main']\" -s USE_SDL=2")
+set(CMAKE_EXE_LINKER_FLAGS "--preload-file ${MEDIA_DIR}/@. -s EXPORTED_FUNCTIONS=\"[ '_passAssetAsArrayBuffer', '_clearScene', '_main']\" -s USE_SDL=2")
 
 add_definitions(-s USE_SDL=2)
 
--- a/RenderSystems/GLSupport/src/EGL/Emscripten/OgreEmscriptenEGLWindow.cpp
+++ b/RenderSystems/GLSupport/src/EGL/Emscripten/OgreEmscriptenEGLWindow.cpp
@@ -52,7 +52,7 @@ namespace Ogre {
           mCSAA(0)
     {
         // already handled by resize
-        //emscripten_set_fullscreenchange_callback(NULL, (void*)this, 1, &EmscriptenEGLWindow::fullscreenCallback);
+        emscripten_set_fullscreenchange_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW, (void*)this, 1, &EmscriptenEGLWindow::fullscreenCallback);
         emscripten_set_webglcontextlost_callback("#canvas", (void*)this, 1, &EmscriptenEGLWindow::contextLostCallback);
         emscripten_set_webglcontextrestored_callback("#canvas", (void*)this, 1, &EmscriptenEGLWindow::contextRestoredCallback);
         emscripten_set_resize_callback("#canvas", (void*)this, 1, &EmscriptenEGLWindow::canvasWindowResized);
@@ -91,6 +91,9 @@ namespace Ogre {
         
 
         EMSCRIPTEN_RESULT result = emscripten_set_canvas_element_size(mCanvasSelector.c_str(), width, height);
+        // This is a workaroud for issue: https://github.com/emscripten-core/emscripten/issues/3283.
+        // The setTimeout of 0 will ensure that this code is runs on the next JSEventLoop.
+        EM_ASM(setTimeout(function(){var canvas = document.getElementById('canvas'); canvas.width = $0; canvas.height = $1;}, 0), width, height);
         
         if(result < 0)
         {
--- a/RenderSystems/GLSupport/src/EGL/Emscripten/OgreEmscriptenEGLWindow.cpp
+++ b/RenderSystems/GLSupport/src/EGL/Emscripten/OgreEmscriptenEGLWindow.cpp
@@ -53,17 +53,17 @@ namespace Ogre {
     {
         // already handled by resize
         //emscripten_set_fullscreenchange_callback(NULL, (void*)this, 1, &EmscriptenEGLWindow::fullscreenCallback);
-        emscripten_set_webglcontextlost_callback(NULL, (void*)this, 1, &EmscriptenEGLWindow::contextLostCallback);
-        emscripten_set_webglcontextrestored_callback(NULL, (void*)this, 1, &EmscriptenEGLWindow::contextRestoredCallback);
-        emscripten_set_resize_callback(NULL, (void*)this, 1, &EmscriptenEGLWindow::canvasWindowResized);
+        emscripten_set_webglcontextlost_callback("#canvas", (void*)this, 1, &EmscriptenEGLWindow::contextLostCallback);
+        emscripten_set_webglcontextrestored_callback("#canvas", (void*)this, 1, &EmscriptenEGLWindow::contextRestoredCallback);
+        emscripten_set_resize_callback("#canvas", (void*)this, 1, &EmscriptenEGLWindow::canvasWindowResized);
     }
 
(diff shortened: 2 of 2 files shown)


guster32:
@paroj @darksylinc Would love to get your feedback on this change. Regards.

paroj:
I had your change implemented at some point, however It broke v1.38.24 compatibility, which is the last version where everything is working #1289 

Maybe they finally fixed that in v2. Will try again with your changes over the weekend.

paroj:
on Ubuntu 20.04 I am getting (both Firefox and Chrome):

ss

and the fullscreen button is still broken (black-screen).

As far as I remember the fullscreen issue is related to the events reporting a 0x0 window (with new emscripten).

As for the shadow issue, I did not yet dig into what the emscripten guys broke there, as the same code used to work and is still working on native GLES2.
Are you aware of something like RenderDoc for WebGL? That would allow tracing down which WebGL call is causing the issue exactly and adding a workaround.

So unfortunately, the newest and greatest emscripten still bears serious regressions which prevents merging this as is.

If, for some reason, you want to use latest emscripten, please surround your changes with
```
#if // with __EMSCRIPTEN_major__ && __EMSCRIPTEN_minor__
```
and leave `BINARYEN_TRAP_MODE`, so we have a sane default.

guster32:
Thank very much for looking into this @paroj. Beside the issues you mentioned I also noticed that with the latest emscripten The ogre can walk past the edges of the map where is with the older version it would not. I will dig deeper into these issues to see If I can figure out what is going on (And learn a ton abt Ogre3D in the process). I can also ping the emscripten guys on the Discord chat to see if they can either fix the issue or provide clarity on what is going on. I am hoping we can address all these issues.

paroj:
> ogre can walk past the edges of the map where is with the older version it would not.

this happens with all versions. There simply is no physics or collision detection in this sample at all.

guster32:
Good to know. Will dig deeper into shadow and fullscreen issues.

guster32:
@paroj The first thing I did to start looking into these issue was to start using emscripten version `1.38.24`. The first thing I noticed was that even when using the older version the the shadow issue is still present and that the fullscreen issue was fixed. This led me to conclude that:

1. The fullscreen issue is a emscripten issue and will continue to look into that
2. The Shadow issue is a regression on ogre3d.

So I then performed a bisect of the commits and I found the [commit](https://github.com/OGRECave/ogre/commit/59897664125e67e31cb46a47b22d7f7788532ecf) that triggered this issue. Would love to understand why this issue would only affect emscripten platform.  Would u be able to fix this issue? I will continue to look into the fullscreen issue next.

paroj:
> Would love to understand why this issue would only affect emscripten platform. Would u be able to fix this issue?

see above. Native GLES2 probably worked as it actually created a GLES3 context, while emscripten sticked to WebGL1.

guster32:
Nice! Thank you. I see this in the log too:
```
**************************************
*** OpenGL ES 2.x Renderer Started ***
**************************************
Shading language version: OpenGL ES GLSL ES 1.00 (WebGL GLSL ES 1.0 (OpenGL ES GLSL ES 1.0 Chromium))
``` 
Interesting... Do you think  we can light up WebGL 2 with this linker flags (info from [here](https://emscripten.org/docs/porting/multimedia_and_graphics/OpenGL-support.html))?   
``-s MIN_WEBGL_VERSION=2`` and `` -s MAX_WEBGL_VERSION=2``

paroj:
> -s MIN_WEBGL_VERSION=2 and -s MAX_WEBGL_VERSION=2

yes, thats what I have done here:
https://ogrecave.github.io/ogre/wasm/

guster32:
Turns out the API changed a bit... I was able to get the events to fire but for some reason the screen still does not render when in fullscreen (BlackScreen).  The new api now 
```
        emscripten_set_fullscreenchange_callback(EMSCRIPTEN_EVENT_TARGET_DOCUMENT, (void*)this, 1, &EmscriptenEGLWindow::fullscreenCallback);
        emscripten_set_webglcontextlost_callback(EMSCRIPTEN_EVENT_TARGET_DOCUMENT, (void*)this, 1, &EmscriptenEGLWindow::contextLostCallback);
        emscripten_set_webglcontextrestored_callback(EMSCRIPTEN_EVENT_TARGET_DOCUMENT, (void*)this, 1, &EmscriptenEGLWindow::contextRestoredCallback);
        emscripten_set_resize_callback(EMSCRIPTEN_EVENT_TARGET_DOCUMENT, (void*)this, 1, &EmscriptenEGLWindow::canvasWindowResized); 
```
According to the [documentation](https://emscripten.org/docs/api_reference/html5.h.html#how-to-use-this-api) the constants could be `EMSCRIPTEN_EVENT_TARGET_WINDOW` `EMSCRIPTEN_EVENT_TARGET_DOCUMENT` `EMSCRIPTEN_EVENT_TARGET_SCREEN` It also seems like there is a new way to trigger a fullscreen. It api has some really interesting [options](https://emscripten.org/docs/api_reference/html5.h.html#c.emscripten_request_fullscreen_strategy) That include Aspect Ration  and HiDPI screen handling.

guster32:
I also tried with WebGL 2... and emscripten 2.0.2 and as expected the shadow issue is is gone. But event the the new events I cant get fullscreen to render

paroj:
> According to the documentation the constants could be EMSCRIPTEN_EVENT_TARGET_WINDOW

yes, I also had that at some point and eventually could get fullscreen on firefox (not on chrome) w/o canvas resizing. I really should have backed up that state in a draft branch..

> Works on fullscreen for me... Was that build with 1.38.24 ?

yes

guster32:
When using the newer constants we run into this issue with the emscripten generated html. And as the issues describe below after hitting the fullscreen button I am now getting a canvas of width and height of 0. I if I resize the canvas manually on the browser's console rendering works as expected.
> The problem is that the fullscreen button uses the old Browser.requestFullscreen API and SDL uses the newer emscripten_request_fullscreen_strategy API. As mentioned in this issue, using the first API breaks the second. (emscripten-core/emscripten#3283)
Also, now that we use emscripten_set_canvas_element_size instead of emscripten_set_canvas_size, the old Browser.setCanvasSize never gets called which results in Browser.requestFullscreensetting the canvas size to 0.
https://github.com/emscripten-ports/SDL2/issues/8
https://github.com/emscripten-core/emscripten/issues/3283
Its too bad the autogenerated html is broken. I suppose we could create our own html.

paroj:
Nice. Thanks for digging into this for the root cause.

Although, I dislike carrying a custom html, because it might get outdated real quick, it is probably the easiest way forward.

On the plus side, we could make it more educational on how to embed JS Ogre into a larger application.

guster32:
Yes true.. I also found a workaround for this issue.
if we set 
```
emscripten_set_fullscreenchange_callback(EMSCRIPTEN_EVENT_TARGET_WINDOW, (void*)this, 1, &EmscriptenEGLWindow::fullscreenCallback);
```
And on the resize add:

```
....
void EmscriptenEGLWindow::resize(uint width, uint height)
    {
        mWidth = width;
        mHeight = height;
        

        EMSCRIPTEN_RESULT result = emscripten_set_canvas_element_size(mCanvasSelector.c_str(), width, height);
        // This is a workaroud for issue: https://github.com/emscripten-core/emscripten/issues/3283.
        // The setTimeout of 0 will ensure that this code is runs on the next JSEventLoop.
        EM_ASM(setTimeout(function(){var canvas = document.getElementById('canvas'); canvas.width = $0; canvas.height = $1;}, 0), width, height);
        
        if(result < 0)
        {
....
```
Fullscreen will work... I am happy to work on whichever route u prefer to go with. Just let me know.

guster32:
Ill update the pull request with the workaround for now.

paroj:
Im fine with shipping the workaround until the issue is fixed upstream.

Samples/Emscripten/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 set(MEDIA_DIR ${CMAKE_BINARY_DIR}/bin/media/)
 
 set(CMAKE_EXECUTABLE_SUFFIX ".html")
-set(CMAKE_EXE_LINKER_FLAGS "--preload-file ${MEDIA_DIR}/@. -s BINARYEN_TRAP_MODE=clamp -s EXPORTED_FUNCTIONS=\"['_passAssetAsArrayBuffer', '_clearScene', '_main']\" -s USE_SDL=2")
+set(CMAKE_EXE_LINKER_FLAGS "--preload-file ${MEDIA_DIR}/@. -s MIN_WEBGL_VERSION=2 -s MAX_WEBGL_VERSION=2 -s EXPORTED_FUNCTIONS=\"[ '_passAssetAsArrayBuffer', '_clearScene', '_main']\" -s USE_SDL=2")
```
paroj, line 4: please drop the WebGL2 flags, as we want to be compatible by default (Safari does not support it)

guster32: Sure.

PR #1707, opened 2020-09-13
Overlay: typedef DisplayString to String and hide UTFString

### LINKED ISSUE #28: [OGRE-135] Italic fonts rendered incorrectly
<b>[reporter="eventhorizon", created="Mon, 28 Jan 2013 22:29:14 +0100"]</b>

<p>I'm currently working with OGRE's font rendering system (finishing an app port, from the Crystal Space engine over to OGRE/Bullet), and am noticing that all italic font glyphs are being clipped on the right and leak into the previous glyph on the left. I've been debugging OgreFont.cpp and checking it against a dumped glyph map image, and getGlyphTexCoords() is returning texture coordinates that are off (U1 starts near the end of the previous glyph, and U2 stops before the end of the current glyph). I'm attaching an image of the same font rendered in both 

OGRE  
glyph_ogre

and CS  
glyph_cs

so you can see the problem. In the OGRE image, you can see the right edge of the 'B' glyph leaking into the image, while the right edge of the 'C' glyph is cut off.  So far it seems to happen in all italic fonts - don't know about others.</p>

<p>The new getCharacterSpacer feature doesn't solve this issue when adjusting, and only affects the leaked glyph on the left side (no change with the right-side clipping).  This happens in all platforms, but the pics were taken using OGRE 1.8.1 on Linux x64 (Debian Squeeze).  This seems to happen on all italic fonts, but I can attach the testing one if you need.</p>

### LINKED ISSUE #1705: DisplayString API abuse
DisplayString/ UTFString should be only used when the decoded string is needed - i.e. when constructing the font texture. Public API should use String.

### LINKED ISSUE #1779: optionally use stb_truetype instead of freetype

### LINKED ISSUE #1859: Use branchless utf-8 decoder
see
https://github.com/ocornut/imgui/blob/58075c4414b985b352d10718b02a8c43f25efd7c/imgui.cpp#L1571

paroj:
the decoded string is only needed inside TextAreaOverlayElement

fixes #1705
fixes #1859
fixes #28
fixes #1779


paroj:
expose `vector<uint32> utftoc32(const String& str)`

PR #1711, opened 2020-09-18
Fix Typo in GpuNamedConstantsSerializer::exportNamedConstants

sercero:
There was a typo wich would lead to misleading OGRE Exception text.


paroj:
thanks

PR #1720, opened 2020-09-27
Substitute postfix iterator increment with prefix iterator increment

akacoding:
I searched and replaced the postfix iterator increments with the prefix version because the temporaries are not needed and cannot be optimized away.
Probably it's a micro optimization but it doesn't harm.
See https://stackoverflow.com/questions/24901/is-there-a-performance-difference-between-i-and-i-in-c for reference.

### DIFF
--- a/Components/Bites/src/OgreApplicationContextBase.cpp
+++ b/Components/Bites/src/OgreApplicationContextBase.cpp
@@ -428,7 +428,7 @@ void ApplicationContextBase::locateResources()
         Ogre::ConfigFile::SettingsMultiMap::const_iterator i;
 
         // go through all resource paths
-        for (i = settings.begin(); i != settings.end(); i++)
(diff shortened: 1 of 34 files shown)


paroj:
> Probably it's a micro optimization but it doesn't harm.

I doubt that it is a optimization at all:
https://godbolt.org/z/ave6KG

even at the lowest optimization level, gcc can optimize everything away.

akacoding:
> I doubt that it is a optimization at all:
> https://godbolt.org/z/ave6KG

Yes you're obviously right. It should had come to my mind to check it in assembler first.
It's up to you what to do with the pull request. You can decline it if you want.

paroj:
yes, this is currently just noise.

A more valuable change would be the switch to range based for loops. This would not only improve readability, but also spare one pointer redirection:
https://godbolt.org/z/hj8v9b

PR #1723, opened 2020-09-29
destroyInstance method shoud be marked virtual

balrogcpp:
Overriding of destroyInstance of class MovableObjectFactory is used in OgreOggSound. And what is point to keep it non-virtual when user must define creation logic in createInstanceImpl anyway?

### DIFF
--- a/OgreMain/include/OgreMovableObject.h
+++ b/OgreMain/include/OgreMovableObject.h
@@ -608,7 +608,7 @@ namespace Ogre {
             const String& name, SceneManager* manager, 
             const NameValuePairList* params = 0);
         /** Destroy an instance of the object */
-        void destroyInstance(MovableObject* obj) { delete obj; }
+        virtual void destroyInstance(MovableObject* obj) { delete obj; }
 
         /** Does this factory require the allocation of a 'type flag', used to 
             selectively include / exclude this type from scene queries?


paroj:
> And what is point to keep it non-virtual when user must define creation logic in createInstanceImpl anyway?

because it provides the only sane implementation?

if it would have been non-virtual from the start, that abomination in OgreOggSound probably could have been avoided. But as it stands I will have to merge this..

balrogcpp:
**delete obj** will not run derived class destructor. So question is, where to put custom destruction logic then? Maybe optional destroyInstanceImpl method will be better?

paroj:
> delete obj will not run derived class destructor. 

it will:
https://github.com/OGRECave/ogre/blob/b03c83ebdd64cfedc749ccaaf5310b617f721de7/OgreMain/include/OgreMovableObject.h#L174

balrogcpp:
Indeed it will call, my fault :) But putting code from destroyInstance to destructor cause segfault in OgreOggSoundFactory. Thank you for support

paroj:
see https://github.com/OGRECave/ogre-audiovideo/commit/1e3ef74def704f0a26d4d937a8e60f7842bffbd7

balrogcpp:
tested and all works fine with v1.12.9

PR #1724, opened 2020-09-29
Added cast from OverlayElement to OverlayContainer

ChristinaDickmannORGADATA:
I needed to cast from OverlayElement to OverlayContainer to recreate this tutorial: http://wiki.ogre3d.org/Creating+Overlays+via+Code


paroj:
thanks

PR #1729, opened 2020-10-04
Devel

paroj:


OgreMain/src/OgreMovableObject.cpp
```diff
@@ -292,6 +289,14 @@ namespace Ogre {
         // fallback
         return Affine3::IDENTITY;
     }
+
+    Real MovableObject::getBoundingRadiusScaled() const
+    {
+        const Vector3& scl = mParentNode->_getDerivedScale();
+        Real factor = std::max(std::max(scl.x, scl.y), scl.z);
+        return getBoundingRadius() * factor;
```
akacoding, line 297: I know it's just moved code. But what if scaling is negative? Is it okay that the factor is negative. I think it should be rare, but from code it's possible.

paroj: yeah.. lets throw in an abs, while we are at it. Generally you should avoid negative scale though, as it inverts your vertex winding. But Ogre tries to fix that, so we can consider it here too.

PR #1736, opened 2020-10-10
prototype sample improves cpu perf by at least 30%.

creadmefford:
obviously this is just for review and not to actually merge.

i compile this in my particle editor, not as part of ogre so i add a '2' to everything to avoid clashes.  (slightly cleaned up the '2's to reduce file diff noise for review purposes.)

none of this is cleaned up, just the min to take perf measurements.  i also don't think i'd include the per-particle width/height as you could always create multiple emitters of varying size.

main classes to look at:
Ogre::Particles, Ogre::ParticleSystem, Ogre::ParticleEmitter, Ogre::PointEmitter.

### DIFF
--- a/OgreMain/include/OgreParticles.h
+++ b/OgreMain/include/OgreParticles.h
@@ -1,0 +1,90 @@
+/*
+-----------------------------------------------------------------------------
+This source file is part of OGRE
+(Object-oriented Graphics Rendering Engine)
+For the latest info, see http://www.ogre3d.org/
+
+Copyright (c) 2000-2014 Torus Knot Software Ltd
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-----------------------------------------------------------------------------
+*/
+#ifndef __Particles2_H__
+#define __Particles2_H__
+
+#include "OgrePrerequisites.h"
+#include "OgreHeaderPrefix.h"
+#include "OgreColourValue.h"
+#include "OgreMath.h"
+#include "OgreVector.h"
+
+#include "sh/cpp/block_n.hpp"
+
+namespace Ogre {
+
+
+    struct ParticleTtl
+    {
+        Real remaining;
+        Real total;
+    };
+
+    struct ParticleTranslation
+    {
+        Vector3 position;
+        Vector3 direction;
+    };
+
+    struct ParticleRotation
+    {
+        Real angle;
+        Real speed;
+    };
+
+    struct ParticleSize
+    {
+        Real width;
+        Real height;
+    };
+
+    struct Particles2 : public FXAlloc, private block_n <ParticleTtl, ParticleTranslation, ParticleRotation, ColourValue, ParticleSize>
+    {
+        Particles2 ();
+        ~Particles2 ();
+
+        unsigned size;
+        unsigned active;
+
+        ParticleTtl* ttls;
+        ParticleTranslation* translations;
+        ParticleRotation* rotations;
+        ColourValue* colors;
+        ParticleSize* sizes;
+
+        void clear ();
+        void create (unsigned count, unsigned& start, unsigned& end);
+        void resize (unsigned newsize);
+        void swap (unsigned idx1, unsigned idx2);
+    };
+}
+
+#include "OgreHeaderSuffix.h"
+
+#endif
+
--- a/OgreMain/src/OgreParticleSystem.cpp
+++ b/OgreMain/src/OgreParticleSystem.cpp
@@ -201,15 +201,15 @@ namespace Ogre {
         mEmitters.clear();
     }
     //-----------------------------------------------------------------------
-    ParticleAffector2* ParticleSystem::addAffector(const String& affectorType)
+    ParticleAffector* ParticleSystem::addAffector(const String& affectorType)
     {
-        ParticleAffector2* af = 
-            ParticleSystemManager2::getSingleton()._createAffector(affectorType, this);
+        ParticleAffector* af = 
+            ParticleSystemManager::getSingleton()._createAffector(affectorType, this);
         mAffectors.push_back(af);
         return af;
     }
     //-----------------------------------------------------------------------
-    ParticleAffector2* ParticleSystem::getAffector(unsigned short index) const
+    ParticleAffector* ParticleSystem::getAffector(unsigned short index) const
     {
         assert(index < mAffectors.size() && "Affector index out of bounds!");
         return mAffectors[index];
@@ -224,7 +224,7 @@ namespace Ogre {
     {
         assert(index < mAffectors.size() && "Affector index out of bounds!");
         ParticleAffectorList::iterator ai = mAffectors.begin() + index;
-        ParticleSystemManager2::getSingleton()._destroyAffector(*ai);
+        ParticleSystemManager::getSingleton()._destroyAffector(*ai);
         mAffectors.erase(ai);
     }
     //-----------------------------------------------------------------------
@@ -233,7 +233,7 @@ namespace Ogre {
         // DON'T delete directly, we don't know what heap these have been created on
         for (auto a : mAffectors)
         {
-            ParticleSystemManager2::getSingleton()._destroyAffector(a);
+            ParticleSystemManager::getSingleton()._destroyAffector(a);
         }
         mAffectors.clear();
     }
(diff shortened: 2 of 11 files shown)


OgreMain/include/OgreParticles.h
```diff
@@ -0,0 +1,90 @@
...
+        Real speed;
+    };
+
+    struct ParticleSize
+    {
+        Real width;
+        Real height;
+    };
+
+    struct Particles2 : public FXAlloc, private block_n <ParticleTtl, ParticleTranslation, ParticleRotation, ColourValue, ParticleSize>
```
creadmefford, line 66: this is really the main difference here

OgreMain/src/OgreParticleSystem.cpp
```diff
@@ -549,62 +500,37 @@ namespace Ogre {
...
+        if (!mLocalSpace)
         {
-            // Create a new particle & init using emitter
-            // The particle is a visual particle if the emit_emitter property of the emitter isn't set 
-            Particle* p = 0;
-            String  emitterName = emitter->getEmittedEmitter();
-            if (emitterName.empty())
-                p = createParticle();
-            else
-                p = createEmitterParticle(emitterName);
```
paroj, line 569: this prevents you from creating "Fireworks"-style systems (as in the ParticleFX sample).

OgreMain/include/OgreParticles.h
paroj, line 66: This class is quite close to Ogre::VertexData...

OgreMain/src/OgreParticleSystem.cpp
creadmefford: i am planning to add the emitter emitters back in -- this was just far enough to test out the standard perf.

OgreMain/include/OgreParticles.h
creadmefford: so you're saying move the data to a vertex buffer instead of cpu?

paroj: If you use a shadow buffer, you can ensure that the data will be read/ writable on the CPU. This is a thing I am still figuring out though.
This Particles2 class is a nicer API for that shadow buffer in a sense. Also this interferes with the ParticleRenderer that currently owns vertex buffers and takes care of quad creation.
With sv/gl_vertexid one could do that on the GPU though..

OgreMain/src/block_n.hpp
```diff
@@ -0,0 +1,219 @@
...
+    auto mem = malloc (bytes);
+    auto moveable = std::min (size, newsize);
+    block_relocate_memcpyable (newsize, moveable, mem, t1, args...);
+    block_free (*t1); // free the old allocation
+    size = newsize;
+    block_assign (size, mem, t1, args...);
+}
+
+template <typename... args_t>
+struct block_n
```
paroj, line 208: is this any different than `std::vector<std::tuple<args...>>`?

creadmefford: i think the vector<tuple> approach is similar although i'm not sure of the exact syntax.  for one, you'd be referring to indexes (ex. 0, 1, 2) instead of variable names (ex. 'TTLs').  1. block_n uses a single allocation for all of the arrays and therefore all of the arrays are ajacent in memory.  i.e. the TTLs array is immediatly followed by the translations array which is followed by the rotations data... ex. <TTL><TTL><TTL><POS><POS><POS><ROT><ROT><ROT> etc.  so when particle "expire" loop is called, it decrements all the TTLS.  then each <POS> is updated when velocity is applied, followed by each of the <ROT> etc etc. 2. block_n does a memcpy instead of std::move() which for classes (ex. Vector3) ensures ctors() and dtors() are called which we don't care about in this case.  3. vector slightly larger footprint.  4. i use block_n for my ECS classes and the particle work is a rapid prototype so it was easier to use.  (note if you remove the "_memcpyable" then block_n supports move operations instead of memcpy).

OgreMain/src/OgreParticleSystem.cpp
creadmefford: a better design for emitting emitters is to create a new emitter (say EmitEmitter) in fx rather than include this code in OgreParticleSystem in OgreMain.  if you pollute the OgreParticleSystem class with "emitting emitters" you add unnecessary complexity and decrease the performance of all particle types even when you're not using emitting emitters.  you should only take the hit for emitting emitters if you're using them.

creadmefford: ogre needs a "technique" concept like particle universe (which is feature rich but too complicated to refactor).  internally i'm calling mine an "effect"... which allows a script to include and combine multiple ParticleSystems into a single visual effect.  

this allows you to mix techniques and not require the core particle design to be overly complicated.

for example, rather than allowing individual particles to be of any variety of sizes (which bloats memory, adds more code paths per frame, etc), you just create an "effect" which is comprised of multiple ParticleSystems that each use their own size... so you're only paying for what you use and not punishing the particle system even when not using this feature.

OgreMain/src/block_n.hpp
creadmefford: the debug drawing API refactoring is an *excellent* example of what needs to be done everywhere in ogre.
i..e. pulling out unrelated parts and keeping them separate as to avoid polluting the runtime code.  for example, the particle systems, emitters, affector runtime code should be separate from the script parsing, particle system manager templates as singletons, etc.

OgreMain/include/OgreParticles.h
creadmefford: i'm still using directx 9 so i really need to get up to date on shareable buffers etc.  same with the sv/gl_vertexid concept.

OgreMain/src/block_n.hpp
creadmefford: (and i think this is my issue with unreal, unity, godot, ogre 2.x+ is that it feels tightly coupled.  it's beautiful and super fast, but i'd rather mix and match what i get, and have a loosely coupled system where i pick and choose from what i want)

OgreMain/src/OgreParticleSystem.cpp
creadmefford: also note the +30% perf improvement is on a single particle system with a single emitter and no affectors.  here's torchlight 2's blood splatter particle effect:  (blood shoots everywhere, chunks bounce off the ground, etc...)

image

each line in this "particle" is a single particle system.  

here is the same effect expanded but not showing everything:

image

imagine the savings for each as all these affectors again and again and again iterate through the linked lists of large old school Ogre::Particle structs which are huge.  the cache is trashed over and over and over.

OgreMain/src/block_n.hpp
paroj: > we could move the entire legacy particles to the FX plugin.

that would be a "ParticleFX" component, as you typically do not use any Plugin specific API. The current PluginFX implementation provides affectors and emitters that are only exposed through the common StringInterface property API.
This is also why Particle is needed in Main - to define the common API to be used by Plugins.

> that would be FX3 plugin and the old FX and FX2 would just never get upgrades anymore.

the thing is, that nobody likes updating their **working code** - no matter what the benefit is. You said yourself, you would like to have the particle thing cleared up in order to never have to thing about it again.
Imagine having to port it to FX2, with a FX3 at the horizon.

However, I am quite confident that we can come up with an *public* API that will suit a FX3 implementation at lets say 90% of the optimal performance.
The key ingredient is probably just switching from list<Particle*> to vector<Particle>. With the latter iteration is fast and we can just memcpy it (which is cheap) to a hardware buffer or whatever will be needed down the road.

OgreMain/src/OgreParticleSystem.cpp
paroj: > if you pollute the OgreParticleSystem class with "emitting emitters" you add unnecessary complexity and decrease the performance of all particle types even when you're not using emitting emitters.

yes, the current implementation is bad. However, I think if you let observe emitters a particle for positioning instead of making them one, you can make "emitted emitters" free if you do not use them, while still having them in the API.

> for example, rather than allowing individual particles to be of any variety of sizes 

individual particle size just takes 50% of individual particle colour or +16% of (position + velocity + colour + ttl) and it allows non-uniform scale affectors. I would not say it is that bad.

> imagine the savings for each as all these affectors again and again and again

I get what you are aiming at, but it is still 30% :wink:

OgreMain/include/OgreParticles.h
paroj: see e.g. https://webgl2fundamentals.org/webgl/lessons/webgl-drawing-without-data.html

creadmefford: awesome, will take a look.  thanks!

OgreMain/src/block_n.hpp
creadmefford: i was imagining you would update your plugin.cfg to use the particles you want and set the #include headers to look in the correct folder.  i.e. just a compiler and runtime configuration change, not break existing code.

paroj: >  set the #include headers to look in the correct folder.

if you #include the plugin headers, you have to actually link against the plugin - at which point it is no longer a plugin, but an ordinary library. Thats why I would rather call it "Component".

creadmefford: ah gotcha. makes sense.

OgreMain/src/OgreParticleSystem.cpp
creadmefford: got the perf up to a 50%+ improvement.  i had to copy/paste Ogre::BillboardSet and manually "unroll" the injectBillboard() function because it's being called now in a tight loop and it frequently accesses member variables.  the members do not appear to vary across the tight loop but i imagine the compiler can't make that assumption.  i wound up pulling up all the "options" and query them up front, then pass down pointers to member variables based on the decision.

i'll stop here... i'm happy with a 2x particle budget increase.  

thanks for all the help!

creadmefford: for example, i first check to see how the vertexes should be created:
        void (Ogre::BillboardSetFx::*genVerticesFn)(const Ogre::Vector3* const offsets, const Ogre::Billboard & pBillboard) = nullptr;
        auto vertexPolicy = mBillboardSet->getVertexPolicy ();
        if (vertexPolicy == Ogre::BillboardSetFx::rotation)
            genVerticesFn = &Ogre::BillboardSetFx::genVerticesRotation;
        else if (vertexPolicy == Ogre::BillboardSetFx::vertex)
            genVerticesFn = &Ogre::BillboardSetFx::genVerticesVertex;
        else if (vertexPolicy == Ogre::BillboardSetFx::point)
            genVerticesFn = &Ogre::BillboardSetFx::genVerticesPoint;
and then pass this function down into injectBillboard().

creadmefford: it's ugly but sometimes i'm getting close to 60fps again with 100k particles @4k resolution.

creadmefford: lastly, in my runtime i'll allocate the particle systems in the same block of memory as the affectors and emitters.

paroj:
if you want 10% more and can live with point billboards, make sure you enable point sprite rendering as in the Examples/GreenyNimbus particle_system.

creadmefford:
> if you want 10% more and can live with point billboards, make sure you enable point sprite rendering as in the Examples/GreenyNimbus particle_system.

oh nice, i will check that out, thanks!!!

PR #1740, opened 2020-10-13
custom billboard particle renderers also need to access Billboard variables.

creadmefford:
or is there another approach i should take?


paroj:
you can use the `get*` methods

PR #1786, opened 2020-11-22
Always include shadow casting lights in local list

balrogcpp:
Culling of shadow casting lights from SceneNode's local list cause mismatch with frame shadow texture list. This breaks a texture shadowing if there are at least two spotlights that casts shadows in the scene. Here is small fix for this.

### DIFF
--- a/OgreMain/src/OgreSceneManager.cpp
+++ b/OgreMain/src/OgreSceneManager.cpp
@@ -364,7 +364,7 @@ void SceneManager::_populateLightList(const Vector3& position, Real radius,
         // Calc squared distance
         lt->_calcTempSquareDist(position);
 
-        if (lt->getType() == Light::LT_DIRECTIONAL)
+        if (lt->getType() == Light::LT_DIRECTIONAL || (lt->getCastShadows() && isShadowTechniqueTextureBased()))
         {
             // Always included
             destList.push_back(lt);
--- a/OgreMain/src/OgreSceneManager.cpp
+++ b/OgreMain/src/OgreSceneManager.cpp
@@ -364,7 +364,7 @@ void SceneManager::_populateLightList(const Vector3& position, Real radius,
         // Calc squared distance
         lt->_calcTempSquareDist(position);
 
-        if (lt->getType() == Light::LT_DIRECTIONAL || (lt->getCastShadows() && isShadowTechniqueIntegrated()))
+        if (lt->getType() == Light::LT_DIRECTIONAL)
         {
             // Always included
             destList.push_back(lt);
@@ -376,7 +376,7 @@ void SceneManager::_populateLightList(const Vector3& position, Real radius,
             {
                 destList.push_back(lt);
             }
-        }
+      }
     }
 
     // Sort (stable to guarantee ordering on directional lights)
@@ -398,14 +398,14 @@ void SceneManager::_populateLightList(const Vector3& position, Real radius,
         std::stable_sort(destList.begin(), destList.end(), lightLess());
     }
 
+    //Skip this, light indexes updated in findLightsAffectingFrustum
+
     // Now assign indexes in the list so they can be examined if needed
-    size_t lightIndex = 0;
-    for (LightList::iterator li = destList.begin(); li != destList.end(); ++li, ++lightIndex)
-    {
-        (*li)->_notifyIndexInFrame(lightIndex);
-    }
-
-
+    //size_t lightIndex = 0;
+    //for (LightList::iterator li = destList.begin(); li != destList.end(); ++li, ++lightIndex)
+    //{
+    //    (*li)->_notifyIndexInFrame(lightIndex);
+    //}
 }
 //-----------------------------------------------------------------------
 void SceneManager::_populateLightList(const SceneNode* sn, Real radius, LightList& destList, uint32 lightMask) 
@@ -2203,7 +2203,42 @@ void SceneManager::renderSingleObject(Renderable* rend, const Pass* pass,
             {
                 pLightListToUse = &rendLightList;
             }
-            lightsLeft = 0;
+
+          if (isShadowTechniqueIntegrated() && mLightsDirtyCounter > 0) {
+            size_t shadowTexIndex = mShadowRenderer.getShadowTexIndex(lightIndex);
+            unsigned short numShadowTextureLights = 0;
+
+            for (const auto it : *pLightListToUse) {
+              // potentially need to update content_type shadow texunit
+              // corresponding to this light
+              size_t textureCountPerLight = mShadowRenderer.mShadowTextureCountPerType[it->getType()];
+              size_t textureCounterPerLight = 0;
+
+              for (size_t j = 0; j < textureCountPerLight && shadowTexIndex < mShadowRenderer.mShadowTextures.size(); ++j) {
+                // link the numShadowTextureLights'th shadow texture unit
+                ushort tuindex = pass->_getTextureUnitWithContentTypeIndex(TextureUnitState::CONTENT_SHADOW, numShadowTextureLights);
+                if (tuindex > pass->getNumTextureUnitStates()) break;
+
+                TextureUnitState *tu = pass->getTextureUnitState(tuindex);
+
+                //pick up correct shadow texture index by light's frame index + texture counter per light
+                const TexturePtr &shadowTex = mShadowRenderer.mShadowTextures[mShadowRenderer.mShadowTextureIndexLightList[it->_getIndexInFrame()] + textureCounterPerLight];
+                tu->_setTexturePtr(shadowTex);
+
+                Camera *cam = shadowTex->getBuffer()->getRenderTarget()->getViewport(0)->getCamera();
+                tu->setProjectiveTexturing(!pass->hasVertexProgram(), cam);
+                mAutoParamDataSource->setTextureProjector(cam, numShadowTextureLights);
+                ++numShadowTextureLights;
+                ++shadowTexIndex;
+                ++textureCounterPerLight;
+                // Have to set TU on rendersystem right now, although
+                // autoparams will be set later
+                mDestRenderSystem->_setTextureUnitSettings(tuindex, *tu);
+              }
+            }
+          }
+
+          lightsLeft = 0;
         }
 
         // issue the render op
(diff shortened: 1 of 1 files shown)


paroj:
this effectively disables per-object light culling when texture shadows are enabled.
As far as I can tell, this is permitted and you can handle it by reading ACT_LIGHT_COUNT.

Can you explain your specific problem in more detail? The SkeletalAnimation Sample uses the light setup you describe without issues.

balrogcpp:
Good that you mentioned SkeletalAnimation Sample -- I tried to reproduce issue by modifying of sample and failed. 

Difference can be somewhere in material -- I use integrated shadow with single pass, and iterating inside ACT_LIGHT_COUNT. Well, lightning works perfect with ACT_LIGHT_POSITION_ARRAY[i] and other arrays etc. but shadow textures breaks when one of the lights is culled out.

balrogcpp:
So I am closing this request -- seems like problem is somewhere else.

paroj:
the sample uses the built-in shadows which are rendered multi-pass by rendering a shadow "overlay" and then multiplying it into the frame-buffer.
There might be very well a bug when using integrated shadows. Maybe the shadowmaps are not re-ordered properly when the respective light is culled?

balrogcpp:
I thought same way, so putted light reordering in renderSingleObject when light dirty flag is up. Also frame light index was updated per object, now it's per frame.

After this shadows works as it should, at least I didn't find anything during testing with all light combination. Of course patch can still have some issues, but in general idea with shadow texture reordering works.

OgreMain/src/OgreSceneManager.cpp
```diff
@@ -398,14 +398,14 @@ void SceneManager::_populateLightList(const Vector3& position, Real radius,
...
-    for (LightList::iterator li = destList.begin(); li != destList.end(); ++li, ++lightIndex)
-    {
-        (*li)->_notifyIndexInFrame(lightIndex);
-    }
-
+    //Skip this, light indexes updated in findLightsAffectingFrustum
 
+    // Now assign indexes in the list so they can be examined if needed
+    //size_t lightIndex = 0;
+    //for (LightList::iterator li = destList.begin(); li != destList.end(); ++li, ++lightIndex)
```
paroj, line 405: please do not leave commented out code behind. it does not help anyone

OgreMain/src/OgreSceneManager.cpp
```diff
@@ -2203,7 +2203,48 @@ void SceneManager::renderSingleObject(Renderable* rend, const Pass* pass,
...
+
+                        TextureUnitState *tu = pass->getTextureUnitState(tuindex);
+
+                        //pick up correct shadow texture index by light's frame index + texture counter per light
+                        const TexturePtr &shadowTex = mShadowRenderer.mShadowTextures[mShadowRenderer.mShadowTextureIndexLightList[it->_getIndexInFrame()] + textureCounterPerLight];
+                        tu->_setTexturePtr(shadowTex);
+
+                        Camera *cam = shadowTex->getBuffer()->getRenderTarget()->getViewport(0)->getCamera();
+                        tu->setProjectiveTexturing(!pass->hasVertexProgram(), cam);
+                        mAutoParamDataSource->setTextureProjector(cam, numShadowTextureLights);
```
paroj, line 2236: this part of the loop looks familiar. Can we merge it with the existing loop to reduce complexity?

paroj:
we are currently have issues with the CI, so I will have to sort them out first.

OgreMain/src/OgreSceneManager.cpp
balrogcpp: ok, deleted

OgreMain/src/OgreSceneManager.cpp
balrogcpp: This have to be investigated. Loop goes completely different depending on iteratePerLigh bool

paroj:
ok, CI is running again.Time to get back to this.

Could you elaborate on what exact issue you are fixing here and how? e.g.

> previously the data was like:
> - shadow textures: [tex0, tex1, tex2]
> - light matrices: [light0, light2]
> 
> goal is:
> - shadow textures: [tex0, tex2]
> - light matrices: [light0, light2]
> 
> this is fixed by searching X in Y.

while I kind of get that the "magic" is in `//pick up correct shadow texture index..`, I need to check how this affects existing code and how it fits into the big picture.

Maybe the proper fix would be adding a `ACT_LIGHT_NUMBER_ARRAY` (c.f. ACT_LIGHT_NUMBER)

paroj:
based on [sortLightsAffectingFrustum](https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_scene_manager_1_1_listener.html#a5ef1be858fe331867653fbb8aa819494), I think your initial patch was correct

paroj:
@balrogcpp does https://github.com/OGRECave/ogre/pull/1792/commits/5bf61876674a4087d0975101aaef1030d67773e2 fix the issue as well?

balrogcpp:
https://github.com/OGRECave/ogre/commit/cf25cf280f368d5f98b8bfc63d98507eac55a02b is working as it is just more elegant way to disable culling.

But it will affect performance, isn't it? For some objects light that was previously culled out now will be calculated. That's why reassign of shadowmaps may be preferable.

balrogcpp:
About fix: each light contains it's own "index" which is same as ACT_LIGHT_NUMBER. And same value is used in mShadowRenderer.mShadowTextureIndexLightList array to check what shadow texture (first one) is corresponding to which light index.

So here is magic -- 

mShadowRenderer.mShadowTextures[mShadowRenderer.mShadowTextureIndexLightList[it->_getIndexInFrame()] + textureCounterPerLight] 

is piking up texture with help of mShadowTextureIndexLightList array, that couple light index with texture index.

balrogcpp:
One more issue with that -- light index was updated per object, when by description it should be updated by frame. So I moved update code to findLightsAffectingFrustum that is fired when lights are culled out for frame.

paroj:
> light index was updated per object, when by description it should be updated by frame

where did you get that from? _getIndexInFrame doc says that it can very per object.

paroj:
>  is working as it is just more elegant way to disable culling.

yes, as the documented behavior is that shadow casting lights are only culled per frame, but not per object

> But it will affect performance, isn't it? For some objects light that was previously culled out now will be calculated. That's why reassign of shadowmaps may be preferable.

actually, performance will improve as it is much cheaper to waste some lighting calculations in the pixel shader then re-binding the shadow textures: https://computergraphics.stackexchange.com/questions/37/what-is-the-cost-of-changing-state/46#46

This is of course not true when you have a ton of shadow casting spot lights. But then a GPU culling technique would be preferable anyway. see e.g. http://www.aortiz.me/2018/12/21/CG.html

balrogcpp:
> > light index was updated per object, when by description it should be updated by frame
> 
> where did you get that from? _getIndexInFrame doc says that it can very per object.

Hmm then I can't get what is it for. 

From description of ACT_LIGHT_NUMBER: This binding provides the global light index for a local index. So I expected it can be used to map local light indexes (per object) to index that used in frame. And I didn't find any uses of _getIndexInFrame in OgreMain, only in AutoParamDataSource when passing to shader.

balrogcpp:
As you have proof that per-frame culling for shadow casters is faster then lets stand for it.

balrogcpp:
To dot the i I will compare fps in my scene. Performance also depends on how often lightsDirtyCounter is up, as only then shadowmaps update needed, when GPU culling many times per second.

balrogcpp:
Indeed, when lightsDirtyCounter is not changed culling technique give significant performance drop. Scene from screenshot drops from 750 fps (average) to 550 fps. Because in static scene lightsDirtyCounter is never up and so shadows textures are not rebinded.

Two spot lights in scene, one is behind the camera. Ambient light is off.

image

paroj:
> > > light index was updated per object, when by description it should be updated by frame
> > 
> > 
> > where did you get that from? _getIndexInFrame doc says that it can very per object.
> 
> Hmm then I can't get what is it for.
> 
> From description of ACT_LIGHT_NUMBER: This binding provides the global light index for a local index. So I expected it can be used to map local light indexes (per object) to index that used in frame. And I didn't find any uses of _getIndexInFrame in OgreMain, only in AutoParamDataSource when passing to shader.

the old manual gives some more details on how this was intended (c.f. light_number):
https://www.ogre3d.org/docs/manual18/manual_23.html

I agree that storing the object lights index there does not make much sense, as one can easily derive that from pass_iteration_number.

I will enqueue changing that to the frame index for 1.13 - to avoid breaking existing code for 1.12 users (although I doubt anyone is using that ACT right now)

paroj:
> Indeed, when lightsDirtyCounter is not changed culling technique give significant performance drop. Scene from screenshot drops from 750 fps (average) to 550 fps. Because in static scene lightsDirtyCounter is never up and so shadows textures are not rebinded.
> 
> Two spot lights in scene, one is behind the camera. Ambient light is off.

you deliberately constructed the worst case there, did you? Two tiny spot lights and a screen-filling terrain result in 50% wasted lighting computations for all pixels.

However, if you move the second light further away (or reduce its attenuationRange) it would get culled on the frustum level. That should give you an even greater performance boost, as the according shadowmap will not be updated either.

Also, using keeping the camera static to avoid bumping lightsDirtyCounter, is not very realistic. In this case you could just not call renderOneFrame() to get infinite performance.

Therefore, I think the performance difference you measured is not actually *relevant*.

balrogcpp:
> > Indeed, when lightsDirtyCounter is not changed culling technique give significant performance drop. Scene from screenshot drops from 750 fps (average) to 550 fps. Because in static scene lightsDirtyCounter is never up and so shadows textures are not rebinded.
> > Two spot lights in scene, one is behind the camera. Ambient light is off.
> 
> you deliberately constructed the worst case there, did you? Two tiny spot lights and a screen-filling terrain result in 50% wasted lighting computations for all pixels.
> 
> However, if you move the second light further away (or reduce its attenuationRange) it would get culled on the frustum level. That should give you an even greater performance boost, as the according shadowmap will not be updated either.
> 
> Also, using keeping the camera static to avoid bumping lightsDirtyCounter, is not very realistic. In this case you could just not call renderOneFrame() to get infinite performance.
> 
> Therefore, I think the performance difference you measured is not actually _relevant_.

Yes, this is the worst case. In some middle case difference will be smaller. But how often dirty flag is up? in code it is updated only when light list is changed, should be no more then once per second. Scripted benchmark may help figure out real performance difference.

In conclusion -- right now I am okay with https://github.com/OGRECave/ogre/commit/cf25cf280f368d5f98b8bfc63d98507eac55a02b :) It solves problem and in average does not cause significant performance penalty.

paroj:
> Yes, this is the worst case. In some middle case difference will be smaller. But how often dirty flag is up? in code it is updated only when light list is changed, should be no more then once per second.

this depends on the camera speed and light topology :wink: 

anyway.. closing as the fix was deployed with #1792

PR #1812, opened 2020-12-20
Add implot v0.8 to ogreimgui 

kenkit:
To use implot in ogre you need to 
build ogre with 

`-DOGRE_BUILD_COMPONENT_IMGUI_IMPLOT=1`

Then when you build your app 

add the following in your CmakeLists.txt 
`
add_compile_definitions(OGRE_BUILD_COMPONENT_IMGUI_IMPLOT=1) 
`
or 

`#define OGRE_BUILD_COMPONENT_IMGUI_IMPLOT 1`


kenkit:
There is still more to fix, don't merge yet

paroj:
why are these changes to ogre needed to use implot? It seems to be trivially usable in a client application as is.

kenkit:
> why are these changes to ogre needed to use implot? It seems to be trivially usable in a client application as is.

Even with those changes I keep getting an error that I did not call ``CreateContext();```
I need some help @paroj  to get it working. I tried using it without adding it to ogreoverlay but I still got the same problem even after adding it into ogreoveraly.

kenkit:
Let me try cleaning my environment and build again, maybe some left around stale files might have made it not work.

kenkit:
@paroj with current configuration I link to OGRE_LIBRARIES In cmake and yet I still get 

```
:78: undefined reference to `ImPlot::EndPlot()'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/research_app.dir/build.make:113: research_app] Error 1
make[1]: *** [CMakeFiles/Makefile2:124: CMakeFiles/research_app.dir/all] Error 2
make: *** [Makefile:103: all] Error 2
```

kenkit:
When I run it afte rmanually adding implots clasess to my app 
```
research_app: /home/aiden/Desktop/DATA/programming/ogre/build/implot-0.8/implot.cpp:1224: bool ImPlot::BeginPlot(const char*, const char*, const char*, const ImVec2&, ImPlotFlags, ImPlotAxisFlags, ImPlotAxisFlags, ImPlotAxisFlags, ImPlotAxisFlags): Assertion `(GImPlot != __null) && "No current context. Did you call ImPlot::CreateContext() or ImPlot::SetCurrentContext()?"' failed.
```

epezent:
I don't see any obvious mistakes in your implementation. Are you confident that `OGRE_BUILD_COMPONENT_IMGUI_IMPLOT` is getting defined? If not, `ImPlot::CreateContext()` might not compile. The other thing that might cause this is if ImGui and/or ImPlot are linked dynamically. However, it appears that you are adding the source files directly to the OGRE build. Is that correct?

kenkit:
@epezent  I've tried both ways even linking the sources directly to my app and it still didn't work. It crushed somewhere within imgui

kenkit:
I even tried removing the ifdef macros just to be sure

kenkit:
I also made sure to check if the implot symbols were present in the resulting libraries and they were present.

kenkit:
In your app you must add 
```cpp
ImPlot::SetCurrentContext(imguiOverlay->get_implot());```

kenkit:
Screenshot_2021-03-09_14-57-26
And this is the result I wanted

PR #1825, opened 2020-12-29
fix spelling errors reported by lintian

simonschmeisser:
https://games-team.pages.debian.net/-/ogre/-/jobs/1292553/artifacts/debian/output/lintian.html

complains about a few typos in strings. I couldn't find all of them but also found a few additional occurrences in variable names etc.


paroj:
:+1: thanks

PR #1841, opened 2021-01-05
gl.h and glext.h are unused and under non-free license

simonschmeisser:
This is an experimental PR to see if those two headers are indeed unused. They apparently are under a non-free license. They were later relicensed but if they are actually unused we don't need to bother with updating...

https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#Bad_Licenses

https://fedoraproject.org/wiki/Licensing/GLX_Public_License


paroj:
nice catch :+1:

PR #1868, opened 2021-01-29
Devel

### LINKED ISSUE #1864: OverlayElement setMaterial and setMaterialName affect the passed material
The signature of the setMaterial method of OverlayElement suggests that it does not affect the passed material as it's a reference to a const MaterialPtr, but it actually does, [see setMaterial in OgreOverlayElement.cpp](https://github.com/OGRECave/ogre/blob/2c16eda34d64f593eb88f412996bfb60bf7838d0/Components/Overlay/src/OgreOverlayElement.cpp#L333).

The method documentation states this behaviour, [see OgreOverlayElement.h](https://github.com/OGRECave/ogre/blob/master/Components/Overlay/include/OgreOverlayElement.h#L266).

But I think the current behaviour is very error prone by changing a given material that may shared with other entities.
Wouldn't it be better to throw an exception if the material doesn't meet the actual requirement.

paroj:
fixes #1864

### DIFF
--- a/Components/Overlay/src/OgreOverlayElement.cpp
+++ b/Components/Overlay/src/OgreOverlayElement.cpp
@@ -35,6 +35,8 @@ THE SOFTWARE.
 #include "OgreOverlayContainer.h"
 #include "OgreResourceGroupManager.h"
 #include "OgreOverlayElementCommands.h"
+#include "OgreTechnique.h"
+#include "OgreLogManager.h"
 
 namespace Ogre {
 
@@ -329,6 +331,15 @@ namespace Ogre {
             return;
 
         mMaterial->load();
+
+        auto dstPass = mMaterial->getTechnique(0)->getPass(0); // assume this is representative
+        if (dstPass->getLightingEnabled() || dstPass->getDepthCheckEnabled())
+        {
+            LogManager::getSingleton().logWarning(
+                "force-disabling 'lighting' and 'depth_check' of Material " + mat->getName() +
+                " for use with OverlayElement " + getName());
+        }
+
         // Set some prerequisites to be sure
         mMaterial->setLightingEnabled(false);
         mMaterial->setReceiveShadows(false);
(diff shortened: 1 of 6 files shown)


Components/Overlay/src/OgreOverlayElement.cpp
```diff
@@ -329,10 +331,19 @@ namespace Ogre {
             return;
 
         mMaterial->load();
+
+        auto dstPass = mMaterial->getBestTechnique()->getPass(0);
+        if (dstPass->getLightingEnabled() || dstPass->getDepthCheckEnabled())
```
akacoding, line 336: What is about "ReceiveShadows"?

Components/Overlay/src/OgreOverlayElement.cpp
```diff
@@ -329,10 +331,19 @@ namespace Ogre {
...
+                "force-disabling 'lighting' and 'depth_check' of Material " + mat->getName() +
+                " for use with OverlayElement " + getName());
+        }
+
         // Set some prerequisites to be sure
-        mMaterial->setLightingEnabled(false);
+        dstPass->setLightingEnabled(false);
         mMaterial->setReceiveShadows(false);
-        mMaterial->setDepthCheckEnabled(false);
+        dstPass->setDepthCheckEnabled(false);
```
akacoding, line 346: Isn't it changed logic now?
The calls to the material changes all techniques with all passes.
Now you're just changing one technique with one pass.
Or do I get something wrong?

Components/Overlay/src/OgreOverlayElement.cpp
paroj: I only added this recently. Not sure whether I will keep it this way.

Adding it to the check triggers with every overlay material of the sample browser. So it will likely spam all users.

Components/Overlay/src/OgreOverlayElement.cpp
paroj: the idea was to make this specific to the actually used pass. But the pass might change if the user switches the viewport scheme. So I should probably leave this alone.

PR #1877, opened 2021-02-05
SWIG: split core types out of Ogre.i

paroj:
to reduce resource pressure during compilation


paroj:
does not seem to help

PR #1892, opened 2021-02-17
Add DotScene usage scenarios

sercero:
I had a lot of trouble understanding how to use the Plugin, most of the information is in this thread: https://forums.ogre3d.org/viewtopic.php?t=96069 but it is not clear how to do it the modern way.
Eventually found the code in OgreSceneLoader.cpp, but new users should have an easier time using this important plugin.
If this gets accepted, I will see if the manual can be modified to put an example.

### DIFF
--- a/PlugIns/DotScene/README.md
+++ b/PlugIns/DotScene/README.md
@@ -1,4 +1,3 @@
-
 # DotScene Overview
 
 DotScene (aka .scene) is just a standardized XML file format.
@@ -54,7 +53,7 @@ In `plugins.cfg`, add the following line:
 Plugin=Plugin_DotScene
 ```
 
-Include the header `#include <OgreDotSceneLoader.h>`, located in: `OgreSDK\ogre-1.12.11\include\OGRE\Plugins\DotScene`
+Include the header `#include <Ogre.h>`, located in: `OgreSDK\ogre-1.12.11\include\OGRE`
 
 And to use the library , create a DataStream and pass it to `Ogre::Codec`:
 ```
@@ -74,5 +73,5 @@ attachmentNode->getUserObjectBindings().getUserAny("TerrainGroup");
 ```
 The type is std::shared_ptr<Ogre::TerrainGroup>, hence the attachmentNode owns it and will take it down on destruction.
 
-The Codec is now a central registry for encoding/decoding data. 
+The Codec is now a central registry for encoding/decoding data.
 Previously it was only used for Images, but now it also handles Meshes and Scenes.
--- a/PlugIns/DotScene/README.md
+++ b/PlugIns/DotScene/README.md
@@ -47,16 +47,16 @@ To add logic properties to the scene you can use the `<userData>` node as follow
 
 ## How to use DotScene
 In recent OGRE3D versions DotScene has been incorporated into the main OGRE repo and made into a Plugin.
-So it has to be loaded in `plugins.cfg`, add the following line:
+So it has to be loaded as another OGRE Plugin.
+
+In `plugins.cfg`, add the following line:
 ```
 Plugin=Plugin_DotScene
 ```
-Also, it is required to complie against the library (in a similar fashion to OgreBites) located in: `OgreSDK\ogre-1.12.11\lib\OGRE`
 
-Include the header `#include <OgreDotSceneLoader.h>
-`, located in: `OgreSDK\ogre-1.12.11\include\OGRE\Plugins\DotScene`
+Include the header `#include <OgreDotSceneLoader.h>`, located in: `OgreSDK\ogre-1.12.11\include\OGRE\Plugins\DotScene`
 
-And to use the library , create a DataStream and pass it to the loader:
+And to use the library , create a DataStream and pass it to `Ogre::Codec`:
 ```
 Ogre::String groupName = "Scene";
 Ogre::String filename = "myScene.scene";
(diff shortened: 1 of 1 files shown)


PlugIns/DotScene/README.md
```diff
@@ -42,4 +43,46 @@ To add logic properties to the scene you can use the `<userData>` node as follow
...
\ No newline at end of file
+```
+
+## How to use DotScene
+In recent OGRE3D versions DotScene has been incorporated into the main OGRE repo and made into a Plugin.
+So it has to be loaded in `plugins.cfg`, add the following line:
+```
+Plugin=Plugin_DotScene
+```
+Also, it is required to complie against the library (in a similar fashion to OgreBites) located in: `OgreSDK\ogre-1.12.11\lib\OGRE`
```
paroj, line 54: actually, you do not have to do that.

PlugIns/DotScene/README.md
```diff
@@ -42,4 +43,46 @@ To add logic properties to the scene you can use the `<userData>` node as follow
...
+
+And to use the library , create a DataStream and pass it to the loader:
+```
+Ogre::String groupName = "Scene";
+Ogre::String filename = "myScene.scene";
+Ogre::SceneNode attachmentNode = mSceneMgr->getRootSceneNode();
+
+Ogre::DataStreamPtr stream(Ogre::Root::openFileStream(filename, groupName));
+
+Ogre::DotSceneLoader *loader = new Ogre::DotSceneLoader();
```
paroj, line 67: instead do
```cpp
Codec::getCodec("scene")->decode(stream, attachmentNode);
```
---
while we are at it, you can add:

if a TerrainGroup is defined in the .scene file, you can get it by:
```cpp
attachmentNode->getUserObjectBindings().getUserAny("TerrainGroup");
```
the type is `std::shared_ptr<Ogre::TerrainGroup>`, hence the `attachmentNode` owns it and will take it down on destruction.

PlugIns/DotScene/README.md
```diff
@@ -42,4 +43,46 @@ To add logic properties to the scene you can use the `<userData>` node as follow
...
+loader->load(stream, groupName, attachmentNode );
+```
+
+## How to use DotScene (DEPRECATED)
```
paroj, line 71: please drop that section. Docs tend to get updated at a slower pace and soon it will refer to non-existing code.

paroj (changes requested):

paroj:
note, I deprecated `SceneLoaderManager` after improving the Codec API, which it is now duplicating.

typically that nasty `Codec::decode` call would be encapsulated somewhere (like is the case for Mesh). However, I am not sure where to put it.
`SceneManager::setWorldGeometry` seems suitable, but it is currently very much tailored to loading BSP "worlds" - whereas you can encapsulated arbitrary chunks of your scene in `.scene`.

PlugIns/DotScene/README.md
sercero: But if I don't do it then there are unresolved symbols at link time...

It was surprinsing because the deprecated method does not need it, unfortunately I am not doing things the way you specified in the manual and that might be the cause.

PlugIns/DotScene/README.md
sercero: That's a strange way to do it...

It is simpler, but does not seem very intuitive that a codec would do that.

PlugIns/DotScene/README.md
sercero: OK

PlugIns/DotScene/README.md
sercero: Are there other UserObjectBindings() registered for the attachmentNode, besides TerrainGroup?

sercero:
> note, I deprecated `SceneLoaderManager` after improving the Codec API, which it is now duplicating.
> 
> typically that nasty `Codec::decode` call would be encapsulated somewhere (like is the case for Mesh). However, I am not sure where to put it.
> `SceneManager::setWorldGeometry` seems suitable, but it is currently very much tailored to loading BSP "worlds" - whereas you can encapsulated arbitrary chunks of your scene in `.scene`.

I think the syntaxis should be something like this:
```
Ogre::DotSceneLoader *loader = new Ogre::DotSceneLoader();
loader->load("level.scene", "ResourceGroup", attachmentNode);
```

And behind the scenes there is the codec.

That seems to me like the most intuive way to do it...

PlugIns/DotScene/README.md
paroj: All "userData" content is put there for their parent nodes/ entities.

PlugIns/DotScene/README.md
paroj: > But if I don't do it then there are unresolved symbols at link time...

yes, you must use the Codec

sercero:
I just tested the code: `Codec::getCodec("scene")->decode(stream, attachmentNode);`

With this code I don't need to link against `libPlugin_DotScene` as you said, but now it is using the Resource Group "General", so I can't specify a resource group for the scene.

paroj:
it uses `ResourceGroupManager::getSingleton().getWorldResourceGroupName()` internally

paroj:
> I think the syntaxis should be something like this:

we cannot use the `DotSceneLoader` type or we would have to link against the plugin. 

Also, the user should not care about the format, so both `XXX.load("foo.scene")` and `XXX.load("foo.bsp")` could work if the according plugin is present.

PlugIns/DotScene/README.md
paroj: > It is simpler, but does not seem very intuitive that a codec would do that.

the Codec is now a central registry for encoding/ decoding data. Previously it was only used for Images, but now it also handles Meshes and Scenes.

PlugIns/DotScene/README.md
```diff
@@ -42,4 +43,36 @@ To add logic properties to the scene you can use the `<userData>` node as follow
...
+## How to use DotScene
+In recent OGRE3D versions DotScene has been incorporated into the main OGRE repo and made into a Plugin.
+So it has to be loaded as another OGRE Plugin.
+
+In `plugins.cfg`, add the following line:
+```
+Plugin=Plugin_DotScene
+```
+
+Include the header `#include <OgreDotSceneLoader.h>`, located in: `OgreSDK\ogre-1.12.11\include\OGRE\Plugins\DotScene`
```
paroj, line 57: the header is not needed either

sercero: I tried removing the header but then I complains about incomplete type:
```
C:\OGRE\Built\ogre-1.12.11\include\OGRE\OgreConfigFile.h|94|note: declared here|
C:\OGRE\Dev\ShaderTest\ShaderTest.cpp|188|error: incomplete type 'Ogre::Codec' used in nested name specifier|
```

paroj: `#include <OgreCodec.h>`

paroj: or even better: `#include <Ogre.h>`

sercero: Isn't it overkill to `#include <Ogre.h>`?

I thought it was better to include only what you need.

paroj: > I thought it was better to include only what you need.

generally, I would say no. There might be special cases, like headers that get included 1000x times (literally) which you want to keep small or compile-time increases.
But as a user of Ogre, you likely want for stuff to just work which Ogre.h gives you.

Also, it gives me the freedom to drop classes without breaking your code. E.g. OgreVector3.h is only kept for backward compatibility.

PlugIns/DotScene/README.md
```diff
@@ -42,4 +43,36 @@ To add logic properties to the scene you can use the `<userData>` node as follow
...
+Ogre::Codec::getCodec("scene")->decode(stream, mSceneMgr->getRootSceneNode());
+```
+
+If there is a TerrainGroup defined in the .scene file, you can get it by:
+```
+attachmentNode->getUserObjectBindings().getUserAny("TerrainGroup");
+```
+The type is std::shared_ptr<Ogre::TerrainGroup>, hence the attachmentNode owns it and will take it down on destruction.
+
+The Codec is now a central registry for encoding/decoding data.
```
paroj, line 77: trailing white-space here

PlugIns/DotScene/README.md
sercero: OK

paroj (approved):

PR #1902, opened 2021-02-22
Correct angle parsing

sercero:
The inner and outer angles of the SpotLight are being interpreted as Degrees, but the blender2ogre exporter is exporting the value in Radians.
For some reason in that context Angle is interpreting the input as Degrees.
I also changed parseQuaternion() for consistency although it seems to be working alright there.


paroj:
> For some reason in that context Angle is interpreting the input as Degrees.

https://github.com/OGRECave/ogre/blob/5d79056a87b222357f7ed3bbf57ed765aac7bf48/OgreMain/src/OgreMath.cpp#L54

> I also changed parseQuaternion() for consistency although it seems to be working alright there.

probably because some other overload is used.

PR #1906, opened 2021-02-24
Fix typo

sercero:
Fix typo that prevents convenient navigation


OgreMain/include/OgreInstanceManager.h
```diff
@@ -50,7 +50,7 @@ namespace Ogre
         This class takes care of managing batches automatically, so that more are created when
         needed, and reuse existing ones as much as possible; thus the user doesn't have to worry
         of managing all those low level issues.
-        @see InstanceBatch & @see InstanceEntity for more information.
```
paroj, line 53: "@see" usage is generally wrong here as it breaks formatting:
https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_instance_manager.html#details

Can you replace it by "see @ref" while at it?

PR #1911, opened 2021-03-02
build fix: remove unnecessary #ifdef __FreeBSD__

niacat:
bswapX macros are defined on other BSD systems too, so it's easier to just undefine them if they're already defined.

this fixes the build on NetBSD.


paroj:
:+1: thanks

Issue #1912, opened 2021-03-02
Terrain::updateCompositeMapWithDelay causes MaterialManager to set incorrect scheme


SNiLD:
#### System Information
- Ogre Version: 1.12.9
- Operating System / Platform: Windows 10 64-bit
- RenderSystem: OpenGL3+

#### Detailled description
Using `updateCompositeMapWithDelay` causes Terrain to start a counter that will be triggered in `preFindVisibleObjects`, which is called during `SceneManager::_renderScene`. This causes Terrain to render the composite map with different scene manager inside another scene manager's `_renderScene` (i.e. `_renderScene` will be called again, but for different scene manager). This would not be a problem except `SceneManager::_renderScene` calls `SceneManager::setViewport` which calls `MaterialManager::getSingleton().setActiveScheme(vp->getMaterialScheme());` which results in MaterialManager's active scheme to change during `_renderScene` (for the other scene manager in question).

I think real fix for this (in a more general sense) would be to make the `_renderScene` to use stack for setting active scheme in the same sense that `CompositorChain::_compile()` already does (i.e take the current state and put it back after we're done).

Here is a (stripped) stack trace of the situation in question:
```
[Inline Frame] OgreMain.dll!Ogre::SceneManager::setViewport(Ogre::Viewport *) Line 2800
	at d:\vcpkg\buildtrees\ogre\src\eddf310f0b-3998636895\ogremain\src\ogrescenemanager.cpp(2800)
OgreMain.dll!Ogre::SceneManager::_renderScene(Ogre::Camera * camera, Ogre::Viewport * vp, bool includeOverlays) Line 1339
	at d:\vcpkg\buildtrees\ogre\src\eddf310f0b-3998636895\ogremain\src\ogrescenemanager.cpp(1339)
OgreMain.dll!Ogre::Camera::_renderScene(Ogre::Viewport * vp, bool includeOverlays) Line 498
	at d:\vcpkg\buildtrees\ogre\src\eddf310f0b-3998636895\ogremain\src\ogrecamera.cpp(498)
OgreMain.dll!Ogre::RenderTarget::_updateViewport(Ogre::Viewport * viewport, bool updateStatistics) Line 204
	at d:\vcpkg\buildtrees\ogre\src\eddf310f0b-3998636895\ogremain\src\ogrerendertarget.cpp(204)
OgreMain.dll!Ogre::RenderTarget::_updateAutoUpdatedViewports(bool updateStatistics) Line 183
[...]

PR #1914, opened 2021-03-03
Storing active scheme name for current render in a stack

SNiLD:
Fixes https://github.com/OGRECave/ogre/issues/1912

### DIFF
--- a/OgreMain/include/OgreSceneManager.h
+++ b/OgreMain/include/OgreSceneManager.h
@@ -908,6 +908,10 @@ namespace Ogre {
         void fireSceneManagerDestroyed();
         /** Internal method for setting the destination viewport for the next render. */
         void setViewport(Viewport *vp);
+        /** Internal method for storing current material scheme and using the one provided for current render. */
+        void _pushActiveMaterialScheme(const String& scheme);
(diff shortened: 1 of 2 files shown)


OgreMain/include/OgreSceneManager.h
```diff
@@ -944,6 +948,8 @@ namespace Ogre {
 
         CompositorChain* mActiveCompositorChain;
         bool mLateMaterialResolving;
+        typedef std::deque<String> ActiveMaterialSchemeStack;
+        ActiveMaterialSchemeStack mActiveMaterialSchemeStack;
```
paroj, line 952: why do you introduce this instead of using the actual function-call stack?

SNiLD: Mainly because `Root::_pushCurrentSceneManager(SceneManager* sm)` uses same logic and I didn't know if it actually is possible end up there from somewhere else and mess it up. `SceneManager::_render()` loop is extremely complicated when combined with the possibility of another scenemanager being called from the main scene manager and them having their own render calls which all end up in `SceneManager::_render()` as well.

paroj:
Thanks for the research done here. I used the simpler fix in #1916

PR #1922, opened 2021-03-13
Improve RTShaderSystem diagnostic support by logging a description and file path when using a cached shader program

swebb2066:
This PR adds the cached shader program file names and description to the Ogre log.

This PR has only one effect when not using the cache:
- The empty function description line is not written into the shader source code. Writing a description would unnecessarily change the hash, preventing reuse of otherwse identical shaders.

### DIFF
--- a/Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
+++ b/Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
@@ -278,16 +278,13 @@ GpuProgramPtr ProgramManager::createGpuProgram(Program* shaderProgram,
 
     // Generate program name.
     String programName = generateHash(source, shaderProgram->getPreprocessorDefines());
-    const char* programType = " ";
     if (shaderProgram->getType() == GPT_VERTEX_PROGRAM)
     {
         programName += "_VS";
-        programType = " vertex shader ";
     }
     else if (shaderProgram->getType() == GPT_FRAGMENT_PROGRAM)
     {
         programName += "_FS";
-        programType = " fragment shader ";
     }
 
     // Try to get program by name.
@@ -298,8 +295,8 @@ GpuProgramPtr ProgramManager::createGpuProgram(Program* shaderProgram,
     if(pGpuProgram) {
         if (!cachePath.empty())
             LogManager::getSingleton().logMessage(shaderProgram->getEntryPointFunction()->getDescription()
-                + programType
-                + "cached as "
+                + ' ' + to_string(shaderProgram->getType())
+                + " shader cached as "
                 + programName
                 );
         return static_pointer_cast<GpuProgram>(pGpuProgram);
@@ -343,7 +340,7 @@ GpuProgramPtr ProgramManager::createGpuProgram(Program* shaderProgram,
         }
 
         LogManager::getSingleton().logMessage(shaderProgram->getEntryPointFunction()->getDescription()
-            + programType
+            + ' ' + to_string(shaderProgram->getType()) + " shader "
             + loadStore
             + programFileName
             );
--- a/Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
+++ b/Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
@@ -278,14 +278,16 @@ GpuProgramPtr ProgramManager::createGpuProgram(Program* shaderProgram,
 
     // Generate program name.
     String programName = generateHash(source, shaderProgram->getPreprocessorDefines());
-
+    const char* programType = " ";
     if (shaderProgram->getType() == GPT_VERTEX_PROGRAM)
     {
         programName += "_VS";
+        programType = " vertex shader ";
     }
     else if (shaderProgram->getType() == GPT_FRAGMENT_PROGRAM)
     {
         programName += "_FS";
+        programType = " fragment shader ";
     }
 
     // Try to get program by name.
@@ -294,6 +296,12 @@ GpuProgramPtr ProgramManager::createGpuProgram(Program* shaderProgram,
             programName, ResourceGroupManager::INTERNAL_RESOURCE_GROUP_NAME);
 
     if(pGpuProgram) {
+        if (!cachePath.empty())
+            LogManager::getSingleton().logMessage(shaderProgram->getEntryPointFunction()->getDescription()
+                + programType
+                + "cached as "
+                + programName
+                );
         return static_pointer_cast<GpuProgram>(pGpuProgram);
     }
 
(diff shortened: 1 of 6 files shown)


Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
```diff
@@ -278,14 +278,16 @@ GpuProgramPtr ProgramManager::createGpuProgram(Program* shaderProgram,
 
     // Generate program name.
     String programName = generateHash(source, shaderProgram->getPreprocessorDefines());
-
+    const char* programType = " ";
```
paroj, line 281: there is Ogre::to_string(GpuProgramType)

paroj:
can you provide some details on your usage here? Why do you need the association of pass and shader in the general case of no error in the shader?
Should all of this maybe be guarded with `#if OGRE_DEBUG`? Also, probably we should rather store the origin `Pass*` to construct the string on the fly and make things visible in the debugger.

swebb2066:
I was struggling to understand what was causing the differences in appearance of my application when I moved from OpenGL 1 to APIs without the fixed function pipeline using the RTShaderSystem.

As part of the process I used setShaderCachePath() to try to understand what  the RTShaderSystem was doing. However I could not work out which material each shader related to until I adding the logging. 

I also discovered your (relatively recent) change to ProgramManager::createGpuProgram where cached files are loaded if they already exist. That allowed me to experiment with shader code changes without having to rebuild my application. This is a very useful facility which is only useful when you know the shader-to-material relationship that the log provides.

As a general comment, I believe the Ogre users suffer from the lack of more finely targeted log activation. The log4j/log4cxx family allow logging level to be set for each of an arbitary number of named loggers. Good logging support is a critical feature of a large software system. Linux addressed this recently with the eBPF support.

Let me know if you want me to fix the syntax errors.

Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
```diff
@@ -294,6 +293,12 @@ GpuProgramPtr ProgramManager::createGpuProgram(Program* shaderProgram,
             programName, ResourceGroupManager::INTERNAL_RESOURCE_GROUP_NAME);
 
     if(pGpuProgram) {
+        if (!cachePath.empty())
+            LogManager::getSingleton().logMessage(shaderProgram->getEntryPointFunction()->getDescription()
+                + ' ' + to_string(shaderProgram->getType())
```
paroj, line 298: after a closer look, this is actually redundant with `programName` as the latter is suffixed wit _VS/ _FS

Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
```diff
@@ -294,6 +293,12 @@ GpuProgramPtr ProgramManager::createGpuProgram(Program* shaderProgram,
             programName, ResourceGroupManager::INTERNAL_RESOURCE_GROUP_NAME);
 
     if(pGpuProgram) {
+        if (!cachePath.empty())
```
paroj, line 296: I think abusing `cachePath` for extended logging is not right. What do you thing about `#if OGRE_DEBUG_MODE`

Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
```diff
@@ -322,14 +328,22 @@ GpuProgramPtr ProgramManager::createGpuProgram(Program* shaderProgram,
 
             outFile << source;
             outFile.close();
+            loadStore = "stored to ";
```
paroj, line 331: why do you want to log this? You can assume that if the file exists in the RTShaderCache, it will be read on the next run.

If you drop the load/store/cache info, you could just log at the start of this function.

paroj:
> The log4j/log4cxx family allow logging level to be set for each of an arbitary number of named loggers.

ogre also supports multiple log levels - altough setting one is a bit convoluted:

https://github.com/OGRECave/ogre/blob/56833c49119da275669cc1c168fa53c0e5ed1d65/OgreMain/include/OgreLog.h#L45

There is also support for multiple named loggers - the codebase just always uses the default one.

Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
swebb2066: I agree - abusing cachePath is not ideal.

A solution which is more consistent with other Ogre classes (and a better approach for a library) would be to add a Listner interface to RTShader::ProgramManager. Something like:
```
class CacheListener
{
virtual void notifyGpuProgram(Program* program, Pass* pass,  const String& programName, const String& cachePath) = 0;
};
```

Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
swebb2066: As a material may use an existing cache entry, this is needed to get all the material-to-shader mappings.

Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
swebb2066: As the cached files persist across application invocations, and the cached file content could have been edited (as I did when I was experimenting), it is important to know when the filesystem version is used and when the shader is freshly generated.

The time between application invocations could be large, so it is good to be reminded when Ogre is reading a shader from disk.

swebb2066:
Probably only the default log is used because Ogre::Log is missing some convenience macros like:
```
#if 4 < OGRE_LOG_THRESHOLD
/// Logs an \c importance type \c message to \c logger if its detail is level LL_LOW or higher.
#define OGRE_LOG_INFO(logger, importance, message) do { \
		if (LL_LOW <= logger->getLogDetail()) {\
			Log::Stream os; \
			logger->logMessage(os.str(os << message), importance); }} while (0)
#else
#define OGRE_LOG_INFO(logger, importance, message)
#endif
```

paroj:
that is exactly what logMessage does internally:
https://github.com/OGRECave/ogre/blob/ad241e4f8c19777bc3c46a7b3de9fb4fe4be25ee/OgreMain/src/OgreLog.cpp#L93-L96

Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
paroj: "cache" is a bit of a misnomer here as it does not save any computations. One would not enable this, unless you want to look at the generated code. So in hindsight using `cachePath` for extended logging should be fine.
This will also immediately tell you whether, "caching" is enabled.

paroj:
thanks for the valuable discussion, but ultimately I will rather go with a less invasive approach in #1925

PR #1929, opened 2021-03-18
CMake: fix static linkage of SDL2 and use static SDL2 by default

paroj:


Apteryks:
Why would you want to use static linkage to SDL2 by default?

paroj:
ease of shipping and a smaller binary.

Note, that this assumes you only use OgreBites and not any SDL2 API on top. However, this is guaranteed for our precompiled SDK or the Python package.

Apteryks:
> ease of shipping and a smaller binary.

Smaller binary? That's surprising, given static means it has to bundle all the needed symbols; smaller binaries are typically one of the advantages touted by dynamic linkage.

> Note, that this assumes you only use OgreBites and not any SDL2 API on top. However, this is guaranteed for our precompiled SDK or the Python package.

I'm not sure I understand the subtlety here, but as a package manager, my biased opinion would be that shared linkage by default makes more sense. Static linkage can always be requested by the Ogre SDK build scripts or in the specific cases where it is desired.

My 2 cents :-).

paroj:
ok, this issue is written from the snap/ windows/ flatpak/ PIP perspective, where we bundle all our dependencies and none of them can be shared with other processes. 

Here, static linkage means only the symbols we use are shipped, while a shared lib comes with all symbols and we need to copy it to the correct location.

Distributions on the other hand can use OGRE_BUILD_DEPENDENCIES=OFF and do whatever they want.

Apteryks:
I see, thanks for the explanations. I had overlooked that this change would only affect the building & linking of bundled dependencies. If this leaves a system-provided SDL2 to be linked dynamically, then I have nothing to say :-).

PR #1933, opened 2021-03-20
Overlay: add setVisible to OgreOverlay and OgreOverlayElement

akacoding:
Just added `setVisible` because with `show` and `hide` I have to evaluate what function to call in my code which results in a branch.

I do not like that now `show`, `hide` and `setVisible` are all virtual in `OgreOverlayElement`.
There are better alternatives to allow the inherited classes special behavour on visibility changes in this case.
But I do not want to introduce a breaking API change.

### DIFF
--- a/Components/Overlay/include/OgreOverlay.h
+++ b/Components/Overlay/include/OgreOverlay.h
@@ -136,6 +136,9 @@ namespace Ogre {
         /** Hides the overlay if it was visible. */
         void hide(void);
 
+        /** Shows or hides this overlay. */
+        void setVisible(bool visible);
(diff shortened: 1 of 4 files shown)


paroj:
> But I do not want to introduce a breaking API change.

actually this is a non-breaking API change as without unit-tests explicitly aiming at that you will not notice it.

In general, this is even worse. However, old Ogre code often tries to mimic Java - i.e. everything is virtual by default. So unless it is overridden somewhere in the code or "obviously" designed to be overridden, I accept removal of virtual even between API breaking releases.

PR #1934, opened 2021-03-21
When the mesh has no material assigned an assertion is triggered

sercero:
It is possible to generate a `.mesh` without an assigned material so the name will be blank.

In this case the following assert is triggered:
```
File: D:\OGRE2\Sources\ogre-master\OgreMain\src\OgreResourceManager.cpp, Line 50
Expression: (!name.empty()) && ("resource name must not be empty")
```

### DIFF
--- a/Tools/MeshUpgrader/src/main.cpp
+++ b/Tools/MeshUpgrader/src/main.cpp
@@ -969,8 +969,14 @@ struct MaterialCreator : public MeshSerializerListener
 {
     void processMaterialName(Mesh *mesh, String *name)
     {
-        // create material because we do not load any .material files
-        MaterialManager::getSingleton().createOrRetrieve(*name, mesh->getGroup());
+		if(name->empty()) {
+            OGRE_EXCEPT(Exception::ERR_INVALIDPARAMS,
+                "The provided mesh file has an empty material name. See https://ogrecave.github.io/ogre/api/latest/_mesh-_tools.html#Exporters");
+		}
+        else {
+            // create material because we do not load any .material files
+            MaterialManager::getSingleton().createOrRetrieve(*name, mesh->getGroup());
+        }
     }
 
     void processSkeletonName(Mesh *mesh, String *name) {}
@@ -1167,7 +1173,7 @@ int main(int numargs, char** args)
     }
     catch (Exception& e)
     {
-        cout << "Exception caught: " << e.getDescription();
+        cout << "Exception caught: " << e.getDescription() << std::endl;
         retCode = 1;
     }
 


paroj:
in the default configuration, `OgreAssert` will throw an exception that will be caught and print an error. With this change the empty name will be silently swallowed.

sercero:
It's not working for me.

I was trying to use the Ogremeshupgrader to generate some LODs and I was always getting an exception and no output mesh.

It just so happens that the mesh I was using does not have any materials. 
I know it is not the usual case, but it can be confusing.

It is true that an exception was being thrown when using the 'Release' version, but even with the exception being caught it does nothing, the LODs are not generated. The program catches the exception and finishes.

I found out about the assert when using the 'Debug' version to see where the exception was coming from.

paroj:
> It is true that an exception was being thrown when using the 'Release' version, but even with the exception being caught it does nothing, the LODs are not generated. The program catches the exception and finishes.

this is a feature actually :wink: You *will* arrive at that assertion sooner or later (unless you set the material programmatically), but then it will happen deep inside the render-loop which is way harder to debug.

Therefore, I would either do
1. Do nothing and bail out (as now). But add some message how to resolve this. e.g. convert to XML, set material manually, convert back.
2. As users expect the mesh to be modified, actually fix this: *warn* about it and set the material to "BaseWhite"

for reference, this is what the meshviewer does:
https://github.com/OGRECave/ogre-meshviewer/blob/ee95ee0858a48d720abcd929d72be9801231f9a3/ogre_mesh_viewer.py#L52-L55

sercero:
@paroj, I was thinking that this should actually be shown as an error.

What do you think about this message:
ERROR: The provided mesh file has an empty material name, refusing to process any further.
To set a material name for the mesh, you have these options:
 1. Re-export the mesh from your preferred DCC (Digital Content Creator) exporter, making sure that a material has been assigned.
 2. Edit the mesh.xml file to set a material name and reprocess the xml with OgreXMLConverter.
 3. Edit the mesh file with MeshMagick (https://github.com/OGRECave/meshmagick) to set a material name

paroj:
the wording is fine, however I would add the info to the manual and only print
> Error: resource name must not be empty. See https://ogrecave.github.io/ogre/api/latest/_mesh-_tools.html#Exporters

on the CLI.

sercero:
Updated the code to throw an exception if the mesh name is empty and direct the user to the manual (If I just checked the type of exception it was: RuntimeAssertionException, which is too generic).

Also modified the code to be a little less hard on people with OCD (inconsistent spacing).

Tools/MeshUpgrader/src/main.cpp
```diff
@@ -961,16 +950,20 @@ void resolveColourAmbiguities(Mesh* mesh)
...
-
 }
 
 struct MaterialCreator : public MeshSerializerListener
 {
     void processMaterialName(Mesh *mesh, String *name)
     {
-        // create material because we do not load any .material files
-        MaterialManager::getSingleton().createOrRetrieve(*name, mesh->getGroup());
+		if(*name == "") {
```
akacoding, line 959: Why not `name->empty()`?

sercero: No reason, if @paroj agrees I'll change it.

paroj: generally, `string::empty()` is faster as `string::compare()` involves a call to `strlen`

paroj:
as a sidenote: please put whitespace changes, bugfixes and new features into separate commits. See this for details:
https://kernelnewbies.org/PatchPhilosophy#How_to_break_up_changes

sercero:
Yes, it was sloppy of me to do that.

paroj:
:+1:

PR #1941, opened 2021-03-24
Draft terrain

### LINKED ISSUE #1662: Make SWIG support Terrain
currently there is no terrain support in swig for C#

natan-fernandes:
fixes #1662

### DIFF
--- a/Components/Terrain/include/OgreTerrain.i
+++ b/Components/Terrain/include/OgreTerrain.i
@@ -6,6 +6,7 @@
 %{
 /* Includes the header in the wrapper code */
 #include "Ogre.h"
+#include "OgreDefaultDebugDrawer.h"
 #include "OgreTerrainLayerBlendMap.h"
 #include "OgreTerrainMaterialGeneratorA.h"
 #include "OgreTerrain.h"
@@ -15,6 +16,7 @@
 %}
 
 %include std_string.i
+%include std_vector.i
 %include exception.i
 %import "Ogre.i"
 
@@ -33,7 +35,8 @@
 %template(TerrainRayResult) std::pair<bool, Ogre::Vector3>;
 %include "OgreTerrain.h"
 
-%ignore Ogre::TerrainGroup::getTerrainIterator;
+%ignore Ogre::TerrainGroup::rayIntersects;
+%ignore Ogre::TerrainGroup::getTerrainIterator; // deprecated
 %include "OgreTerrainGroup.h"
 
 %include "OgreTerrainLodManager.h"
--- a/Components/Terrain/include/OgreTerrain.i
+++ b/Components/Terrain/include/OgreTerrain.i
@@ -1,0 +1,40 @@
+#ifdef SWIGPYTHON
+%module(package="Ogre") Terrain
+#else
+%module OgreTerrain
+#endif
+%{
+/* Includes the header in the wrapper code */
+#include "Ogre.h"
+#include "OgreTerrainLayerBlendMap.h"
+#include "OgreTerrainMaterialGeneratorA.h"
+#include "OgreTerrain.h"
+#include "OgreTerrainQuadTreeNode.h"
+#include "OgreTerrainAutoUpdateLod.h"
+#include "OgreTerrainGroup.h"
+%}
+
+%include std_string.i
+%include exception.i
+%import "Ogre.i"
+
+#define _OgreTerrainExport
+#define __inline
+#define __forceinline
+
+%ignore Ogre::TerrainMaterialGenerator::setActiveProfile;
+%include "OgreTerrainMaterialGenerator.h"
+%include "OgreTerrainMaterialGeneratorA.h"
+
+%include "OgreTerrainLayerBlendMap.h"
+%include "OgreTerrainQuadTreeNode.h"
+
+%template(LayerInstanceList) std::vector<Ogre::Terrain::LayerInstance>;
+%template(TerrainRayResult) std::pair<bool, Ogre::Vector3>;
+%include "OgreTerrain.h"
+
+%ignore Ogre::TerrainGroup::getTerrainIterator;
+%include "OgreTerrainGroup.h"
+
+%include "OgreTerrainLodManager.h"
+%include "OgreTerrainPrerequisites.h"
--- a/Components/Csharp/CMakeLists.txt
+++ b/Components/Csharp/CMakeLists.txt
@@ -32,6 +32,11 @@ if(OGRE_BUILD_COMPONENT_RTSHADERSYSTEM)
 	list(APPEND SWIG_INPUT_MODULES ../RTShaderSystem/include/OgreRTShader.i)
 endif()
 
+if(OGRE_BUILD_COMPONENT_TERRAIN)
+	set_source_files_properties(../Terrain/include/OgreTerrain.i PROPERTIES CPLUSPLUS ON)
+	list(APPEND SWIG_INPUT_MODULES ../Terrain/include/OgreTerrain.i)
+endif()
+
 if(OGRE_BUILD_COMPONENT_OVERLAY)
 	set_source_files_properties(../Overlay/include/OgreOverlay.i PROPERTIES CPLUSPLUS ON)
 	list(APPEND SWIG_INPUT_MODULES ../Overlay/include/OgreOverlay.i)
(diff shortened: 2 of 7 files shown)


Components/Terrain/include/OgreTerrain.i
```diff
@@ -0,0 +1,47 @@
...
+// #include "OgreUnifiedHighLevelGpuProgram.h"
+%}
+
+%include std_string.i
+%include exception.i 
+%import "Ogre.i"
+
+#define _OgreTerrainExport
+
+%include "OgreTerrainLayerBlendMap.h"
```
paroj, line 21: to be sure, you should also `#include` these in the section above.

Components/Terrain/include/OgreTerrain.i
```diff
@@ -0,0 +1,47 @@
+%module(package="Ogre") Terrain
+%{
+/* Includes the header in the wrapper code */
+#include "Ogre.h"
+#include "OgreTerrain.h"
+#include "OgreTerrainAutoUpdateLod.h"
+#include "OgreTerrainGroup.h"
+#include "OgreTerrainPagedWorldSection.h"
```
paroj, line 8: for now, you should stay away for the `*Paged*` stuff, as it depends on a separate component and thus complicates stuff.
Paging is not needed for "normal" Terrain usage.

Components/Terrain/include/OgreTerrain.i
paroj, line 21: you must `%include "OgreTerrain.h"` first, so TerrainLayerBlendMap knows about it. This is what the error:
```
/home/runner/work/ogre/ogre/build/csharp/OgreTerrainCSHARP_wrap.cxx:450:3: error: ‘Terrain’ was not declared in this scope
   Terrain *arg1 = (Terrain *) 0 ;
   ^~~~~~~
/home/runner/work/ogre/ogre/build/csharp/OgreTerrainCSHARP_wrap.cxx:450:3: note: suggested alternative:
In file included from /home/runner/work/ogre/ogre/build/csharp/OgreTerrainCSHARP_wrap.cxx:383:0:
/home/runner/work/ogre/ogre/Components/Terrain/include/OgreTerrain.h:261:30: note:   ‘Ogre::Terrain’
```
means.

natan-fernandes: Should I do a `%include` for every `#include`? I thought it wasn't necessary

paroj: generally, it is not necessary. However %include is not recursive, so you might end up with missing declarations.

natan-fernandes:
Clicked on ready for review by mistake.
I did all the changes you said and I get this error on `OgreTerrain.h`:
`Syntax error - possibly a missing semicolon` at line: 1854

But I don't see any missing semicolon
https://github.com/OGRECave/ogre/blob/master/Components/Terrain/include/OgreTerrain.h#L1854

paroj:
add `#define OGRE_FORCE_INLINE`

natan-fernandes:
Now I get 2 erros: 
```
Macro 'OGRE_FORCE_INLINE' redefined @ OgreTerrain.i
previous definition of 'OGRE_FORCE_INLINE'. @ OgrePlataform.h ln 133
```

natan-fernandes:
Where can I remove the previous `#define` or tell it to aways force?

Components/Terrain/include/OgreTerrain.i
```diff
@@ -0,0 +1,47 @@
...
+//#include "OgrePageManager.h"
+//#include "OgrePagedWorld.h"
+//#include "OgreTerrainPaging.h"
+%}
+
+%include std_string.i
+%include exception.i 
+%import "Ogre.i"
+
+#define _OgreTerrainExport
```
paroj, line 19:
```suggestion
#define _OgreTerrainExport
#define OGRE_FORCE_INLINE
```
should be here

natan-fernandes: Had the same issue 🙁: 
```
Macro 'OGRE_FORCE_INLINE' redefined.
```

The previous definition is at `OgrePlataform.h`. Somehow the file I have is different from the one on this repository. But here's how it's in mine
```
/* See if we can use __forceinline or if we need to use __inline instead */
#if OGRE_COMPILER_MIN_VERSION(OGRE_COMPILER_MSVC, 1200)
    #define OGRE_FORCE_INLINE __forceinline
#elif OGRE_COMPILER_MIN_VERSION(OGRE_COMPILER_GNUC, 340)
    #define OGRE_FORCE_INLINE inline __attribute__((always_inline))
#else
        #define OGRE_FORCE_INLINE __inline   <----------------------- This is being defined
#endif
```

Components/Terrain/include/OgreTerrain.i
```diff
@@ -0,0 +1,23 @@
...
+#include "OgreTerrainGroup.h"
+#include "OgreTerrainLayerBlendMap.h"
+%}
+
+%include std_string.i
+%include exception.i 
+%import "Ogre.i"
+
+#define _OgreTerrainExport
+#define OGRE_FORCE_INLINE
```
paroj, line 16: maybe this?
```suggestion
#define __inline
#define __forceinline
```

Components/Terrain/include/OgreTerrain.i
```diff
@@ -0,0 +1,24 @@
+%module(package="Ogre") Terrain
```
paroj, line 1:
```suggestion
#ifdef SWIGPYTHON
%module(package="Ogre") Terrain
#else
%module OgreTerrain
#endif
```

natan-fernandes:
Working! 
Now I get 
```
'TerrainLayerDeclaration': undeclared identifier
```
and already tried `#include` and `%include` "OgreTerrainMaterialGenerator.h" but had no success. Is there any header am I missing?

paroj:
you need to add that before `%include "OgreTerrain.h"`. If that gives you other errors you can add `%ignore Ogre::TerrainLayerDeclaration`.

natan-fernandes:
Already tried 
```
%nodefaultctor;
    %include "OgreTerrainGroup.h"
%clearnodefaultctor;
```
but didn't work

paroj:
please update this PR, so I can take a look at the full error log. Probably you should %ignore the fuction associated with MapIterator

natan-fernandes:
If I ignore it will I be able to use it in my C# code?

paroj:
```
%ignore TerrainGroup::getTerrainIterator
```
it will not show up in C#, but you would not be able to use it anyway as the IteratorWrapper does not work with SWIG.

Instead, something like this needs to be done:
https://github.com/OGRECave/ogre/blob/fc789998fa455ea8a4e930c79d40cb930f1f1319/OgreMain/include/OgreSceneNode.h#L233-L245

natan-fernandes:
I need this to do this like Mogre did in the past:
```
foreach (TerrainGroup.TerrainSlot ts in mTerrainGroup.GetTerrainIterator())
{
    ...
}
```

So I need to search for all objects that are `TerrainSlot`? using `getAttachedObjects()`

natan-fernandes:
I will try to do this. 
Also is there a way to create a pointer to a Vector3 like this?
```
Vector3 vAux = new Vector3(1);
Vector3* vPosTerrain = &vAux;
```

It throws `cannot take the address of get the size of or declare a pointer to a managed type(Vector3)`. Is it related to SWIG?

paroj:
assuming the bindings are correct, you never should have to use `Vector3*`

Components/Terrain/include/OgreTerrain.i
```diff
@@ -24,8 +24,12 @@
 %include "OgreTerrainMaterialGenerator.h"
 %include "OgreTerrainLayerBlendMap.h"
 %include "OgreTerrainQuadTreeNode.h"
+
+%template(LayerInstanceList) std::vector<Ogre::Terrain::LayerInstance>;
+%template(Pair) std::pair<bool, Ogre::Vector3>;
```
natan-fernandes, line 33: Need to find a better name to Pair, as there are another Pairs in the code that aren't <bool, Vector3>

paroj:
@natan-fernandes are you still working on this?

natan-fernandes:
I got a job so atm I don't have time to work on it but I've made some changes on some cpp files to add `getTerrainSlots()`, I will do a PR tomorrow

PR #1942, opened 2021-03-25
Fix PrecompiledHeader.cmake for old CMake versions

oold:
Using GREATER_EQUAL breaks build file generation for CMake versions less than 3.7. Use "NOT ... VERSION_LESS" instead.


paroj:
thanks

PR #1956, opened 2021-04-02
Devel

paroj:


OgreMain/include/Ogre.i
```diff
@@ -699,6 +705,7 @@ SHARED_PTR(Material);
     %ignore Ogre::Light::getDirection;
     %include "OgreLight.h"
     %ignore Ogre::Node::getChildIterator;
+    %template(ChildNodeMap) std::vector<Ogre::Node*>;
```
akacoding, line 708: In an earlier PR I saw the name `xyzMap` for the `std::map` type.
Here you have a `std::vector`. I think you should use a different name.

paroj: note that we have this inconsistency in C++ as well to increase backwards-compatibility with pre 1.10 code.
 
But for new SWIG based code, we can use correct names. Will update NodeObjectMap below for 1.13.

Issue #58, opened 2015-11-29
[OGRE-265] build dir layout issues


paroj:
<b>[reporter="corristo", created="Sun, 15 Sep 2013 16:04:02 +0200"]</b>

<p>I'm building Ogre using following cmake call:</p>

<div class="preformatted panel" style="border-width: 1px;"><div class="preformattedContent panelContent">
<pre>#!/bin/sh
cmake \
-D CMAKE_C_COMPILER=/usr/bin/clang \
-D CMAKE_CXX_COMPILER=/usr/bin/clang++ \
-D CMAKE_OSX_ARCHITECTURES="x86_64" \
-D OGRE_BUILD_SAMPLES=FALSE \
-D OGRE_BUILD_TOOLS=FALSE \
-D CMAKE_BUILD_TYPE=Release \
-D CMAKE_INSTALL_PREFIX=/usr/local/Cellar/Ogre/1.9.0-unstable \
..
</pre>
</div></div>

<p>Build dir is located inside source dir.</p>

<p>As you can see from screenshot I got two distinct Ogre.framework copies. But one has headers that other doesn't. It gets even worse after doing "make install", only one instance of Ogre.framework got copied to install prefix.</p>

<p>Also, why there's "macosx" intermediate folder in $build/lib? It can confuse some cmake scripts that assume that libs must be right under lib. It's especially important when installing Ogre in /usr/local (i.e. system-wide) for example. /usr/local/lib/macosx/Release just isn't right path for libraries =\</p>

<p>I'm using revision 4ddd8edfc4f25b72b674f520bbcc1358f54b8c6d</p>

paroj:
<b><i>[author="corristo", created="Sun, 15 Sep 2013 17:15:21 +0200"]</i></b>
<p>Official Ogre 1.9RC1 SDK build has the same issue: Ogre.framework hasn't "Threading" and "OSX" header subfolders.</p>

paroj:
<b><i>[author="masterfalcon", created="Sun, 15 Sep 2013 19:31:48 +0200"]</i></b>
<p>The reason for macosx folder is in case people are including iOS libraries as well and trying to install them into /usr/local.</p>

<p>The duplicate may be a side effect of building with makefiles. Have you tried generating a Xcode project and seeing if the results are the same?</p>

paroj:
<b><i>[author="corristo", created="Fri, 27 Sep 2013 10:23:43 +0200"]</i></b>
[...]

PR #1974, opened 2021-04-16
DotSceneLoader - Node animation support

sercero:
Add support for node animation, the idea originates from [NodeAnimation exporter #58](https://github.com/OGRECave/blender2ogre/issues/58)
 - The code is based on the EasyOgreExporter that has already implemented it
 - Modified the DTD to reflect the node animations
 - Added functions to DotSceneLoader to process the animation data
 - Added logging to most functions in DotSceneLoader

### DIFF
--- a/PlugIns/DotScene/misc/dotscene.dtd
+++ b/PlugIns/DotScene/misc/dotscene.dtd
@@ -22,7 +22,7 @@
 
 <!ELEMENT nodes (node*, position?, rotation?, scale?)>
 
-<!ELEMENT node (position?, rotation?, scale?, lookTarget?, trackTarget?, userData?, node*, entity*, light*, camera*, particleSystem*, billboardSet*, plane*)>
+<!ELEMENT node (position?, rotation?, scale?, lookTarget?, trackTarget?, userData?, animations?, node*, entity*, light*, camera*, particleSystem*, billboardSet*, plane*)>
 <!ATTLIST node
     name        CDATA    #IMPLIED
     id            ID        #IMPLIED
@@ -43,7 +43,7 @@
     type            (point | directional | spot ) "point"
     visible            (true | false) "true"
     castShadows        (true | false) "true"
-    powerScale		CDATA	"1.0"
+    powerScale      CDATA   "1.0"
 >
 
 <!ELEMENT camera (clipping?, userData?)>
@@ -273,4 +273,23 @@
     name CDATA #REQUIRED
     type ( bool | float | int | str ) "str"
     data CDATA #REQUIRED
+>
+
+<!-- https://www.w3schools.com/xml/xml_dtd_elements.asp -->
+
+<!ELEMENT animations (animation+)>
+
+<!ELEMENT animation (keyframe+)>
+<!ATTLIST animation
+    name    PCDATA          #REQUIRED
+    length  PCDATA          #REQUIRED
+    enable  (true | false)  "false"
+    loop    (true | false)  "false"
+    interpolationMode           (linear | spline)       "linear"
+    rotationInterpolationMode   (linear | spherical)    "linear"
+>
+
+<!ELEMENT keyframe (translation?, rotation?, scale?)>
+<!ATTLIST keyframe
+    time    PCDATA  #REQUIRED
 >
--- a/PlugIns/DotScene/src/DotSceneLoader.cpp
+++ b/PlugIns/DotScene/src/DotSceneLoader.cpp
@@ -206,6 +206,8 @@ void DotSceneLoader::processScene(pugi::xml_node& XMLRoot)
 
 void DotSceneLoader::processNodes(pugi::xml_node& XMLNode)
 {
+    LogManager::getSingleton().logMessage("[DotSceneLoader] Processing Nodes...", LML_TRIVIAL);
+
     // Process node (*)
     for (auto pElement : XMLNode.children("node"))
     {
@@ -241,6 +243,8 @@ void DotSceneLoader::processExternals(pugi::xml_node& XMLNode)
 
 void DotSceneLoader::processEnvironment(pugi::xml_node& XMLNode)
 {
+    LogManager::getSingleton().logMessage("[DotSceneLoader] Processing Environment...", LML_TRIVIAL);
+
     // Process camera (?)
     if (auto pElement = XMLNode.child("camera"))
         processCamera(pElement);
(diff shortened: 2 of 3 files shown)


PlugIns/DotScene/src/DotSceneLoader.cpp
```diff
@@ -789,6 +827,97 @@ void DotSceneLoader::processUserData(pugi::xml_node& XMLNode, UserObjectBindings
...
+}
+
+void DotSceneLoader::processKeyframe(pugi::xml_node& XMLNode, NodeAnimationTrack* pTrack)
+{
+    // Process node animation keyframe (*)
+	Real time = getAttribReal(XMLNode, "time");
+
+    LogManager::getSingleton().logMessage("[DotSceneLoader] Processing Keyframe: " + StringConverter::toString(time));
+
+	auto keyframe = pTrack->createNodeKeyFrame(time);
```
paroj, line 897: please convert all tabs to spaces (4 space indentation)

PlugIns/DotScene/src/DotSceneLoader.cpp
```diff
@@ -152,7 +152,7 @@ void DotSceneLoader::load(DataStreamPtr& stream, const String& groupName, SceneN
         LogManager::getSingleton().logError("DotSceneLoader - Invalid .scene File. Missing <scene formatVersion='x.y' >");
         return;
     }
-
+
```
paroj, line 155: trailing whitespace. Run
```sh
git diff --check master
```
to run the check locally.

PlugIns/DotScene/misc/dotscene.dtd
```diff
@@ -274,3 +274,44 @@
...
+>
+
+<!ELEMENT translation EMPTY>
+<!ATTLIST translation
+	x	CDATA	#REQUIRED
+	y	CDATA	#REQUIRED
+	z	CDATA	#REQUIRED
+>
+
+<!ELEMENT rotation EMPTY>
```
paroj, line 304: this is already defined above

PlugIns/DotScene/misc/dotscene.dtd
```diff
@@ -274,3 +274,44 @@
...
+<!ATTLIST rotation
+	qx	CDATA	#REQUIRED
+	qy	CDATA	#REQUIRED
+	qz	CDATA	#REQUIRED
+	qw	CDATA	#REQUIRED
+>
+
+<!ELEMENT scale EMPTY>
```
paroj, line 312: this is already defined as well

PlugIns/DotScene/misc/dotscene.dtd
```diff
@@ -274,3 +274,44 @@
...
+	rotationInterpolationMode	(linear | spline)	"linear"
+>
+	
+<!ELEMENT keyframe (translation?, rotation?, scale?)>
+<!ATTLIST keyframe
+	time	PCDATA	#REQUIRED
+>
+
+<!ELEMENT translation EMPTY>
+<!ATTLIST translation
```
paroj, line 298: we already have position, which is semantically the same. Therefore I would rather use that - even at the cost of incompatibility with EOE. I would rather fixup EOE.

paroj (changes requested):

PlugIns/DotScene/src/DotSceneLoader.cpp
```diff
@@ -311,6 +317,8 @@ void DotSceneLoader::processLight(pugi::xml_node& XMLNode, SceneNode* pParent)
     // Process attributes
     String name = getAttrib(XMLNode, "name");
 
+    LogManager::getSingleton().logMessage("[DotSceneLoader] Processing Light: " + name);
```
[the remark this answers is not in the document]
sercero: No, you are right.

I didn't know how to mark it as such, thanks.

PlugIns/DotScene/src/DotSceneLoader.cpp
```diff
@@ -789,6 +827,97 @@ void DotSceneLoader::processUserData(pugi::xml_node& XMLNode, UserObjectBindings
...
+    LogManager::getSingleton().logMessage("[DotSceneLoader] Processing Node Animation: " + name);
+	
+	Real length = getAttribReal(XMLNode, "length");
+	
+	Animation* anim = mSceneMgr->createAnimation(name, length);
+
+	bool enable = getAttribBool(XMLNode, "enable", false);
+	bool loop = getAttribBool(XMLNode, "loop", false);
+
+    String interpolationMode = getAttrib(XMLNode, "interpolationMode");
```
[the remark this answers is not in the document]
sercero: Actually this was a mistake, one was interpolationMode and the other was rotationInterpolationMode.

PlugIns/DotScene/misc/dotscene.dtd
sercero: @paroj the problem is that this is actually a translation and not a position, see the manual:
[Ogre::TransformKeyFrame::setTranslate()](https://ogrecave.github.io/ogre/api/1.12/class_ogre_1_1_transform_key_frame.html#a8d82e03beea51388e7daf005b76b00a1)

sercero: Nevermind, I see you already changed EOE besides it makes it easier to merge with `_ogre_node_helper` which uses position.

paroj: I would prefer to keep the DTD simple, instead of following the API naming. It is inconsistent anyway:
position and translation are the same thing, while all values specified here are *offsets* to the previous keyframe..

PlugIns/DotScene/misc/dotscene.dtd
```diff
@@ -274,3 +274,44 @@
...
+<!ELEMENT animations (animation+)>
+
+<!ELEMENT animation (keyframe+)>
+<!ATTLIST animation
+	name	PCDATA			#REQUIRED
+	length	PCDATA			#REQUIRED
+	enable	(true | false)	"false"
+	loop	(true | false)	"false"
+	interpolationMode			(linear | spline)	"linear"
+	rotationInterpolationMode	(linear | spline)	"linear"
```
paroj, line 289:
```suggestion
	rotationInterpolationMode	(linear | spherical)	"linear"
```

paroj (approved):

PR #2000, opened 2021-05-02
Main InstanceManager - unshare LODs as well

erikogenvik:
This fixes a segfault which will appear if a mesh with shared vertices and existing edge data is used.
It's triggered by the calls to "buildEdgeList()", since when doing this the LODs will be accessed.
And since they haven't been altered, they will rely on the now deleted shared geometry, and
their indices will be out of bounds, causing segfault.


paroj:
thanks

PR #2004, opened 2021-05-04
D3D*: add missing RSC_AUTOMIPMAP_COMPRESSED

paroj:


paroj:
not missing: they are actually not supported

PR #2011, opened 2021-05-07
Main: DefaultDebugDrawer -Skip bounding boxes that are infinite.

erikogenvik:
Since this will cause an assert error when built in debug mode,
and undefined behaviour if not.


paroj:
thanks

PR #2032, opened 2021-05-19
Disable vptr checks for Singleton constructor

oold:
UBSan complains a lot about the this pointer not being of type T* which is correct during construction, but ultimately doesn't matter. This change disables the vptr check on the Singleton constructor to avoid errors from UBSan.

### DIFF
--- a/OgreMain/include/OgreSingleton.h
+++ b/OgreMain/include/OgreSingleton.h
@@ -80,7 +80,7 @@ protected:
     static T* msSingleton;
 
 public:
-#if defined(__GNUC__) || defined(__clang__)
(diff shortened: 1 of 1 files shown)


paroj:
oh the joy of cross-platform code :wink:

The docs say you cant check for __has_attribute like that:
https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html

which means that you can use the recommended version, and the __GNUC__ check is redundant.

oold:
Oh, wow. I even tested this with different versions of GCC and Clang.
Also very interesting: `__has_attribute` doesn't seem to be documented until GCC 9, but even GCC 5 seems to know it.
I was actually quite sure that the way I did it wouldn't work, so I was surprised when the old GCCs accepted the code.

PR #2060, opened 2021-06-03
Main: fixup incomplete getConstantDefinitions const change

paroj:


Joilnen:
cool that was raising android compilation warnings, thanks

PR #2065, opened 2021-06-05
fixing close curly bracket error

Joilnen:
getting compilation error in linux


paroj:
my guess would be that you have a newer glslang and thus have to comment in this:
https://github.com/OGRECave/ogre/blob/7fc8890f8f115a33f7b7f8d8ba546ba257435176/PlugIns/GLSLang/src/OgreGLSLang.cpp#L110

Joilnen:
where this is declared TBuiltInResource ?

Joilnen:
Redoing this better

Joilnen:
Yes for glslang 11 it doesn't compile without that last parameter before limits, but it works for glslang 8.x

PR #2125, opened 2021-07-23
fix compile warning when passing a 32 bit type into an 8 bit type.

creadmefford:
this warning prevents my app from compiling cleanly.  did not occur on 1.12.9.


paroj:
#2126 should fix the root cause of this

creadmefford:
excellent, thanks!

PR #2164, opened 2021-08-17
Fix compiling errors due to "Out of line definitions".

n3rd0x:
Looks like there are mismatch between Real and float, in header and source files. These changes are based on the definition in the header files.


paroj:
:+1:  thanks

PR #2194, opened 2021-09-20
Enable OBJC and OBJCXX for apple build

catap:
Without this small changes build on macOS fails with error like:
```
error: Objective-C was disabled in PCH file but is currently enabled
```

Here a two small changes:
 - increase minimal version of cmake to 3.16.0.
 - enable languages `OBJC` and `OBJCXX`;

CMake-3.16.0 was the first one which is supported `OBJC` and `OBJCXX`: https://discourse.cmake.org/t/cmake-3-16-0-available-for-download/262

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,11 @@ if (APPLE AND NOT ANDROID)
   set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++11")
   set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++")
 
+  if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16.0")
+    enable_language(OBJC)
+    enable_language(OBJCXX)
+  endif()
+
   if(POLICY CMP0068)
     cmake_policy(SET CMP0068 NEW)  # CMake 3.9+: `RPATH` settings on macOS do not affect `install_name`.
   endif()


CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 # and provide build configuration options.
 ######################################################################
 
-cmake_minimum_required(VERSION 3.10.0)
+cmake_minimum_required(VERSION 3.16.0)
```
paroj, line 17: I would rather not bump the required cmake version mid-release

CMakeLists.txt
```diff
@@ -46,6 +46,9 @@ if (APPLE AND NOT ANDROID)
   set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD "c++11")
   set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++")
 
+  enable_language(OBJC)
```
paroj, line 49:
```suggestion
if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16.0")
  enable_language(OBJC)
```

paroj:
also, this seems to be specific to a certain xcode/ cmake combination as our runner works fine:
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

catap:
@paroj I'm trying macOS Big Sur (aka 11) with latest xcode and without `enable_language(OBJC)` build simple fails.

I'll try your suggesting soon.

catap:
Also, I'm updating MacPorts port of `ogre` and I can encounter this issue when I build it via macports.

It is possible that MacPorts somehow broke detection of enabled languages.

PR #2199, opened 2021-09-21
Allow to control resources path

catap:
This commit introduced two CMake options `OGRE_MEDIA_PATH` and `OGRE_PLUGINS_PATH` to control such paths.

Also, it keeps `OGRE_CFG_INSTALL_PATH`, `OGRE_CMAKE_DIR` and `OGRE_PLUGIN_DIR_CMAKE` untouched when it exist.


CMake/InstallResources.cmake
```diff
@@ -11,33 +11,51 @@
...
 
+if (NOT OGRE_MEDIA_PATH)
+  if (WIN32 OR APPLE)
+    set(OGRE_MEDIA_PATH "Media")
+  elseif (UNIX)
+    set(OGRE_MEDIA_PATH "share/OGRE/Media")
+  endif()
+endif()
+
+set(OGRE_MEDIA_DIR_REL "${CMAKE_INSTALL_PREFIX}/${OGRE_MEDIA_PATH}")
```
paroj, line 22: this changes the behavior compared to the current state, which will break things.

The difference between `Media`, `./Media` and `../Media` is important

paroj (changes requested):

catap:
@paroj I hope that I recovers all of them

PR #2211, opened 2021-10-03
fix some warnings messages in android ndk r23 compilation

Joilnen:
[warnings.log](https://github.com/OGRECave/ogre/files/7272605/warnings.log)


paroj:
these fixes are wrong. The results will be 0 for all of them.

Joilnen:
I see the problem

PR #2218, opened 2021-10-05
Use Qt 6 for OgreBitesQt if present

blakharaz:
Use the cmake commands of
https://doc-snapshots.qt.io/qt6-dev/cmake-qt5-and-qt6-compatibility.html#supporting-older-qt-5-versions
to support both Qt 5 and Qt 6.


paroj:
thanks!

PR #2231, opened 2021-10-15
Use safe idiom for math constants.

mahiuchun:
Constexpr is used to ensure that the constants are initialized in the "constant initialization" phase.

Reference:
- https://abseil.io/tips/140
- https://en.cppreference.com/w/cpp/language/initialization

### DIFF
--- a/OgreMain/include/OgreMath.h
+++ b/OgreMain/include/OgreMath.h
@@ -358,7 +358,7 @@ namespace Ogre
         static inline Real Log (Real fValue) { return std::log(fValue); }
 
         /// Stored value of log(2) for frequent use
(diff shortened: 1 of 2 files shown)


paroj:
this switches these constants to internal linkage, which we probably do not want.

Does this fixes an actual issue you encountered?

mahiuchun:
What do you mean by internal linkage? The constants are accessible outside OgreMath.cpp.

The issue I had was that AddressSanitizer's initialization-order-fiasco check fails in Ogre code. This isn't a complete fix as things like `Ogre::Vector3::ZERO` also causes the said check to fail for the same reason.

PR #2301, opened 2021-12-05
Fedora setup

bkmgit:
Add instructions for Fedora dependencies

### DIFF
--- a/BuildingOgre.md
+++ b/BuildingOgre.md
@@ -77,7 +77,8 @@ these will enable input handling in the SampleBrowser, the X11 ConfigDialog and 
 * Doxygen: http://doxygen.org
 * Cg: http://developer.nvidia.com/object/cg_toolkit.html
 * Remotery: https://github.com/Celtoys/Remotery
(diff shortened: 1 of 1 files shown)


BuildingOgre.md
```diff
@@ -66,7 +77,8 @@ these will enable input handling in the SampleBrowser, the X11 ConfigDialog and
 * Doxygen: http://doxygen.org
 * Cg: http://developer.nvidia.com/object/cg_toolkit.html
 * Remotery: https://github.com/Celtoys/Remotery
-* %Assimp: https://www.assimp.org/
```
paroj, line 69: please keep the "%" - it is needed as the page is also used by Doxygen:
https://ogrecave.github.io/ogre/api/latest/building-ogre.html#autotoc_md259

paroj (changes requested):

BuildingOgre.md
bkmgit: Thanks for the feedback. Changed.

PR #2317, opened 2021-12-14
Python: Try to detect correct python sitelib path before guessing one

susnux:
Try to detect the sitelib path from python instead of guessing it.
Used by openSUSE to fix packaging issues.

### DIFF
--- a/Components/Python/CMakeLists.txt
+++ b/Components/Python/CMakeLists.txt
@@ -8,7 +8,19 @@ include_directories(${PYTHON_INCLUDE_PATH})
 include_directories("${PROJECT_BINARY_DIR}/include" "${PROJECT_SOURCE_DIR}/OgreMain/include")
 include(${SWIG_USE_FILE})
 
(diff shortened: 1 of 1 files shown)


paroj:
turned out to be incomplete:

this always puts things into system python (/usr/lib/) - even when we want things somewhere else (/usr/local/lib)

susnux:
@paroj What do you think would be the best option?

1. Leave it as it is (requires patching it for e.g. openSUSE as the python path does not match the one required on most distributions).
2. Readd the changes from this PR, but with option to override
3. Do not implement this "guessing" logic, but simply provide a CMake option to override the path?

paroj:
I suggest to go with a CMake option named `OGRE_PYTHON_DIRECTORY` (to match what we already have for LIB &BIN).
That defaults to the currently used path.

While one could put some more work into your approach to make it work (see [e.g. here](https://github.com/opencv/opencv/blob/ad6e82942b37be8ee2c71c1d9bc7fe79cd16f7ab/cmake/OpenCVDetectPython.cmake#L180)). It seems to be simpler to just ask the user at build-time instead of implementing dozens of heuristics.

PR #2318, opened 2021-12-14
Assimp: Fix detection of assimp 5.1, upgrade bundled assimp, remove cmake workaround for CI

susnux:
This solve an issue where assimp >= 5.1 is not found by cmake, as the `package_name` is `assimp` not `ASSIMP`.

But this also updates the assimp version used for msvc builds and drops the cmake workaround for assimp 5.0, as this is fixed with 5.1.

This should be of cause discussed if dropping 5.0 support is ok, did this because then also the workaround can be dropped, but only changing the `package_name` and keep the workaround should work if 5.0 support should stay.

### DIFF
--- a/CMake/Dependencies.cmake
+++ b/CMake/Dependencies.cmake
@@ -251,11 +251,11 @@ find_package(ZLIB)
 macro_log_feature(ZLIB_FOUND "zlib" "Simple data compression library" "http://www.zlib.net" FALSE "" "")
 
 # Assimp
-find_package(ASSIMP QUIET)
-macro_log_feature(ASSIMP_FOUND "Assimp" "Needed for the AssimpLoader Plugin" "https://www.assimp.org/" FALSE "" "")
+find_package(assimp QUIET)
+macro_log_feature(assimp_FOUND "Assimp" "Needed for the AssimpLoader Plugin" "https://www.assimp.org/" FALSE "" "")
 
-if(ASSIMP_FOUND)
-  # workaround horribly broken assimp cmake
+if(assimp_FOUND)
+  # workaround horribly broken assimp cmake, fixed with assimp 5.1
   add_library(fix::assimp INTERFACE IMPORTED)
   set_target_properties(fix::assimp PROPERTIES
       INTERFACE_LINK_LIBRARIES "${ASSIMP_LIBRARIES}"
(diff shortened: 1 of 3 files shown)


CMake/Dependencies.cmake
```diff
@@ -251,20 +251,8 @@ find_package(ZLIB)
...
-  add_library(fix::assimp INTERFACE IMPORTED)
-  set_target_properties(fix::assimp PROPERTIES
-      INTERFACE_LINK_LIBRARIES "${ASSIMP_LIBRARIES}"
-      INTERFACE_LINK_DIRECTORIES "${ASSIMP_LIBRARY_DIRS}"
-  )
-  if(EXISTS "${ASSIMP_INCLUDE_DIRS}")
-    set_target_properties(fix::assimp PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${ASSIMP_INCLUDE_DIRS}")
-  endif()
-endif()
+find_package(assimp 5.1 QUIET)
```
paroj, line 254: it should still be possible to use distro packages, and only allowing the 1 month old v5.1 will not work. Therefore we need to keep a workaround of some sort

paroj:
build is broken due to https://github.com/OGRECave/ogre/blob/088ec83dc237023590c398ef07d2b1591873821b/PlugIns/Assimp/CMakeLists.txt#L10

susnux:
I think the issue with the CI is solved, lets wait for the result. The linux part is tested locally and works fine with distribution provided assimp 5.1

paroj:
Tubleweed should not be the only Distro where it works though:
https://repology.org/project/assimp/versions

previous code would work with assimp 4. I think we should at least keep assimp 5.0.1 (Ubuntu 20.04) support.

CMake/Dependencies.cmake
```diff
@@ -251,11 +251,11 @@ find_package(ZLIB)
...
+find_package(assimp QUIET)
+macro_log_feature(assimp_FOUND "Assimp" "Needed for the AssimpLoader Plugin" "https://www.assimp.org/" FALSE "" "")
 
-if(ASSIMP_FOUND)
-  # workaround horribly broken assimp cmake
+if(assimp_FOUND)
+  # workaround horribly broken assimp cmake, fixed with assimp 5.1
   add_library(fix::assimp INTERFACE IMPORTED)
   set_target_properties(fix::assimp PROPERTIES
       INTERFACE_LINK_LIBRARIES "${ASSIMP_LIBRARIES}"
```
paroj, line 261:
```suggestion
      INTERFACE_LINK_LIBRARIES "${ASSIMP_LIBRARIES};pugixml"
```

PR #2320, opened 2021-12-16
Fix plugins not installed into OGRE_PLUGINS_PATH

susnux:
Currently you can configure `OGRE_PLUGINS_PATH`, but that only changes the location written into the configuration file.
The plugins are not actually installed into that path.

This PR installs the plugins into the selected directory.


susnux:
I see the failed IOS build, so this is WIP, will have a look into it

paroj:
you additionally need to change this
https://github.com/OGRECave/ogre/blob/efbfc2bfa1fcda9464e89debf3bb44ecd67fee01/CMake/InstallResources.cmake#L38

to "${OGRE_LIB_DIRECTORY}/OGRE"

CMake/Utils/OgreConfigTargets.cmake
```diff
@@ -137,9 +131,16 @@ function(ogre_install_target TARGETNAME SUFFIX EXPORT)
...
 		FRAMEWORK DESTINATION "${OGRE_LIB_DIRECTORY}${OGRE_DEBUG_PATH}/Debug")
 	endif()
-
 endfunction(ogre_install_target)
 
+function(ogre_install_plugin TARGETNAME)
+    if (OGRE_BUILD_LIBS_AS_FRAMEWORKS)
+        ogre_install_target(${TARGETNAME} "/" TRUE)
+    else()
+        install(TARGETS ${PLUGINNAME} DESTINATION ${OGRE_PLUGINS_PATH})
```
paroj, line 140: the EXPORT part is missing here, which breaks cmakeexports

PR #2346, opened 2022-01-10
GLSupport: GLX - fix selecting maximal supported context version

stefanroellin:
As long as context_attrib[1] is larger than majorVersion, the minor Version must not be checked.

Fixes commit 680150caaf16632578e5e294df5dfc43eacd30fe


paroj:
fixed case is: 4.1 > 3.3 would return false

PR #2382, opened 2022-02-11
Add some info to Shadow Mapping doc

sercero:
- Changed name to lowercase
- Add some info on how to debug the Shadow Map
- Add some info on how to init the Shadow Cameras
- Explain the differente types of Shadow Cameras

### DIFF
--- a/Docs/src/ogre-shadows.md
+++ b/Docs/src/ogre-shadows.md
@@ -42,28 +42,28 @@ coordinates, and \f$\vec{u} = [u_1,u_2,u_3,1]^t\f$ the shadow map
 coordinates.
 
 \f[
-    \begin{aligned}
-    \left[ \begin{array}{c}
-        u_1  w_l \\
-        u_2  w_l \\
-        u_3  w_l \\
-        w_l     \end{array} \right] =  P_l M_l \left[ \begin{array}{c}
-        x_1 \\
-        x_2 \\
-        x_3 \\
-        1     \end{array} \right]\end{aligned}
+	\begin{aligned}
+	\left[ \begin{array}{c}
+		u_1  w_l \\
+		u_2  w_l \\
+		u_3  w_l \\
+		w_l     \end{array} \right] =  P_l M_l \left[ \begin{array}{c}
+		x_1 \\
+		x_2 \\
+		x_3 \\
+		1     \end{array} \right]\end{aligned}
 \f]
 
 \f[\begin{aligned}
-    \left[ \begin{array}{c}
-        y_1  w_c \\
-        y_2  w_c \\
-        y_3  w_c \\
-        w_c     \end{array} \right] =  P_c M_c \left[ \begin{array}{c}
-        x_1 \\
-        x_2 \\
-        x_3 \\
-        1     \end{array} \right]\end{aligned}\f]
+	\left[ \begin{array}{c}
+		y_1  w_c \\
+		y_2  w_c \\
+		y_3  w_c \\
+		w_c     \end{array} \right] =  P_c M_c \left[ \begin{array}{c}
+		x_1 \\
+		x_2 \\
+		x_3 \\
+		1     \end{array} \right]\end{aligned}\f]
 
 These equations can be written more concisely as:
 \f$\vec{u}w_l = P_l M_l \vec{x}\f$ and \f$\vec{y} w_c = P_c M_c \vec{x}\f$.
@@ -78,7 +78,11 @@ exploited to combat the aliasing issue.
 
 ## Depth Biasing {#DepthBias}
 
-![](images/depthbias.svg) Shadow map sample must use one float to represent a range of possible depth values. A depth sample is chosen in the middle. Any camera image point in between the two camera rays will see the geometry, and depending on distance from light will report differently on shadowed versus lit. However, every such point should be lit.
+![](images/depthbias.svg) Shadow map sample must use one float to represent a range of possible depth values.
+A depth sample is chosen in the middle.
+Any camera image point in between the two camera rays will see the geometry, 
+and depending on distance from light will report differently on shadowed versus lit.
+However, every such point should be lit.
 
 Due to the finite precision of floating point representations and
 inherent inability of one number to represent a range of values, it is
@@ -189,9 +193,9 @@ The many variants can, however, be broken up into three broad
 categories:
 1. Those that store additional information beyond a single
 float,
-2. those that divide up shadow frusta into multiple frusta to be
+2. Those that divide up shadow frusta into multiple frusta to be
 handled separately, and
-3. those that propose less naive \f$P_l\f$ and
+3. Those that propose less naive \f$P_l\f$ and
 \f$M_l\f$ to use and thereby affect the sampling distribution. 
 
 Algorithms in each category usually work quite independently and so many hybrid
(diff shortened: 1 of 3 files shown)


Docs/src/ogre-shadows.md
```diff
@@ -397,15 +400,136 @@ Non-PCF block. Note that after doing this, the uSTexWidth and
...
 
+## Improving Shadow Quality
+
+### Shadow Camera Setups {#sm_shadow_camera_setup}
+
+The default projection used when rendering shadow textures is a uniform frustum. 
+This is pretty straight forward but doesn't make the best use of the space in the shadow map 
+since texels closer to the camera will be larger, resulting in 'jaggies'. 
+There are several ways to distribute the texels in the shadow texture differently.
+Ogre is provided with several alternative shadow camera setups:
```
paroj, line 482: there is some overlap with https://ogrecave.github.io/ogre/api/latest/_shadows.html#Texture_002dbased-Shadows

sercero: True, but it is a small introduction to the code, perhaps both documents should be revised to see what should go where.

Also I wanted to ask if it is OK to add some Shader Code for the PSSM projection, it is based on the code that comes before but with modifications to make it work with PSSM...

Docs/src/ogre-shadows.md
```diff
@@ -397,15 +400,136 @@ Non-PCF block. Note that after doing this, the uSTexWidth and
...
 
-The following shows how to activate plane optimal shadow mapping given
-some pointer to a MovablePlane and a pointer to a light.
+## Debugging Shadows
+Since shadows are a difficult subject, so it is a good idea to have the Shadow Map 
+projected on a Mini-Screen where it is possible to see how the Depth Caster is performing.
+
+```cpp
+// Create vertex program definition
+auto vertexProgram = Ogre::HighLevelGpuProgramManager::getSingleton().createProgram(
```
paroj, line 409: using a .material script here would be probably more readable.

Also, I dont think you need to write any shaders yourself - the RTSS should be able to display the texture just fine.

Docs/src/ogre-shadows.md
```diff
@@ -397,15 +400,136 @@ Non-PCF block. Note that after doing this, the uSTexWidth and
...
+Ogre::Pass* firstPass = material->getTechnique(0)->getPass(0);
+firstPass->setVertexProgram("DebugVP");
+firstPass->setFragmentProgram("DebugFP");
+auto textureUnit = firstPass->createTextureUnitState();
+textureUnit->setTextureAddressingMode(Ogre::TAM_CLAMP);
+textureUnit->setTextureFiltering(Ogre::TFO_NONE);
+textureUnit->setContentType(Ogre::TextureUnitState::CONTENT_SHADOW);
+
+// Create rectangle for the Mini-Screen and attach to node
+Ogre::Rectangle2D* miniScreen = new Ogre::Rectangle2D(true);
```
paroj, line 433:
```suggestion
Ogre::Rectangle2D* miniScreen = mSceneMgr->createScreenSpaceRect(true);
```
otherwise you leak memory

Docs/src/ogre-shadows.md
paroj: > Also I wanted to ask if it is OK to add some Shader Code for the PSSM projection

sure

Docs/src/ogre-shadows.md
sercero: Yes, but I want to leave some documentation on how to do these kind of stuff and also to do things without the RTSS as well. (The ideal is to use the RTSS but it is algo good to have the knowlede of the underlying complexity).

paroj: true, but I dont think it is a good idea to teach about underlying complexity of the RTSS in a section called "Debugging Shadows".

Instead, started adding API examples to the respective script pages here:
https://github.com/OGRECave/ogre/pull/2386/files

sercero: Great. Yes, that is what I was talking about.

Docs/src/ogre-shadows.md
```diff
@@ -397,15 +400,386 @@ Non-PCF block. Note that after doing this, the uSTexWidth and
...
+#version 330 core
+
+layout (location = 0) in vec4 vertex;
+layout (location = 8) in vec2 uv0;
+
+out vec2 TexCoords;
+
+void main()
+{
+    gl_Position = vec4(vertex.x, vertex.y, 0.0, 1.0);
```
paroj, line 458: this does not account for vertical texture flipping on GL. With ShadowMaps you will not notice, but still.
See the StdQuad_Tex2_vp.glsl sample

Docs/src/ogre-shadows.md
```diff
@@ -397,15 +400,386 @@ Non-PCF block. Note that after doing this, the uSTexWidth and
...
+```cpp
+// General scene setup
+mSceneMgr->setShadowTechnique(Ogre::SHADOWTYPE_TEXTURE_ADDITIVE_INTEGRATED);
+mSceneMgr->setShadowTextureCasterMaterial(Ogre::MaterialManager::getSingletonPtr()->getByName("PSSMCaster"));
+mSceneMgr->setShadowTextureSelfShadow(true);
+mSceneMgr->setShadowFarDistance(3000);
+// 3 textures per directional light (PSSM)
+mSceneMgr->setShadowTextureCountPerLightType(Ogre::Light::LT_DIRECTIONAL, 3);
+mSceneMgr->setShadowTextureCount(3);
+mSceneMgr->setShadowTextureConfig(0, 2048, 2048, Ogre::PF_FLOAT16_RGB);
```
paroj, line 540: this wastes 66% of shadow texture memory. Use PF_DEPTH16 instead. That way you can also use the default Shadow caster material.

Docs/src/ogre-shadows.md
```diff
@@ -397,15 +400,386 @@ Non-PCF block. Note that after doing this, the uSTexWidth and
...
+				tex_border_colour 1.0 1.0 1.0 1.0
+				content_type shadow
+                filtering none
+            }
+        }
+    }
+}
+```
+
+Program definition:
```
paroj, line 596: these shaders are a bad example... Ideally one would only show how to sample the shadow textures, however
- there is vertex lighting mixed in
- there is depth biasing with magic numbers (should not be done that way)
- there is PCF filtering which requires `inverse_texture_size`, but is #defined out

Docs/src/ogre-shadows.md
```diff
@@ -397,15 +400,386 @@ Non-PCF block. Note that after doing this, the uSTexWidth and
...
+
+Material definition: `shadow_debug.material`
+```
+material ShadowDebug
+{
+	technique
+    {
+        pass 
+        {
+            vertex_program_ref DebugVP {}
```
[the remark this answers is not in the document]
paroj: replace the program_ref by "lighting off"

Docs/src/ogre-shadows.md
sercero: But this shader is based on the one from the samples...

The only change is to make it work with PSSM.

paroj: ah, I see.. the sample shader is no good either - I just did not get around to refactor the Shadows sample.

ok, then lets leave it like that and I will refactor all at once later

paroj:
could you rebase on master? (where the CI is fixed)

paroj:
there are still some whitespace errors. You can check them locally by running
```
git diff --check origin/master
```

sercero:
> there are still some whitespace errors. You can check them locally by running
> 
> ```
> git diff --check origin/master
> ```

There are still some trailing whitespaces but they are from broken lines so they are necessary.

PR #2411, opened 2022-03-21
add OGRE_PREFIX_DIR/include to cmake include dirs

v4hn:
make exported include paths consistent between pkg-config and cmake.

[OGRE.pc exports both paths](https://github.com/OGRECave/ogre/blob/1231b716446e46dd61ba9fed2d73f75edb544e65/CMake/Templates/OGRE.pc.in#L12) and there is code around that includes OGRE headers via `#include <OGRE/...>`.

Either make it clear that `OGRE/` is no valid include path (and thus invalid in pkg-config as well) or merge&release this fix please.
As a downstream user, I prefer the additional prefix for better modularity.


paroj:
OGRE.pc is wrong. `OGRE/` should not be part of the include path. It should follow whatever we do in CMake.

I scheduled changing this for 13.4, as this will probably break some downstream.

v4hn:
Thanks for the clarification. it is obviously your decision which include paths you specify as the interface. Are there downsides to the prefix, aside from the name redundancy?

An annoying consequence of this decision is that you cannot really get rid of or point out projects that include headers with the wrong prefix as long as Linux distributions bundle the folder as `/usr/include/OGRE`. They will however break with custom installs of ogre. My proposal would have kept these builds working.

thanks lots for the quick response!

paroj:
all inner Ogre #includes omit the `OGRE/` prefix, so you must have `OGRE/` as part of your includes for things to work out anyway.

If you dont really care about your project, you can always add `${OGRE_INCLUDE_DIRS}/../` to your includes, which will keep existing code working.

v4hn:
> all inner Ogre #includes omit the OGRE/ prefix, so you must have OGRE/ as part of your includes for things to work out anyway.

that's true, even though it's only a problem for some headers. most use `#include "Ogre..."`, which picks up headers relative to the current include location. some headers in subfolders actually use [`#include "../Ogre..."`](https://github.com/OGRECave/ogre/blob/d617d5ac20e0b52f2d0862ade50c24669dd219aa/OgreMain/include/Threading/OgreDefaultWorkQueueStandard.h#L29) which works without the `OGRE/` include path as well.
Most others in subfolders do not use this include style though.

paroj:
fixed by https://github.com/OGRECave/ogre/commit/dea94de963436e093e3f03d3414364e8f793f08f

v4hn:
unless I misread that commit, you accidentally removed the wrong include path (by your explanation above) in the pkgconfig file, didn't you?

paroj:
no, only the dark red part is being removed

v4hn:
You are right. Never read commits on smartphones right after waking up... Sorry for all the noise on this minor detail.

PR #2415, opened 2022-03-23
RTSS: Delete default program processors.

erikogenvik:
Otherwise there's memory leakage.

### DIFF
--- a/Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
+++ b/Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
@@ -138,6 +138,9 @@ void ProgramManager::destroyDefaultProgramProcessors()
     // removing unknown is not an error
     for(auto lang : {"glsl", "glsles", "glslang", "hlsl"})
         removeProgramProcessor(lang);
+    for (auto processor : mDefaultProgramProcessors) {
+        OGRE_DELETE processor;
+    }
     mDefaultProgramProcessors.clear();
 }
 


paroj:
you should also set the pointer to NULL - otherwise this will cause double-frees

erikogenvik:
> you should also set the pointer to NULL - otherwise this will cause double-frees

I skipped that because the next line is
```
mDefaultProgramProcessors.clear();
```
One question though is how Ogre looks at using std::unique_ptr, along with a custom deleter for handling OGRE_DELETE? I've used that to great effect in my own code base.
It really helps with cases like this, where it's easy to miss needed deletion.

paroj:
you dont need to consider OGRE_DELETE as we dont support custom memory-allocators any more.

erikogenvik:
> 

Ah, that's good to know.

My experience with unique_ptr has only been positive so far. It mainly shines in two areas:
1) It takes care of releasing memory automatically, obviously. But this also has a subtle advantage in that it also plays nicely with exceptions.
2) It greatly improves the semantics and understanding of the code, since it's now obvious where ownership lies. If there's ever a pointer in the code developers automatically knows that it points to an object that isn't owned, since only unique_ptr are used for ownership. I would argue that this is the main advantage, as it makes understanding sometimes complex structures much easier.

paroj:
yeah, you dont need to convince me. I do use unique_ptr for new code. Here, I refactored legacy code and kept changes minimal to get Ogre 13 out. Things are private now, so we can introduce unique_ptr during the 13.x cycle without worrying about API breaks.

erikogenvik:
> yeah, you dont need to convince me. I do use unique_ptr for new code. Here, I refactored legacy code and kept changes minimal to get Ogre 13 out. Things are private now, so we can introduce unique_ptr during the 13.x cycle without worrying about API breaks.

Ah, I hadn't noticed but I see now that unique_ptr is used throughout Ogre in places. Nice!

I'm preaching for the choir, literally. :)

PR #2423, opened 2022-04-06
Samples: Fixed compilation error.

erikogenvik:
Fixed ```Fresnel.h:216:48: error: invalid conversion from 'const char*' to 'Ogre::ResourceHandle' {aka 'unsigned int'} [-fpermissive]```

Also appears on the 13.3.4 tag.

### DIFF
--- a/Samples/Simple/include/Fresnel.h
+++ b/Samples/Simple/include/Fresnel.h
@@ -213,7 +213,7 @@ protected:
         mFishSplines.clear();
 
         MeshManager::getSingleton().remove("water", ResourceGroupManager::DEFAULT_RESOURCE_GROUP_NAME);
-        MaterialManager::getSingleton().unload("Examples/FresnelReflectionRefraction");
+        MaterialManager::getSingleton().unload("Examples/FresnelReflectionRefraction", RGN_DEFAULT);
     }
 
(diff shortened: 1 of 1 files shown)


paroj:
what compiler is this on? I dont oppose the change, but none of our CI systems seems to trigger this

erikogenvik:
```
gcc (SUSE Linux) 11.2.1 20220316 [revision 6a1150d1524aeda3381b2171712e1a6611d441d6]
```
Yeah, I was a bit surprised that the tests didn't pick it up. But I guess it otherwise does an implicit conversion from the char pointer into "unsigned int".

paroj:
we dont test gcc11 yet - probably it is using c++17 or something

paroj:
tried building v13.3.4 with gcc12 and c++20. worked without any issues.

which cmake do you have? I tried 3.16.3 and 3.22.1

erikogenvik:
This one got me confused. It's not caused strictly by any implicit conversion as I first assumed, it's caused by me building Ogre with "OGRE_RESOURCEMANAGER_STRICT=1", which removes the default for "group" in ResourceManager::unload.
Since the compiler then tries to match the function to call by the number of arguments, it only finds the "ResourceHandle" overloaded version, and thus fails.

PR #2433, opened 2022-04-13
Main: fix StaticGeometry::setRenderingDistance

### LINKED ISSUE #2427: StaticGeometry::setRenderingDistance(...) does not work as expected
#### System Information
- Ogre Version: 13.3
- Operating System / Platform: all
- RenderSystem: all

#### Detailled description

The description of the function StaticGeometry::setRenderingDistance says:
> Sets the distance at which batches are no longer rendered.
> @remarks
> This lets you turn off batches at a given distance. This can be useful for things like detail meshes (grass, foliage etc) and could be combined with a shader which fades the geometry out beforehand to lessen the effect.

However, tests have shown that the geometry is always rendered. In StaticGeometry::setRenderingDistance(...), two variables are set, which aren't used anywhere, which verifies that the function does not work as expected.

See also https://forums.ogre3d.org/viewtopic.php?t=56924

stefanroellin:
Setting the rendering distance on a static geometry had
no influence whether batches were skipped while rendering.

Fixes #2427

Broken since commit 060f9705c35b7278a89e03d5daf47b50e36cc319

### DIFF
--- a/OgreMain/src/OgreStaticGeometry.cpp
+++ b/OgreMain/src/OgreStaticGeometry.cpp
@@ -813,6 +813,21 @@ namespace Ogre {
         if (mLodStrategy == 0)
             return;
 
+        // Determine whether to still render
+        mBeyondFarDistance = false;
+
+        Real renderingDist = mParent->getRenderingDistance();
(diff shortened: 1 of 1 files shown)


paroj:
thanks!

PR #2463, opened 2022-05-12
fix noisy compiler warnings.

creadmefford:
fix noisy compiler warnings.


creadmefford:
apologies for this being such a noisy commit for such a simple change. i'll delete my ogre fork after this so i have clean pull requests in the future.

Issue #2482, opened 2022-05-28
BillboardChain not (always) facing camera


rhaschke:
We use a BillboardChain to show various line strips in rviz: https://github.com/ros-visualization/rviz/issues/1287#issuecomment-1140246077
However, camera updates are not correctly considered when rendering the BillboardChain: we can look "behind" the billboard, because the object is not camera-facing as it should.
Only if the same scene is rendered into a `RenderTexture` (when the mouse gets into the aabb of the BillboardChain), the camera pose is updated and the object correctly displays again.
We use Ogre::Root::renderOneFrame() to render the scene into all RenderTargets. Doesn't that call the required update() methods?
Any hints on what is missing are highly appreciated.

https://user-images.githubusercontent.com/5376030/170824502-92ab5ff0-307c-4567-a8dc-81288312a169.mp4

rhaschke:
This seems to be an issue with OGRE 1.9 only (still the default for ROS in Ubuntu 20.04). Updating to OGRE 1.12 this works!
Any ideas, how to fix this in rviz with (existing) OGRE 1.9? E.g. calling some update() manually?

paroj:
no idea. I am afraid you will have to `git bisect` for the fix.

Closing as it is fixed in master.

rhaschke:
I would like to reopen this issue: Preparing a git bisect, I noticed that the bug is still existing in OGRE 1.12.4 (Ubuntu 20.04 system version). It seemed to work last time I checked, because I explicitly updated/recreated the BillboardChain frequently.

I verified that `BillboardChain::preRender` and thus `BillboardChain::updateVertexBuffer` are frequently called for rendering, which I assume are responsible for updating the view. However, the line isn't facing the camera...

rhaschke:
No idea. Our code doesn't use `BillboardSet`.

PR #2492, opened 2022-06-07
Example illustrating #2482

rhaschke:
This is a minimal example illustrating issue #2482. Not intended for merging.
As you expected, @paroj, `BillboardChain` and `BillboardSet` act differently when moving the camera: While the set is always facing the cam as expected, the chain essentially ignores camera updates.

### DIFF
--- a/Samples/Common/src/DefaultSamplesPlugin.cpp
+++ b/Samples/Common/src/DefaultSamplesPlugin.cpp
@@ -31,6 +31,7 @@
 
 #include "AtomicCounters.h"
 #include "BezierPatch.h"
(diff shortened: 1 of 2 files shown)


paroj:
ok, I was only testing with an additional RenderTexture, which did not trigger the bug.

The offending commit is https://github.com/OGRECave/ogre/commit/13282d38453e6d2f247a46a0308046011f94f7a2
Which fixed this: https://forums.ogre3d.org/viewtopic.php?p=454995

Now we only need to figure out a way for fixing this without breaking the fixed condition.

PR #2502, opened 2022-06-29
Swig dynamic link libraries

JanGeykenORGADATA:
The problem was that I could not build the C# wrapper then some of the Ogre_Build_Components there set to false.

The script for the Python wrapper already checks if a component is activated.

I couldn't check the Java wrapper because I have no SDK installed, but the resulting file looked fine


JanGeykenORGADATA:
Sorry I overlooked the old merges. I will submit a clean branch

PR #2505, opened 2022-07-04
Make libxrandr-dev a requirement for Ubuntu

ijnek:
Thank you to the maintainers and contributors of this library. This PR is just a small fix in the installation instructions for Ubuntu.

CMake fails while following BuildingOgre.md when only the necessary packages are installed (and no recommended optional packages):

    CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
    Please set them or make sure they are set and tested correctly in the CMake files:
    X11_Xrandr_LIB (ADVANCED)

Simply installing the missing ``libxrandr`` via apt fixes the issue.

``libxrandr`` is listed as a dependency of [libsdl2-dev](https://packages.ubuntu.com/jammy/libsdl2-dev) (a recommended package), so the issue would have not been visible in CI that installs the recommended packages.

### DIFF
--- a/BuildingOgre.md
+++ b/BuildingOgre.md
@@ -44,7 +44,7 @@ For manually building the dependencies, please refer to the list below and get a
 On linux you additionally need the following system headers to build the GL, GL3+, GLES2 & Vulkan RenderSystems:
 * Ubuntu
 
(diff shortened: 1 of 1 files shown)


ijnek:
Thanks @paroj !

PR #2524, opened 2022-07-26
simplifying remove render targets in more C++11 style

Joilnen:


OgreMain/src/OgreRenderSystem.cpp
```diff
@@ -630,22 +630,15 @@ namespace Ogre {
         // Remove all the render targets. Destroy primary target last since others may depend on it.
         // Keep mRenderTargets valid all the time, so that render targets could receive
         // appropriate notifications, for example FBO based about GL context destruction.
-        RenderTarget* primary = 0;
-        for (RenderTargetMap::iterator it = mRenderTargets.begin(); it != mRenderTargets.end(); /* note - no increment */)
-        {
-            RenderTarget* current = it->second;
-            if (!primary && current->isPrimary())
```
paroj, line 637: I think you should keep the condition. While there should be only one primary window, it is not strictly enforced.

This way we will at least not leak memory if things go wrong.

paroj (changes requested):

Joilnen (commented):
keeping primary like the original code in a second commit

Joilnen (commented):
changed

Joilnen (commented):
to review

PR #2525, opened 2022-07-26
[cmake] Fix dependencies boost and freeimage in pkgconfig

JackBoosY:
https://github.com/OGRECave/ogre/blob/36c8fe6ae25fc36d039dc542669809e365cb21fb/CMake/ConfigureBuild.cmake#L176
https://github.com/OGRECave/ogre/blob/36c8fe6ae25fc36d039dc542669809e365cb21fb/CMake/ConfigureBuild.cmake#L179

The library names of boost-thread and freeimage are hardcoded in pkgconfig, which is inconsistent with cmake. This will cause verification of these two dependencies to fail when using pkgconfig.
Also correct the condition which use boost-threads.

Related: https://github.com/microsoft/vcpkg/pull/25883


CMake/ConfigureBuild.cmake
```diff
@@ -172,11 +172,19 @@ if (UNIX)
...
+      if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+        set(OGRE_ADDITIONAL_LIBS "${OGRE_ADDITIONAL_LIBS} ${Boost_THREAD_LIBRARY_DEBUG}")
+      else()
+        set(OGRE_ADDITIONAL_LIBS "${OGRE_ADDITIONAL_LIBS} ${Boost_THREAD_LIBRARY_RELEASE}")
+      endif()
     endif ()
     # there is no pkgconfig file for freeimage, so we need to add that lib manually
-    set(OGRE_ADDITIONAL_LIBS "${OGRE_ADDITIONAL_LIBS} -lfreeimage")
+    if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+      set(OGRE_ADDITIONAL_LIBS "${OGRE_ADDITIONAL_LIBS} ${FreeImage_LIBRARY_DEBUG}")
```
paroj, line 184:
```suggestion
      set(OGRE_ADDITIONAL_LIBS "${OGRE_ADDITIONAL_LIBS} ${FreeImage_LIBRARY}")
```

FreeImage_LIBRARY_DEBUG is not defined and the above covers both cases

paroj (changes requested):

CMake/ConfigureBuild.cmake
JackBoosY: Ah it should be `FreeImage_LIBRARY_DBG` / `FreeImage_LIBRARY_REL`: https://github.com/OGRECave/ogre/blob/master/CMake/Packages/FindFreeImage.cmake#L43-L44
I will fix this later.

JackBoosY:
So can we merge this now?

PR #2526, opened 2022-07-26
changing loop for C++11 style

Joilnen:
to review

### DIFF
--- a/OgreMain/src/OgreRenderSystem.cpp
+++ b/OgreMain/src/OgreRenderSystem.cpp
@@ -608,7 +608,7 @@ namespace Ogre {
     void RenderSystem::shutdown(void)
     {
         // Remove occlusion queries
-        for (auto &a : mHwOcclusionQueries)
+        for (auto q : mHwOcclusionQueries)
         {
             OGRE_DELETE a;
         }
--- a/OgreMain/src/OgreRenderSystem.cpp
+++ b/OgreMain/src/OgreRenderSystem.cpp
@@ -610,7 +610,7 @@ namespace Ogre {
         // Remove occlusion queries
         for (auto q : mHwOcclusionQueries)
         {
-            OGRE_DELETE a;
+            OGRE_DELETE q;
         }
         mHwOcclusionQueries.clear();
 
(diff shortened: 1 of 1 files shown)


OgreMain/src/OgreRenderSystem.cpp
```diff
@@ -618,10 +618,9 @@ namespace Ogre {
     void RenderSystem::shutdown(void)
     {
         // Remove occlusion queries
-        for (HardwareOcclusionQueryList::iterator i = mHwOcclusionQueries.begin();
-            i != mHwOcclusionQueries.end(); ++i)
+        for (auto &a : mHwOcclusionQueries)
```
paroj, line 621:
```suggestion
        for (auto q : mHwOcclusionQueries)
```
as this a pointer, `auto&` does not save anything. Also `q` for query is more descriptive as a running variable

Joilnen (commented):
to review

OgreMain/src/OgreRenderSystem.cpp
```diff
@@ -240,28 +240,20 @@ namespace Ogre {
     //-----------------------------------------------------------------------
     void RenderSystem::_initRenderTargets(void)
     {
-
-        // Init stats
-        for(
-            RenderTargetMap::iterator it = mRenderTargets.begin();
-            it != mRenderTargets.end();
-            ++it )
+        for(auto &&a : mRenderTargets)
```
paroj, line 243: please use `auto& p` when iterating maps.
- The name (pair) should correspond to the element type.
- rvalue at this place is confusing and has no advantage

OgreMain/src/OgreRenderSystem.cpp
Joilnen: but in this case it makes copy, using reference wouldn't be more efficient ?

paroj: no, a reference has the same cost as copying a pointer. Sometimes more as a copy tells the compiler that the original cannot change.

Joilnen:
I will resubmit

paroj:
you can also force-push to an existing branch, if that is easier for you than opening new PRs

PR #2528, opened 2022-07-27
changes to C++11 style

Joilnen:


paroj (changes requested):
as this purely is a refactoring PR, I am quite picky about style.

Please give the running variables meaningful names, instead of just "a".

PR #2532, opened 2022-07-29
changing to C++11 range loop

Joilnen:


paroj:
can you commit this in #2530 (or vice versa). I would like to squash them to have one commit per changeset (instead of per file)

PR #2533, opened 2022-07-29
changing to C++11 time style

Joilnen:


OgreMain/src/OgreRenderTarget.cpp
```diff
@@ -487,9 +487,9 @@ namespace Ogre {
     //-----------------------------------------------------------------------
     String RenderTarget::writeContentsToTimestampedFile(const String& filenamePrefix, const String& filenameSuffix)
     {
-        struct tm *pTime;
-        time_t ctTime; time(&ctTime);
-        pTime = localtime( &ctTime );
+        std::time_t t { std::time(nullptr) };
```
paroj, line 490:
```suggestion
        auto t = std::time(nullptr);
```

OgreMain/src/OgreRenderTarget.cpp
```diff
@@ -487,9 +487,9 @@ namespace Ogre {
...
+        auto pTime { std::localtime(&t) };
```
paroj, line 491:
```suggestion
        auto pTime = std::localtime(&t);
```

I know its up to taste, but I find this easier to read..

paroj:
the time changes look good now, but it seems your range based for changes ended up here as well, which is not good. (also, there is a white-space related error)

Joilnen:
it passed, after to remove that blank line

PR #2536, opened 2022-08-01
more range loops to modern C++

Joilnen:

PR #2537, opened 2022-08-01
some more range loops in modern C++

Joilnen:


paroj:
please add this to #2536

PR #2538, opened 2022-08-02
clang-tidy to range-based for with some manual naming fixup

paroj:


Joilnen:
trailing white space in OgreShaderParameter.cpp didn't pass in linux tests, is there how to config clang-tidy to care them ?

PR #2564, opened 2022-08-14
refactoring to range loops

Joilnen:


Joilnen:
I will rewrite a better patch

PR #2576, opened 2022-08-31
CI: update NDK to latest LTS

paroj:


Joilnen:
Maybe this should be, I didn't test  
nodePos.x *= float(rng()/rng.max());
nodePos.y *= float(rng()/rng.max());
nodePos.z *= float(rng()/rng.max());
and *arg1 if necessary definition maybe to make some UNUSED macro

PR #2577, opened 2022-08-31
Refactoring

Joilnen:
Just some more refactoring loops and minor changes


CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ if(POLICY CMP0072)
 endif()
 # use legacy swig mode
 if(POLICY CMP0078)
-  cmake_policy(SET CMP0078 OLD)
```
paroj, line 41: AFAIR the Python libs end with the wrong name if we do this

PR #2580, opened 2022-09-06
refactoring and some style fixing

Joilnen:


OgreMain/src/OgreAnimationTrack.cpp
```diff
@@ -196,13 +196,10 @@ namespace Ogre {
         // Pre-allocate memory
         mKeyFrameIndexMap.resize(keyFrameTimes.size());
 
-        size_t i = 0, j = 0;
```
paroj, line 199: I think the previous version of this was more readable..

Joilnen: reverted

PR #2583, opened 2022-09-09
refactoring to range loop

Joilnen:


OgreMain/src/OgreCamera.cpp
```diff
@@ -174,7 +171,7 @@ namespace Ogre {
 
         Quaternion targetWorldOrientation;
 
-        if( mYawFixed )
```
paroj, line 177: I would prefer not to do these whitespace changes manually. They are trivial to do with clang-format. I just refrained to apply them as such commit introduces lots of noise in the code (think of open pull requests) without adding much.

Joilnen: is there a .clang-format ready to use somewhere or we can use llvm with 4 space of indentation ?

paroj: added the clang-format I use to #2584. But still, I do not think we should apply it to existing code just like that..

Joilnen: applied a slightly similar format than your, I think it looks good

paroj (changes requested):
like I said initially, i dont think it make sense to format existing code just for the sake of it..

Joilnen:
Ok I got if that keeps similar the actual style would be worth to make it, anyway I am reverting or creating a new commit.

PR #2603, opened 2022-09-27
prototype to support local space bone arrays for hardware skinning.  …

creadmefford:
…i don't like how there are now 2 large (mostly unused) matrix arrays in the AutoParamDataSource variables.  also, are there other existing mechanisms to accomplish the same idea?

this does, however, allow me to support local space skinning.


creadmefford:
see https://forums.ogre3d.org/viewtopic.php?p=553591#p553591

PR #2663, opened 2022-11-10
range loop refactoring

Joilnen:


OgreMain/src/OgreRenderQueueSortingGrouping.cpp
```diff
@@ -192,12 +192,10 @@ namespace {
             collection = &mSolidsBasic;
         }
 
-
-        Technique::Passes::const_iterator i;
-        for(i = pTech->getPasses().begin(); i != pTech->getPasses().end(); ++i)
+        for(const auto& p : pTech->getPasses())
```
paroj, line 195:
```suggestion
        for(auto* p : pTech->getPasses())
```
(same for the instance below)

paroj: please also fix the other instances

Joilnen: did

OgreMain/src/OgreRenderQueueSortingGrouping.cpp
```diff
@@ -269,11 +265,9 @@ namespace {
             // Hmm, a bit hacky but least obtrusive for now
                     OGRE_LOCK_MUTEX(Pass::msPassGraveyardMutex);
             const Pass::PassSet& graveyardList = Pass::getPassGraveyard();
-            Pass::PassSet::const_iterator gi, giend;
-            giend = graveyardList.end();
-            for (gi = graveyardList.begin(); gi != giend; ++gi)
+            for (const auto& gi : graveyardList)
```
paroj, line 268:
```suggestion
            for (auto* p : graveyardList)
```
and the instance below

PR #2671, opened 2022-11-15
Ignore missing shader parameters

Tulon:
Shader compilers are apperently allowed to remove unused parameters or parameters that can be optimised away, so throwing exceptions when trying to set such a parameter is not a good idea.

I haven't found a way to disable this option programmatically from our code, so the simplest thing to do was just to change this default.

### DIFF
--- a/OgreMain/src/OgreGpuProgramParams.cpp
+++ b/OgreMain/src/OgreGpuProgramParams.cpp
@@ -748,7 +748,7 @@ namespace Ogre
     GpuProgramParameters::GpuProgramParameters() :
         mCombinedVariability(GPV_GLOBAL)
         , mTransposeMatrices(false)
-        , mIgnoreMissingParams(false)
+        , mIgnoreMissingParams(true)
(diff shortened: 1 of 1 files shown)


Tulon:
> you can use GpuProgramParameters::setIgnoreMissingParams to programmatically disable this.

I'll try. I remember I had problems trying to disable this feature programmatically, though I no longer remember the reason for that.

Tulon:
So, at which point am I supposed to call `GpuProgramParameters::setIgnoreMissingParams()`. I tried the following approach, but it doesn't help:
```
auto& hgpMgr = Ogre::HighLevelGpuProgramManager::getSingleton();
auto fProgram = hgpMgr.createProgram(fsHandle, rGrp, kShaderLanguage, gFpg);
fProgram->setSourceFile(fsFile);
fProgram->getDefaultParameters()->setIgnoreMissingParams(true);
```

paroj:
there are also per-pass parameters, so you might need
```
Pass::getGpuProgramParameters(GPT_FRAGMENT_PROGRAM)
```
etc.

PR #2676, opened 2022-11-22
Check the result of std::find to prevent erasing 'end' iterator

FrankGoyens:
It seems to be possible that the result of std::find in this context is the 'end' iterator. So I added a check.

### DIFF
--- a/Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
+++ b/Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
@@ -71,7 +71,7 @@ void ProgramManager::releasePrograms(const ProgramSet* programSet)
             continue;
 
 	const auto it = std::find(mShaderList.begin(), mShaderList.end(), prg);
-        // TODO: this check should not be necessary, but we observed strange prg.use_count() in the wild 
+        // TODO: this check should not be necessary, but we observed strange prg.use_count() in the wild
         if(it != mShaderList.end())
             mShaderList.erase(it);
         GpuProgramManager::getSingleton().remove(prg);
--- a/Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
+++ b/Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
@@ -71,6 +71,7 @@ void ProgramManager::releasePrograms(const ProgramSet* programSet)
             continue;
 
 	const auto it = std::find(mShaderList.begin(), mShaderList.end(), prg);
+        // TODO: this check should not be necessary, but we observed strange prg.use_count() in the wild 
         if(it != mShaderList.end())
             mShaderList.erase(it);
         GpuProgramManager::getSingleton().remove(prg);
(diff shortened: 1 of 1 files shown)


paroj:
did you actually encounter this? It seems to me that the only case that this can happen if `prg == NULL`. In this case we could simplify the check.

FrankGoyens:
yes I have encountered this. The application where this is being used would crash during shutdown, while the Ogre root is being deleted.
I looked at this from a debugging session, prg wasn't null at that time.
I did notice that prg at that time was a fragment program, which I believe was already erased from `mShaderList` from releasing a previous `programSet`. Would that make sense?

FrankGoyens:
At the time right before the erase crash, the use_count of prg is 1

The destructor of ResourceGroupManager hasn't been called yet at this point

paroj:
this is consistent with the case that another programSet already released this. In this case we are the only user and both
- mShaderList.erase
- GpuProgramManager::remove
can be skipped.

The interesting part is why the preceding if failed to prevent this situation. Is ProgramManager::flushGpuProgramsCache() called before we end up here?

Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
```diff
@@ -70,7 +70,9 @@ void ProgramManager::releasePrograms(const ProgramSet* programSet)
         if(prg.use_count() > ResourceGroupManager::RESOURCE_SYSTEM_NUM_REFERENCE_COUNTS + 2)
             continue;
```
paroj, line 72:
```suggestion
// TODO: investiage - this should not happen, but it does
if(prg.use_count() == 1)
    continue;
```
in case you dont feel like diving that deep here

FrankGoyens:
I see, with that in mind I will take another look 

I will do that tomorrow probably

FrankGoyens:
I did some more debugging, it seems that the `use_count` of `prg` can sometimes be 2 or even 3 right before the erase crash occurs. 

ProgramManager::flushGpuProgramsCache() isn't called at all before the crash occurs.

I don't know if this helps, when releasePrograms is called the first time during shutdown, `mShaderList` contains 7 items. All except 2 items have 2 strong references. The other two have 7 and 4 strong references respectively.

Components/RTShaderSystem/src/OgreShaderProgramManager.cpp
```diff
@@ -70,7 +70,9 @@ void ProgramManager::releasePrograms(const ProgramSet* programSet)
...
 
-        mShaderList.erase(std::find(mShaderList.begin(), mShaderList.end(), prg));
+	const auto it = std::find(mShaderList.begin(), mShaderList.end(), prg);
+        if(it != mShaderList.end())
```
paroj, line 74:
```suggestion
        // TODO: this check should not be necessary, but we observed strange prg.use_count() in the wild 
        if(it != mShaderList.end())
```

PR #2686, opened 2022-12-05
range loop refactoring

Joilnen:
and few reformat

### DIFF
--- a/OgreMain/src/OgreNumerics.cpp
+++ b/OgreMain/src/OgreNumerics.cpp
@@ -26,59 +26,56 @@ THE SOFTWARE.
 -----------------------------------------------------------------------------
 */
 
-#include "OgreNumerics.h"
 #include "OgreStableHeaders.h"
+#include "OgreNumerics.h"
+
 
 namespace Ogre
 {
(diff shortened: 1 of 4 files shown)


OgreMain/src/OgreNumerics.cpp
```diff
@@ -26,57 +26,59 @@ THE SOFTWARE.
 -----------------------------------------------------------------------------
 */
 
-#include "OgreStableHeaders.h"
```
paroj, line 29: please revert changes to this file. This does not improve code quality, but merely adds noise to the commit

Joilnen: did

OgreMain/src/OgreNumerics.cpp
```diff
@@ -40,11 +40,11 @@ namespace Ogre
 
         int i, j;
 
-        for(j=0; j<n; j++)
+        for(j = 0; j < n; j++)
```
paroj, line 43: there are still unrelated changes in this file. please revert them as well.

paroj (changes requested):

PR #2699, opened 2022-12-21
Update SDL2 and Assimp to fix compiler erros with newest MinGW-w64

markellus:
The currently used versions of SDL2 and Assimp do not compile with the newest version of MinGW-w64 (GCC 12.2+) from MSYS2. The issues where fixed in SDL2 2.24.1 and Assimp 5.2.5.

SDL2 issue:
https://github.com/libsdl-org/SDL/issues/5728
https://github.com/libsdl-org/SDL/commit/d2c4d74dd0302369251dce9375b49f394e54f76a

I can't find a ticket for Assimp but i get a lot of these errors, which disappear after updating to 5.2.5:
`ingw64/include/c++/12.2.0/bits/new_allocator.h:175:11: error: array subscript 'const aiVector3t<float>[1]' is partly outside array bounds of 'std::__cxx11::list<aiVector3t<float> > [1]' [-Werror=array-bounds]
  175 |         { ::new((void *)__p) _Up(std::forward<_Args>(__args)...); }`

### DIFF
--- a/CMake/Dependencies.cmake
+++ b/CMake/Dependencies.cmake
@@ -113,16 +113,16 @@ if(OGRE_BUILD_DEPENDENCIES AND NOT EXISTS ${OGREDEPS_PATH})
     if(MSVC OR MINGW OR SKBUILD) # other platforms dont need this
         message(STATUS "Building SDL2")
         file(DOWNLOAD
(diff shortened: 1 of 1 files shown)


paroj:
:+1: thanks

PR #2708, opened 2022-12-28
range loop refactoring

Joilnen:


OgreMain/src/OgreRoot.cpp
```diff
@@ -738,11 +732,12 @@ namespace Ogre {
         // Tell all listeners
         bool ret = true;
         for (std::set<FrameListener*>::iterator i = mFrameListeners.begin(); i != mFrameListeners.end(); ++i)
```
paroj, line 734: should be removed

PR #2740, opened 2023-01-14
range loop refactoring

Joilnen:


OgreMain/src/OgreShadowRenderer.cpp
```diff
@@ -843,6 +839,7 @@ void SceneManager::ShadowRenderer::prepareShadowTextures(Camera* cam, Viewport*
         // set the first shadow texture index for this light.
         mShadowTextureIndexLightList.push_back(shadowTextureIndex);
         shadowTextureIndex += textureCountPerLight;
+        ++i;
```
paroj, line 842: where does this come from?

Joilnen: deleted!

PR #2752, opened 2023-01-25
GLSupport: fix requested version of GL context

dmitrylyzo:
Old code produced '4.3' for OpenGL 4.1 (_VMware 16.2.5_).

Regression from f9bf70d8e4b3c3d6e2cb6f9f2cacf98e8312d22c

### DIFF
--- a/RenderSystems/GLSupport/src/win32/OgreWin32GLSupport.cpp
+++ b/RenderSystems/GLSupport/src/win32/OgreWin32GLSupport.cpp
@@ -479,8 +479,16 @@ namespace Ogre {
             break;
         default:
             profile = WGL_CONTEXT_CORE_PROFILE_BIT_ARB;
-            majorVersion = std::max(glMajorMax, 3);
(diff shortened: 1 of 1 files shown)


RenderSystems/GLSupport/src/win32/OgreWin32GLSupport.cpp
```diff
@@ -479,8 +479,19 @@ namespace Ogre {
...
-            minorVersion = std::max(glMinorMax, 3); // 3.1 would be sufficient per spec, but we need 3.3 anyway..
+            if (glMajorMax > 3)
+            {
+                majorVersion = glMajorMax;
+                minorVersion = glMinorMax;
+            }
+            else
+            {
+                majorVersion = 3;
+                if (glMajorMax < 3)
```
paroj, line 490: what is the purpose of the nested if? Cant we just hardcode 3.3 here?

paroj: yes the GL versions are fixed. Please update the PR accordingly

paroj:
:+1: thanks!

PR #2766, opened 2023-02-02
Update visuals of some old (and current) Tools documentation

sercero:
Also fix HTML formatting.

Sorry, was bored at a call conference...


paroj:
> Sorry, was bored at a call conference...

I feel you :wink: 

can you fix the whitespace errors? `git diff --check origin/master`

PR #2769, opened 2023-02-04
Devel

paroj:


GloriousEggroll:
Is there any other way to access createRenderSystemCapabilities?

I was working on rebasing parts of ROS1 to be compatible with upstream ogre and hit this error:
```
/ros_catkin_ws/src/rviz/src/rviz/ogre_helpers/render_system.cpp:184:91: error: ‘virtual Ogre::RenderSystemCapabilities* Ogre::RenderSystem::createRenderSystemCapabilities() const’ is protected within this context
  184 |     const Ogre::RenderSystemCapabilities* caps = renderSys->createRenderSystemCapabilities();
      |                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
```

paroj:
createRenderSystemCapabilities should not be called by user code. What is your usecase?

PR #2790, opened 2023-02-20
range loop refactoring

Joilnen:


OgreMain/src/OgreResourceManager.cpp
```diff
@@ -344,24 +338,18 @@ namespace Ogre {
     {
         OGRE_LOCK_AUTO_MUTEX;
 
-        ResourceMap::iterator i, iend;
-        iend = mResources.end();
-        for (i = mResources.begin(); i != iend; )
+        for (auto& r : mResources)
```
paroj, line 341: this will break, as remove() invalidates the iterator

Joilnen: reverted

PR #2801, opened 2023-03-03
Use all available processor cores to build dependencies

markellus:
Building dependencies like SDK2 and assimp currently takes an unnecessary long time, because they do not use all available processor cores. With this patch, the number of available cores is automatically detected and passed to the dependency CMake scripts. This makes the build process about 2-3 times faster on my 12 Core processor when using MinGW-w64.

### DIFF
--- a/CMake/Dependencies.cmake
+++ b/CMake/Dependencies.cmake
@@ -52,7 +52,7 @@ ProcessorCount(NPROC)
 
 # can not get processor cores, fallback to default value
 if(NPROC EQUAL 0)
-    set(N 2)
+    set(NPROC 2)
 endif()
 
 set(BUILD_COMMAND_OPTS --target install -j ${NPROC} --config ${CMAKE_BUILD_TYPE})
--- a/CMake/Dependencies.cmake
+++ b/CMake/Dependencies.cmake
@@ -48,7 +48,7 @@ set(OGREDEPS_SHARED FALSE)
 
 # get available processor cores
 include(ProcessorCount)
-ProcessorCount(N)
+ProcessorCount(NPROC)
 
 # can not get processor cores, fallback to default value
 if(N EQUAL 0)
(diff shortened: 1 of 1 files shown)


CMake/Dependencies.cmake
```diff
@@ -46,7 +46,16 @@ endif()
 # if we build our own deps, do it static as it generally eases distribution
 set(OGREDEPS_SHARED FALSE)
 
-set(BUILD_COMMAND_OPTS --target install --config ${CMAKE_BUILD_TYPE})
+# get available processor cores
+include(ProcessorCount)
+ProcessorCount(N)
```
paroj, line 51:
```suggestion
ProcessorCount(NPROC)
```

paroj (changes requested):

CMake/Dependencies.cmake
```diff
@@ -46,7 +46,16 @@ endif()
...
+
+# can not get processor cores, fallback to default value
+if(N EQUAL 0)
+    set(N 2)
+endif()
+
+set(BUILD_COMMAND_OPTS --target install -j ${N} --config ${CMAKE_BUILD_TYPE})
```
paroj, line 58:
```suggestion
set(BUILD_COMMAND_OPTS --target install -j ${NPROC} --config ${CMAKE_BUILD_TYPE})
```

CMake/Dependencies.cmake
```diff
@@ -46,7 +46,16 @@ endif()
 # if we build our own deps, do it static as it generally eases distribution
 set(OGREDEPS_SHARED FALSE)
 
-set(BUILD_COMMAND_OPTS --target install --config ${CMAKE_BUILD_TYPE})
+# get available processor cores
+include(ProcessorCount)
+ProcessorCount(N)
+
+# can not get processor cores, fallback to default value
+if(N EQUAL 0)
```
paroj, line 54:
```suggestion
if(NPROC EQUAL 0)
```

CMake/Dependencies.cmake
```diff
@@ -46,7 +46,16 @@ endif()
...
+    set(N 2)
```
paroj, line 55:
```suggestion
    set(NPROC 2)
```

markellus:
Hi, sorry I was not able to make changes in the last week, but I see everything is resolved already 👍

PR #2803, opened 2023-03-08
Use a workaround for gl_ClipDistance on Intel Windows drivers.

jordivg46:
This fixes that the vertex shader compilation would fail with error "'array without size' :  supported in geometry shaders only" on an Intel HD Graphics 4000 integrated video chip. This fix was already done in GLSLShader::compileSource 5 years ago with https://github.com/OGRECave/ogre/commit/d8d99cecf260e0441f77e6a1b1ceceea5e934ac7.


paroj:
I would rather remove the incomplete gl_ClipDistance handling in RTSS altogether, to let the correct code in GL3+ handle it

jordivg46:
Thank you for you quick response.
To avoid code duplication, it is indeed better to let GLSLShader::compileSource add the necessary gl_PerVertex block and don't add it in GLSLProgramWriter::writeOutParameters, if you think that is enough.
Some tests were crashing now too because the RenderSystemCapabilities aren't initialized, so I have too change it anyway.

PR #2827, opened 2023-04-13
Update freetype, SDL and Android NDK

mosfet80:
Update freetype, SDL and OCORNUT external libraries

### DIFF
--- a/CMake/Dependencies.cmake
+++ b/CMake/Dependencies.cmake
@@ -96,14 +96,14 @@ if(OGRE_BUILD_DEPENDENCIES AND NOT EXISTS ${OGREDEPS_PATH})
     if (NOT FREETYPE_FOUND)
         message(STATUS "Building freetype")
         file(DOWNLOAD
-            https://download.savannah.gnu.org/releases/freetype/freetype-2.12.1.tar.gz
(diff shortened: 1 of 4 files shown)


paroj:
this seems to be an issue with imgui.
The check here allows clang, but the pragma is GCC specific:
https://github.com/ocornut/imgui/blob/1ebb91382757777382b3629ced2a573996e46453/misc/freetype/imgui_freetype.cpp#L51

mosfet80:
Probably the problem is in gradle.

I also noticed some inconsistencies between the ci-build.cmake configuration file and the build logs.

Logs show:
Found PythonInterp: /usr/bin/python (found version "3.8.10")
Android NDK: APP_PLATFORM not set. Defaulting to minimum supported version android-19.

In the ci-build.cmake configuration file:
  -DANDROID_PLATFORM=android-16
..the python is not set, the choice of python therefore depends on the platform, version 3.8 is used on android and 3.10 on windows.


Another old library that should be updated is pkg-config:

Found PkgConfig: /usr/bin/pkg-config (found version "0.29.1")

version 0.29.2 (from 2017) fixes many bugs https://github.com/freedesktop/pkg-config/blob/master/NEWS

paroj:
thanks for the efforts :+1:

PR #2828, opened 2023-04-16
[GL]mCurrentContext is null

Suryavarman:
Corrects the use of the mCurrentContext pointer when it points to null.  
—————————————
Corrige l'utilisation du pointeur mCurrentContext alors qu'il pointe sur nul.

See:
https://forums.ogre3d.org/viewtopic.php?p=554471#p554471


paroj:
you can execute the test locally by running:

    git diff --check origin/master

paroj:
thanks :+1:

PR #2836, opened 2023-05-08
Update project.gradle

mosfet80:
It fixes the following issue:

#19360 Upgrade checks to Log4j 2.17.0
Issues fixed in second patch release:

#19300 Mitigations for log4j vulnerability in Gradle builds #19257 Incremental java compilation fails when renaming classname with $ character Issues fixed in first patch release:

#19058 Consider reverting breaking change about test configuration #19067 Fix multiple annotation processing issues discovered by Micronaut

### DIFF
--- a/CMake/Templates/project.gradle
+++ b/CMake/Templates/project.gradle
@@ -5,7 +5,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.3.1'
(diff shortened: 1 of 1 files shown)


paroj:
I dont think this particularly relevant, as we have no Java dependencies in Ogre. Downstream projects will have their own gradle file anyway. So this merely affects the case where someone copy/ pastes our sample project.

On the other hand this does not hurt either..

PR #2843, opened 2023-05-20
DefaultDebugDrawer: transform bones and set axes size

chillywillysoft:
I needed this for my project. It seems to work and thought it might be useful for others. However…

The docs say that Node::_getFullTransform() “should only be called by a SceneManager which knows the derived transforms have been updated before calling this method.” I do not know if that is the case when Entity::_updateRenderQueue() is called. Please advise.

The DefaultDebugDrawer::drawAxes() size parameter could come from DefaultDebugDrawer or from Entity. Rather than add a debug data member to every Entity I thought it better to have it in just the DefaultDebugDrawer. However that means to set the value, we either have to change SceneManager’s interface, or just cast the return of SceneManager::getDebugDrawer() to a DefaultDebugDrawer *, which is what I’m doing.

Not sure about my design decisions. Open to feedback.

### DIFF
--- a/OgreMain/include/OgreDefaultDebugDrawer.h
+++ b/OgreMain/include/OgreDefaultDebugDrawer.h
@@ -16,6 +16,7 @@ class _OgreExport DefaultDebugDrawer : public DebugDrawer
     ManualObject mAxes;
     int mDrawType;
     bool mStatic;
+    float mBoneAxesSize;
     void preFindVisibleObjects(SceneManager* source, SceneManager::IlluminationRenderStage irs, Viewport* v) override;
     void postFindVisibleObjects(SceneManager* source, SceneManager::IlluminationRenderStage irs, Viewport* v) override;
     void beginLines();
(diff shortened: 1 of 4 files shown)


paroj:
>  I do not know if that is the case when Entity::_updateRenderQueue() is called. Please advise.

yes transforms were already updated in `_updateSceneGraph` at this point. You can verify this by placing a breakpoint in `Node::_update` and `Node::_updateRenderQueue`.
Probably there should be a Node update sequence chart similar to what we have for render here:
https://ogrecave.github.io/ogre/api/latest/class_ogre_1_1_root.html#aeaad25268c0ad8007ce8961a32242c77


> The DefaultDebugDrawer::drawAxes() size parameter could come from DefaultDebugDrawer or from Entity. Rather than add a debug data member to every Entity I thought it better to have it in just the DefaultDebugDrawer. However that means to set the value, we either have to change SceneManager’s interface, or just cast the return of SceneManager::getDebugDrawer() to a DefaultDebugDrawer *, which is what I’m doing.
> 
> Not sure about my design decisions. Open to feedback.

these decisions are sound.

PR #2845, opened 2023-05-24
range loop refactoring

Joilnen:


Joilnen:
could we to pass the project to C++17 and to use for_each(std::execution::par,,) ?

paroj:
> could we to pass the project to C++17 and to use for_each(std::execution::par,,) ?

I try to keep Ogre1 conservative in regards to C++, thats why master only switched from C++11 to C++14, so C++17 will take a while.

However note that using parallel for here would make the code actually slower

PR #2840, opened 2023-05-13
Main: setTransformSpace() on NodeAnimationTrack

chillywillysoft:
motivation: https://forums.ogre3d.org/viewtopic.php?p=554574

PR #2848, opened 2023-05-27
Main: NodeAnimationTrack - add setUseLocalTranslation

chillywillysoft:
Allows to select local space for translations on NodeAnimationTrack.

More focused version of #2840 that doesn't affect rotations and only allows choosing between parent and local.


paroj:
after having looked at it some more:
could the same not be achieved by introducing an intermediate scene node to orient the animation? i.e.

```
node(rotated) -> animation (local) 
== 
node -> node (rotated) -> animation (parent)
```

chillywillysoft:
@paroj Yes, that would work. I didn't want to introduce the extra node. But doing it that way (with the extra node) actually makes it almost identical to how I handle Skeletal animations that cause movement, so that sounds like a good idea.

PR #2859, opened 2023-06-12
ResourceGroupManager shutdown crash

CsabaStupak:
On application close the ResourceGroupManager::_notifyAllResourcesRemoved is called which removes all resources from the loadResourceOrderMap.

When we erase the resource from the container the resource can be deleted (if the use-counter = 0). In this case D3D9HLSLProgram d-tor is called which tries to unload its resources (HighLevelGpuProgram::unloadImpl) and it calls ResourceGroupManager::_notifyResourceRemoved. This method removes these resources from the same container what we are operating above which invalidates our iterator and we crash.


paroj:
:+1:  thanks

PR #2872, opened 2023-06-25
Update project.gradle

mosfet80:
Update to last gradle 7.* version 

https://docs.gradle.org/7.4.2/release-notes.html

### DIFF
--- a/CMake/Templates/project.gradle
+++ b/CMake/Templates/project.gradle
@@ -5,7 +5,7 @@ buildscript {
         google()
     }
     dependencies {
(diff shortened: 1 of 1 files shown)


paroj:
I dont think this is relevant, as this does not fix any issue we are having

mosfet80:
The android import project time are reduced by 50% and solve a lot of bug https://docs.gradle.org/7.4.2/release-notes.html

paroj:
In contrast to e.g. SDL2, we dont ship this to users though, so as long as it works on the CI, this does not really matter

PR #2893, opened 2023-07-14
docs: expand API documentation for getByName

ribbon-otter:
this pull request clarifies that the `getByName` method in the various *X*Manager Singletons can get *X* from the corresponding script file.


ribbon-otter:
oh didn't realize the trailing whitespace was a test rather than a bulid step. running tests locally to fix that.

PR #2907, opened 2023-08-04
Fix bug any_cast for iOS clang compiler

HadesD:
iOS clang just only check literal value, it will be failed on some case

### DIFF
--- a/OgreMain/include/OgreAny.h
+++ b/OgreMain/include/OgreAny.h
@@ -220,12 +220,11 @@ namespace Ogre
     template<typename ValueType>
     ValueType * any_cast(Any * operand)
     {
+        const auto& valueTypeInfo = typeid(ValueType);
+        const auto& operandTypeInfo = operand->type();
+        // Must double check because some compiler just check once
         return operand &&
-#if OGRE_COMPILER == OGRE_COMPILER_GNUC && OGRE_COMP_VER < 450
-                (std::strcmp(operand->type().name(), typeid(ValueType).name()) == 0)
-#else
-                (operand->type() == typeid(ValueType))
-#endif
+                ((operandTypeInfo == valueTypeInfo) || (std::strcmp(operandTypeInfo.name(), valueTypeInfo.name()) == 0))
                     ? &static_cast<Any::holder<ValueType> *>(operand->mContent)->held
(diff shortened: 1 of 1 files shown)


OgreMain/include/OgreAny.h
```diff
@@ -220,12 +220,11 @@ namespace Ogre
...
+        const auto& valueTypeInfo = typeid(ValueType);
+        const auto& operandTypeInfo = operand->type();
+        // Must double check because some compiler just check once
         return operand &&
-#if OGRE_COMPILER == OGRE_COMPILER_GNUC && OGRE_COMP_VER < 450
-                (std::strcmp(operand->type().name(), typeid(ValueType).name()) == 0)
-#else
-                (operand->type() == typeid(ValueType))
-#endif
+                ((operandTypeInfo == valueTypeInfo) || (std::strcmp(operandTypeInfo.name(), valueTypeInfo.name()) == 0))
```
paroj, line 227: we certainly dont need both comparisons. We should clearly identify which is needed on which platform.

Also this might be just an issue with your code. I fixed a similar issue in e2d1380a6bf6e493e6756e8455baa6a3569a4a2f

So please also provide an example where the == comparison fails.

HadesD: I can't debug more detail on xcode like android studio (it will jump to std header to compare too, but xcode doesn't)
but when i debug to `ValueType` and `operand` has same `char* type_name = "P5WX11ProxyObjectN;", it failed

paroj: yes, pointer comparison may fail when the typeinfo is duplicated in different dynlibs. See the stackoverflow thread above.
However, this can be caused by missing dllexport on your ProxyObject and is not necessarily a bug in Ogre.

HadesD: @paroj no, i'm build for ios on STATIC lib all, so i think it not about dllexport?

paroj: this is an even stronger hint that something is off with your build. Ogre uses any_cast internally too, so if those work but your own does not it means you are somehow violating the ODR.

paroj:
superceded by #2988

PR #2908, opened 2023-08-04
range loop refactoring

Joilnen:


Components/Bites/src/OgreApplicationContextBase.cpp
```diff
@@ -446,16 +446,13 @@ void ApplicationContextBase::locateResources()
     Ogre::String sec, type, arch;
     // go through all specified resource groups
     Ogre::ConfigFile::SettingsBySection_::const_iterator seci;
```
paroj, line 448: seci is unused now

PR #2913, opened 2023-08-06
adding new light type doc

Joilnen:


paroj:
- we should avoid copy / pasting text
- that makes it four lights, while it still says three

approached differently in #2914

PR #2920, opened 2023-08-21
Update zlib dependency to 1.3

fernandotonon:
Fix build on Windows using MinGW, building the dependencies was throwing this error:

```
-- Building zlib
CMake Error at CMake/Dependencies.cmake:148 (file):
  file DOWNLOAD cannot compute hash on failed download

    status: [22;"HTTP response code said error"]
Call Stack (most recent call first):
  CMakeLists.txt:282 (include)
```
Screenshot from 2023-08-21 12-27-38

The reason is that zlib version 1.12.13 is not available anymore in this link http://zlib.net/zlib-1.2.13.tar.gz
Zlib 1.3 was released on August 18, 2023.

This PR updates the zlib dependency to version 1.13, which builds and runs fine.


paroj:
thanks for the PR. fixed via #2921

PR #2922, opened 2023-08-21
CMAKE_CL_64 is deprecated and only works with Microsoft Visual Studio

sercero:
Instead of using `CMAKE_CL_64`, we should be using `CMAKE_SIZEOF_VOID_P`.
In the official CMAKE documentation it says that `CMAKE_CL_64` is deprecated (https://cmake.org/cmake/help/latest/variable/CMAKE_CL_64.html)
And the 64-bit detection does not work in `MinGW64`.


paroj:
can you rebase this on master? CI is currently failing due to zlib..

PR #2925, opened 2023-08-22
showing pointer

Joilnen:


Joilnen:
Linux ci build is having problems to build rsimage

PR #2929, opened 2023-08-25
range loop refactoring

Joilnen:


Components/RTShaderSystem/src/OgreShaderCGProgramWriter.cpp
```diff
@@ -98,10 +98,10 @@ void CGProgramWriter::writeSourceCode(std::ostream& os, Program* program)
     writeUniformParametersTitle(os, program);
     os << std::endl;
 
-    for (const auto& param : program->getParameters())
+    for (const auto& p : program->getParameters())
```
paroj, line 101: what was wrong with that before?

Joilnen: nothing, probably I touched that code deleted and rewrite in my way in the work, I can change it back to param if you prefer.

PR #2936, opened 2023-09-06
Fix processMainThreadTasks() logic

### LINKED ISSUE #2935: Logic error in WorkQueue?
#### System Information
- Ogre Version: v14.0.1
- Operating System / Platform: Fedora 38
- RenderSystem: OpenGL 3+

#### Detailled description

Hi!

I've been investigating [this issue](https://forums.ogre3d.org/viewtopic.php?t=97107) with forum user **bro** and given the results of the git bisect we've looked more thoroughly at the code. In particular, the logic in `DefaultWorkQueueBase::processMainThreadTasks()` seems a bit odd. The code of the main loop can be simplified as follows:

```c++
        // keep going until we run out of responses or out of time
        while(true)
        {
            if(!mMainThreadTasks.empty())
            {
                std::function<void()> task = mMainThreadTasks.pop_front();
                task();
            }

            // time limit
            if (getCurrentTime() - msStart > mResposeTimeLimitMS)
                break;
        }
```

Written like this, when the task list becomes empty the loop is not exited until the timeout occurs. And even if the task list is empty when `processMainThreadTasks()` is called, the while loop is entered and spins until the timeout is reached.

I think it should be rewritten as follows:

```c++
        // keep going until we run out of responses or out of time
        while(!mMainThreadTasks.empty())
        {
            std::function<void()> task = mMainThreadTasks.pop_front();
            task();

            // time limit
            if (getCurrentTime() - msStart > mResposeTimeLimitMS)
                break;
        }
```

I've attached a proper patch that is compatible with v14.0.1 and can open a pull request if it is considered relevant.

I've tested it on my system and got equivalent performance to v13.6.2.

[commit-5076381.txt](https://github.com/OGRECave/ogre/files/12534907/commit-5076381.txt)

lpea:
Fixes #2935.

### DIFF
--- a/OgreMain/src/OgreWorkQueue.cpp
+++ b/OgreMain/src/OgreWorkQueue.cpp
@@ -173,20 +173,17 @@ namespace Ogre {
         unsigned long msCurrent = 0;
 
         // keep going until we run out of responses or out of time
-        while(true)
+        while(!mMainThreadTasks.empty())
         {
-            if(!mMainThreadTasks.empty())
+            std::function<void()> task;
             {
-                std::function<void()> task;
-                {
-                    OGRE_WQ_LOCK_MUTEX(mResponseMutex);
-                    LogManager::getSingleton().stream(LML_TRIVIAL)
-                        << "DefaultWorkQueueBase('" << mName << "') - PROCESS_MAIN_TASK";
-                    task = std::move(mMainThreadTasks.front());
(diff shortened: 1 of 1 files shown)


lpea:
@paroj I don't know what AppVeyor complains about ("non-mergeable pull request"). The build looks successful on all platforms though. Please let me know if there is anything I can/should do.

PR #2937, opened 2023-09-06
range loop refactoring

Joilnen:


Components/RTShaderSystem/src/OgreShaderGLSLProgramWriter.cpp
```diff
@@ -324,26 +324,20 @@ void GLSLProgramWriter::writeInputParameters(std::ostream& os, Function* functio
...
 {
     const ShaderParameterList& outParams = function->getOutputParameters();
-
-    ShaderParameterConstIterator itParam = outParams.begin();
-    ShaderParameterConstIterator itParamEnd = outParams.end();
-
     int vsOutLocation = 0;
 
-    for ( ; itParam != itParamEnd; ++itParam)
+    for (const auto& p : outParams)
```
paroj, line 329:
```suggestion
    for (const auto* p : outParams)
```

please use pointer type as in the other PR.

`auto&` resolves to `T*&` here

paroj: ah, my bad. This one was correct as this is a `shared_ptr`

PR #2943, opened 2023-09-14
Main: fix spelling error

chillywillysoft:
I introduced a spelling error (frustrum instead of frustum) with [this](https://github.com/OGRECave/ogre/pull/2842) commit. Hopefully fixing it isn't too breaking. I doubt many people have tested for frustums since 14 came out and updating any broken code should be trivial.

### DIFF
--- a/OgreMain/include/OgrePrerequisites.h
+++ b/OgreMain/include/OgrePrerequisites.h
@@ -333,6 +333,8 @@ namespace Ogre
     _OgreExport extern const String MOT_RECTANGLE2D;
     _OgreExport extern const String MOT_STATIC_GEOMETRY;
     _OgreExport extern const String MOT_CAMERA;
+    /// @deprecated use #MOT_FRUSTUM
+    _OgreExport extern const String MOT_FRUSTRUM;
     _OgreExport extern const String MOT_FRUSTUM;
     _OgreExport extern const String MOT_MOVABLE_PLANE;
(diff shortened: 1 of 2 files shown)


paroj:
> I doubt many people have tested for frustums 

the issue forward compatibility. Some code might be written against 14.0 due to distro packaging or similar. This must continue working with any 14.x release.

Therefore, please use this pattern:

https://github.com/OGRECave/ogre/blob/3f0a656c7065f4167ad4d9d5fcbfbe6baa569aee/Components/RTShaderSystem/include/OgreShaderSubRenderState.h#L57-L60

paroj:
For 14.1.x this still would break backwards compatibility. I will merge this in time for 14.2. Set to draft so I dont merge it by accident.

chillywillysoft:
Sounds good. But just so I understand/learn, how does this break backwards compatibility? The string didn't change, it still has the old constant name and now it has a new constant name.

paroj:
> how does this break backwards compatibility? 

imagine someone writes code against 14.1.1 (with this change) and then needs to deploy it with 14.1.0

chillywillysoft:
Ah, I think I see. Not to nitpick but if I understand correctly, what you call not backward compatible is [what semver calls](https://semver.org/#spec-item-7) "new, backward compatible"? So waiting for the next minor makes perfect sense.

PR #2963, opened 2023-11-06
add SceneNode deletion convenience methods

ribbon-otter:
This commit adds convenience method to delete a node and its attached objects, as I needed and people asked about [here](https://forums.ogre3d.org/viewtopic.php?t=28547) and [here](https://forums.ogre3d.org/viewtopic.php?t=53647). 

The use case is if you want to clear a `SceneNode` and put everything back from scratch (a simple if slower way of updating an object to reflect game state), or if you wish to stop rendering a node (and thus things attached to it can be garbage collected). 

This patch adds three new non-virtual methods to `SceneNode` that analogous to the other methods but also delete the detached objects:

`SceneNode::destroyGraphSegment`, analogous to `SceneNode::removeAndDestroyAllChildren`, except it also destroys all MovableObjects attached to `this` and to any of the descendants. In other words, all children are destroyed and any objects attached to them and any objects attached to `this`.

`SceneNode::destroyChildGraphSegment`, analogous to `SceneNode::removeAndDestroyChild`, except it also destroys the objects attached to the child. 

`SceneNode::destroyAllObjects`, analogous to `SceneNode::detatchAllObjects`, except it destroys the objects too. Unlike detatchAllObjects, this function is not virtual. I believe it is not required for it to be, since everything it does could be achieved using external methods. Moreover, I did not want to increase memory footprint of SceneNode. I called it `destroyAllObjects` rather than `detatchAndDestroyAllObjects` to keep the name short and because detatching is already considered part of destroying MovableObjects.

The naming scheme of `GraphSegment` was chosen as `removeAndDestroyAllChildrenAndAttached` seemed too long. `removeAndDestroy` in the original methods' names feel redundant to me, but best not changed as to not break the API. 

There are tests added to test the basic behaviour of the 3 added methods included in this pull request.

This code deletes MovableObjects only; I have never worked with skeletons or animations, so it may not delete those. I am not sure if deleting animations would even be desired behaviour, as animations might be more analogous to meshes than entities.

Please view the code with some skepticism, as I am unfamiliar with the Ogre code base and thus, it is greater risk that I broke expected behaviour or introduced undefined behaviour. And I am of course, happy to make any changes that are needed.

### DIFF
--- a/OgreMain/include/OgreSceneNode.h
+++ b/OgreMain/include/OgreSceneNode.h
@@ -294,7 +294,7 @@ namespace Ogre {
          * Removes and destroys all the children of this node
          * @par
          * Use this method to complete destroy a node, for example,
-         * if you want to recreate its render tree from scratch. 
+         * if you want to recreate its render tree from scratch.
          * @par
          * Does **not** destroy animations, textures, meshes associated with those movable objects
          * Does not destroy the node itself
--- a/OgreMain/include/OgreSceneNode.h
+++ b/OgreMain/include/OgreSceneNode.h
@@ -162,6 +162,12 @@ namespace Ogre {
         */
         virtual void detachAllObjects(void);
 
+        /** Detaches and destroys all objects attached to this node.
+         *
+         * Does not destroy objects attached to children of this node
+        */
+        void destroyAllObjects(void);
+
         /** Determines whether this node is in the scene graph, i.e.
             whether it's ultimate ancestor is the root scene node.
         */
@@ -269,6 +275,31 @@ namespace Ogre {
             node will be detached but will not be destroyed.
         */
         void removeAndDestroyAllChildren(void);
+
+        /** Removes and destroys the child and all movable objects attached to the child,
+         * and does the same to any children of that child node.
+         *
+         * Does **not** destroy animation, textures, meshes associated with those movable objects
+         * */
+        void destroyChildGraphSegment(const String& name);
+        ///@overload
+        void destroyChildGraphSegment(unsigned short index);
+        ///@overload
+        void destroyChildGraphSegment(SceneNode * child);
+
+        /** Destroys everything attatched to or decended from this node
+         * @par
+         * Detaches and destroys all objects attached to this node or
+         * its children.
+         * Removes and destroys all the children of this node 
+         * @par
+         * Use this method to complete destroy a node, for example,
+         * if you want to recreate its render tree from scratch.  
+         * @par
+         * Does **not** destroy animations, textures, meshes associated with those movable objects
+         * Does not destroy the node itself
+         * */
+        void destroyGraphSegment();
 
         /**
          * Load a scene from a file as children of this node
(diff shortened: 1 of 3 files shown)


ribbon-otter:
is there a command to run the trailing whitespace check locally? I have been using `make && make test` and it doesn't seem to be catching it.

paroj:
> is there a command to run the trailing whitespace check locally?

```git diff --check origin/master```

File: OgreMain/src/OgreSceneNode.cpp [hunk 1]
```diff
@@ -194,6 +194,15 @@ namespace Ogre {
         needUpdate();
     }
     //-----------------------------------------------------------------------
+    void SceneNode::destroyAllObjects(void)
+    {
+        while (getAttachedObjects().size() > 0) {
```
paroj, line 199: please use `empty()`and `front()` as below: https://github.com/OGRECave/ogre/pull/2963/files#diff-8b29158c3b96916d83c3b4bbc1c9ee083fee97bff80ae347e6401c6bc49dd5afR358

paroj:
> The naming scheme of GraphSegment

While GraphSegment technically makes perfect sense, it is not used anywhere else within ogre. Therefore I would rather keep to "Children" and "Objects" for now.

OgreMain/include/OgreSceneNode.h
```diff
@@ -270,6 +276,31 @@ namespace Ogre {
...
+         * its children.
+         * Removes and destroys all the children of this node
+         * @par
+         * Use this method to complete destroy a node, for example,
+         * if you want to recreate its render tree from scratch.
+         * @par
+         * Does **not** destroy animations, textures, meshes associated with those movable objects
+         * Does not destroy the node itself
+         * */
+        void destroyGraphSegment();
```
paroj, line 302:
```suggestion
        void destroyAllChildObjects();
```

OgreMain/include/OgreSceneNode.h
```diff
@@ -270,6 +276,31 @@ namespace Ogre {
         */
         void removeAndDestroyAllChildren(void);
 
+        /** Removes and destroys the child and all movable objects attached to the child,
+         * and does the same to any children of that child node.
+         *
+         * Does **not** destroy animation, textures, meshes associated with those movable objects
+         * */
+        void destroyChildGraphSegment(const String& name);
```
paroj, line 284:
```suggestion
        void destroyChildAndObjects(const String& name);
```

OgreMain/src/OgreSceneNode.cpp
```diff
@@ -315,6 +324,49 @@ namespace Ogre {
...
+        SceneNode* pChild = static_cast<SceneNode*>(getChild(name));
+        pChild->destroyGraphSegment();
+
+        removeChild(name);
+        pChild->getCreator()->destroySceneNode(name);
+
+    }
+
+    void SceneNode::destroyChildGraphSegment(unsigned short index) {
+        SceneNode* pChild = static_cast<SceneNode*>(getChildren()[index]);
```
paroj, line 338:
```suggestion
        SceneNode* pChild = static_cast<SceneNode*>(removeChild(index));
```
otherwise pChild might be OOB. This is an issue in the original code as well

ribbon-otter:
I believe I have made the changes you have requested

PR #2964, opened 2023-11-06
fix emscripten build on apple

Altren:
do not set Apple specific options when cross compiling to emscripten


Kristonitas:
This could be solved differently (and maybe more correctly), by pulling up the `project` function higher up. It correctly clears the `APPLE` variable. I just came upon this myself :)

Altren:
Pull request with a better fix would be great

PR #2968, opened 2023-11-13
range loop refactoring

Joilnen:


RenderSystems/GLSupport/src/OgreGLVertexArrayObject.cpp
```diff
@@ -69,13 +69,8 @@ namespace Ogre {
         if(mNeedsUpdate)
             return true;
 
-        VertexDeclaration::VertexElementList::const_iterator elemIter, elemEnd;
-        elemEnd = mElementList.end();
-
-        for (elemIter = mElementList.begin(); elemIter != elemEnd; ++elemIter)
+        for (auto& elem : mElementList)
```
paroj, line 72:
```suggestion
        for (const auto& elem : mElementList)
```

also the loop below

Issue #2949, opened 2023-10-07
SampleBrowserNDK desn't compile


Joilnen:
I got this error, even changing to newer gradle doesn't works
----
gradle assembleRelease
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=on

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'SampleBrowserNDK'.
> Could not resolve all files for configuration ':classpath'.
   > Could not find com.android.tools.build:gradle:7.3.3.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/com/android/tools/build/gradle/7.3.3/gradle-7.3.3.pom
       - https://dl.google.com/dl/android/maven2/com/android/tools/build/gradle/7.3.3/gradle-7.3.3.pom
     Required by:
         project :

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 9s

Joilnen:
Changing that for gradle 8.1.1 I got this message
---
gradle assembleRelease
Picked up _JAVA_OPTIONS: -Dawt.useSystemAAFontSettings=on

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':app'.
> Could not create an instance of type com.android.build.api.variant.impl.ApplicationVariantImpl.
   > Namespace not specified. Specify a namespace in the module's build file. See https://d.android.com/r/tools/upgrade-assistant/set-namespace for information about setting the namespace.
     
     If you've specified the package attribute in the source AndroidManifest.xml, you can use the AGP Upgrade Assistant to migrate to the namespace value in the build file. Refer to https://d.android.com/r/tools/upgrade-assistant/agp-upgrade-assistant for general information about using the AGP Upgrade Assistant.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 16s

PR #2979, opened 2023-11-29
Update project.gradle

mosfet80:
225 issues have been fixed in Gradle 8.0.


paroj:
as long as android studio works, this does not fix anything. see #2949, #2872

PR #2987, opened 2023-12-14
Static Geometry collection access in scenemanager (fixes #113)

### LINKED ISSUE #113: [OGRE-424] Static Geometry iterator in scenemanager
<b>[reporter="cyberkm", created="Tue, 19 Aug 2014 17:15:43 +0200"]</b>

<p>There is no ability to iterate over existing static geometries in scenemanager.</p>

<p>One would like expect something like SceneManager::getMoveableObjectIterator() but for static geometries.</p>

<p>As for now, there is no way to know what static geometries are created in scene manager.</p>

DrDub:
Added access to the full map of names to StaticGeometry held by the SceneManager.

Renamed the internal typedef `StaticGeometryList` to `StaticGeometryMap` to make it similar to `MoveableObjectMap` as I made it public but I didn't renamed the internal `mStaticGeometryList`.

### DIFF
--- a/OgreMain/include/OgreSceneManager.h
+++ b/OgreMain/include/OgreSceneManager.h
@@ -405,6 +405,7 @@ namespace Ogre {
         typedef std::map<String, Camera* > CameraList;
         typedef std::map<String, Animation*> AnimationList;
         typedef std::map<String, MovableObject*> MovableObjectMap;
+        typedef std::map<String, StaticGeometry* > StaticGeometryMap;
     private:
         HardwareVertexBufferPtr mInstanceBuffer;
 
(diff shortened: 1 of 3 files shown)


OgreMain/include/OgreSceneManager.h
```diff
@@ -421,8 +422,7 @@ namespace Ogre {
         /// Queue of objects for rendering
         std::unique_ptr<RenderQueue> mRenderQueue;
 
-        typedef std::map<String, StaticGeometry* > StaticGeometryList;
-        StaticGeometryList mStaticGeometryList;
+        StaticGeometryMap mStaticGeometryList;
```
paroj, line 425: please keep the previous name, as this would break API compatibility

paroj:
thanks :+1:

PR #2993, opened 2024-01-02
range loop refactoring

Joilnen:


paroj:
samples actually are pretty good for refactoring. You could also take care of things like:

- DEFAULT_RESOURCE_GROUP_NAME -> RGN_DEFAULT
- HBU_DYNAMIC_WRITE_ONLY -> HBU_CPU_TO_GPU

this would add too much noise for OgreMain, but is perfect for Samples

PR #2998, opened 2024-01-05
Update ps_1_4.cpp

AdmiralPellaeon:
added some consts


paroj:
could you also edit the header here?
https://github.com/AdmiralPellaeon/ogre/edit/patch-2/RenderSystems/GL/src/atifs/include/ps_1_4.h

I would like to avoid creating a commit that does not compile by merging those two separately.

AdmiralPellaeon:
Done :)

PR #3021, opened 2024-02-01
simpliflying constructor

Joilnen:


paroj:
as we had no test for this, I applied this locally to check row/ column order. There I noticed that coeffs is needlessly duplicated in all instances.
Fixed that and the constructor in  #3022

PR #3028, opened 2024-02-10
MeshLOD: LodOutputProviderMesh - correctly assign indexStart

paroj:


paroj:
@chillywillysoft FYI. I discovered some small typos that degraded the generated LOD, in case you were wondering

PR #3034, opened 2024-02-16
Simplifying use of ApplicationContext framework

Joilnen:
Proposal
try to avoid internal root exposing to user application as much as possible, no break the actual interface,
changed the Bootstrap.cpp tutorial if it is accept I change the rest

### DIFF
--- a/Docs/14-Notes.md
+++ b/Docs/14-Notes.md
@@ -81,7 +81,7 @@ This simplifies many advanced algorithms, such as deferred shading, where you wa
 
 The second argument of `RenderQueueListener::renderQueueStarted` now contains the current camera name and hence is always non-empty.
 
-## Rectangular Area Lights (since 14.1)
+### Rectangular Area Lights (since 14.1)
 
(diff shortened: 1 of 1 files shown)


paroj:
> try to avoid internal root exposing to user application as much as possible

why do you think this is beneficial? The Root object provides access to some important functions like the WorkQueue. Also it is a Singleton, so it is not really hidden..

paroj:
also you accidentally reverted some recent changes in master

JoilnenLeite:
> > try to avoid internal root exposing to user application as much as possible
> 
> why do you think this is beneficial? The Root object provides access to some important functions like the WorkQueue. Also it is a Singleton, so it is not really hidden..

My intention is avoid to expose root as much as possible, and to leave only ScanaManager as point of interaction with Ogre less failing points, the other features from Root I think to wrapper the in ApplicationContext, and made aplications created with this framework smaller and easier.

Joilnen:
> also you accidentally reverted some recent changes in master

My intention is avoid to expose root as much as possible, and to leave only ScanaManager as point of interaction with Ogre less failing points, the other features from Root I think to wrapper the in ApplicationContext, and made aplications created with this framework smaller and easier.

PR #3036, opened 2024-02-19
Simplifying use of ApplicationContext framework

Joilnen:
updating to last commit tried to update last commit but github closed it automatically see #3034

### DIFF
--- a/Components/Bites/include/OgreApplicationContextBase.h
+++ b/Components/Bites/include/OgreApplicationContextBase.h
@@ -292,6 +292,21 @@ namespace OgreBites
         static Ogre::String getDefaultMediaDir();
 
         /**
+         * Main loop for application
+         */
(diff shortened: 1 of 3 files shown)


paroj:
> the other features from Root I think to wrap them in ApplicationContext

I agree that ApplicationContext kind of acts as Root nowadays. However we cannot replace Root with ApplicationContext as that would force optional components (Overlay, RTSS) on users whereas Root only requires OgreMain.
Therefore, we must keep Root anyway and we would be only duplicating the API.
Hence I do not think this is worth it.

Joilnen:
that seems only a duplication but would simplify the application, because ApplicationContext could to create default, SceneManager, Overlay, Viewport and even to set RTSS leaving only for users who needs to create customized ones, this would simplify the applications from new users. But thanks to read it.

PR #3051, opened 2024-03-01
Changes in CMake files to avoid downloading bullet when it is not needed

Bimbloc:
Just a check before downloading bullet making sure the user is interested in building the ogre bullet module since otherwise bullet would not really be considered strictly necessary  unlike sdl for example.


paroj:
:+1: thanks!

paroj:
had to revert this again after taking a closer look. The issue is that Dependencies are handled before the options, as options depend on the dependencies having been built:
https://github.com/OGRECave/ogre/blob/254936ebf20ba9f80386a620803b09734bac44c4/CMakeLists.txt#L324

this leaves OGRE_BUILD_DEPENDENCIES an all or nothing setting for now

Bimbloc:
Didn´t realize that :/.
Hope it wasn't to much trouble!

PR #3052, opened 2024-03-02
Add set/get default color implementation for Bullet DebugDrawer class

wil92:
Currently, is only possible to debug OgreBullet wrapper with the default black wireframe color. The only workaround was possible if you create your own debug wrapper by inhering from the existing one and then overwriting set/get defaultColors methods involving a lot of code for such a simple thing.

### DIFF
--- a/Components/Bullet/include/OgreBullet.h
+++ b/Components/Bullet/include/OgreBullet.h
@@ -143,6 +143,8 @@ class _OgreBulletExport DebugDrawer : public btIDebugDraw
     ManualObject mLines;
     int mDebugMode;
 
+    DefaultColors colors;
+
 public:
     DebugDrawer(SceneNode* node, btCollisionWorld* world)
         : mNode(node), mWorld(world), mLines(""), mDebugMode(DBG_DrawWireframe)
@@ -158,6 +160,10 @@ public:
         if (!mLines.getSections().empty()) // begin was called
             mLines.end();
     }
+
+    void setDefaultColors(const DefaultColors& defaultColors) override { colors = defaultColors; };
+
+    DefaultColors getDefaultColors() const override { return colors; };
 
     void drawLine(const btVector3& from, const btVector3& to, const btVector3& color) override;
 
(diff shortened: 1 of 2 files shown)


paroj:
> with the default black wireframe color

the colors should not be all black by default:
https://github.com/bulletphysics/bullet3/blob/6bb8d1123d8a55d407b19fd3357c724d0f5c9d3c/src/LinearMath/btIDebugDraw.h#L40-L47

to fix the CI, you either have to ignore the new methods or rename the nested class as:
https://github.com/OGRECave/ogre/blob/1662ca06f318a704d47ae61e15914edd1830c806/OgreMain/include/Ogre.i#L734

wil92:
@paroj, sorry for my ignorance, but not sure how could I ignore the methods or rename them, can you please share some documentation that I can follow to fix the pipelines?

About the color: `m_deactivatedObject` should be equal to `btVector3(0, 1, 0)` (green), but instead I got it rendered in black, this is why I said black. In any case, the idea is to be able to manipulate this color from the Ogre wrapper.

paroj:
> but not sure how could I ignore the methods or rename them

the renaming snipped is linked in my post

> should be equal to btVector3(0, 1, 0) (green), but instead I got it rendered in black,

I looked into this, as it should have been working and its a typo:

https://github.com/OGRECave/ogre/pull/3053/commits/f25eb7f8ee82aa70efac552ec7ddd89d41239aee

if the default colours work, is this PR still relevant?

wil92:
@paroj, nice that you found the error. I don't need the change I made, but still, I think it could be good if someone would like to change the default color values. I was trying to fix the pipeline,  but I honestly never worked with SWIG before, and I have no idea what to change to make it work. And I can't reproduce the error locally.

paroj:
ok, lets keep it as a draft for now then. It would be difficult to merge now as the change breaks ABI in 14.2.x. I can come back for 14.3 to this.

paroj:
I dont think this is an important enough feature to justify spending the time wrangling with SWIG

PR #3061, opened 2024-03-09
Main: Material - compile before loading techs

erikogenvik:
If there are techniques that use shaders they are all "unsupported" until compilation has happened.

### DIFF
--- a/OgreMain/src/OgreMaterial.cpp
+++ b/OgreMain/src/OgreMaterial.cpp
@@ -127,15 +127,16 @@ namespace Ogre {
     //-----------------------------------------------------------------------
     void Material::loadImpl(void)
     {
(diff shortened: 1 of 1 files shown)


paroj:
this basically undoes 934950fa0c93a054d72e0a8df3665f1018fabe93

however, I now see why it was incomplete.. need to think about it some more..

erikogenvik:
Yes, the issue was introduced through that commit. Although it's still retaining the change where compilation now happens in "load" rather than "prepare".

paroj:
it compilation needs to be done in prepare though - otherwise the prepare event is not correctly propagated.

I did not think that commit through.

closing in favor of full revert in #3062

PR #3065, opened 2024-03-17
fix mismatch operand type for operator ==

niu2x:
variable `format` 's type is `PixelFormat`, it shouldn't compared to `GL_NONE`.

it  is  mismatched.

i change it to `GLformat`


paroj:
thanks!

PR #3100, opened 2024-05-08
Font: Memory leak

roiqd03:
If HAVE_FREETYPE macro is not defined, the memory allocated in buffer is not freed

### DIFF
--- a/Components/Overlay/src/OgreFont.cpp
+++ b/Components/Overlay/src/OgreFont.cpp
@@ -528,6 +528,13 @@ namespace Ogre
                 // Advance a column
                 if(width)
                     l += (width + char_spacer);
+
+#ifndef HAVE_FREETYPE
+                if (buffer != NULL)
+                {
+                    STBTT_free(buffer, font.userdata);
+                }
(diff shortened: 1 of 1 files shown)


Components/Overlay/src/OgreFont.cpp
```diff
@@ -528,6 +528,18 @@ namespace Ogre
                 // Advance a column
                 if(width)
                     l += (width + char_spacer);
+
+#ifndef HAVE_FREETYPE
+                if (buffer != NULL)
+                {
+                    int ix0, iy0, ix1, iy1;
+                    stbtt__bitmap gbm;
+                    stbtt_GetGlyphBitmapBoxSubpixel(&font, stbtt_FindGlyphIndex(&font, cp), scale, scale, 0.0f, 0.0f, &ix0, &iy0, &ix1, &iy1);
```
paroj, line 537: why are you calling this?

roiqd03: My bad, I missunderstood it, now I think it is okay

Issue #2432, opened 2022-04-12
Wayland Support


ciandonovan:
When building OGRE with the flag -DOGRE_GLSUPPORT_USE_EGL=1 to attempt to get it to run natively on Wayland, the sample applications still call on functions that require GLX or expect an X11 display socket and attempt to connect to that instead.

There was some discussion on your forums here (https://forums.ogre3d.org/viewtopic.php?t=96208) where others also attempted this, but from what I could see it never got resolved.

The application I'm trying to build is RViz, which uses the OGRE engine, and my aim is to have it work natively with the Weston Wayland compositor. Windows' WSL uses Weston internally for graphical Linux applications, and there's footage here (https://devblogs.microsoft.com/commandline/the-initial-preview-of-gui-app-support-is-now-available-for-the-windows-subsystem-for-linux-2/) of RViz running on Weston in WSL. However, I'm not familiar with how WSL works internally and whether they in fact use XWayland for compatibility.

Does there exist an option in OGRE to disable the X11/GLX components and have it use EGL/Wayland exclusively, or would that require significant refactoring to achieve?

paroj:
to run Ogre on Wayland, we also need a `WaylandWindow` and `WaylandEGLSupport` implementation for EGL - similarly to what we have for X11 or Android right now:
https://github.com/OGRECave/ogre/tree/master/RenderSystems/GLSupport/src/EGL

as you can see there is currently none, so feel free to write one and contribute :smile:

ciandonovan:
Thanks for pointing me in the right direction!

So writing a OgreWaylandEGLSupport.cpp/OgreWaylandEGLWindow.cpp and replicating the functionality from the other EGL render systems with Wayland-equivalent functions should be sufficient for Wayland-native windowing? Or are there other areas of ORGE that would also need refactoring?

paroj:
yes, that would be sufficient

wjwwood:
Just checking in here, @ciandonovan did you manage to make any progress on this?

ciandonovan:
Unfortunately not, we'd love to use RViz plugins on Weston/Wayland in an embedded usecase, but this is the blocker. I saw it was added to the milestones for the next major release so hopeful there's movement somewhere, especially now since more distros are moving the Wayland by default too.

joakimono:
I’m working on an attempt to achieve this. I have pushed my current progress on this in a forked repo on my github account. There are several issues that needs to be fixed before a pull request is ready. 

[...]

PR #3142, opened 2024-06-21
Add Wayland support

joakimono:
This PR fixes issue #2432

Implement support for Wayland with EGL in GLSupport.
The functionality is enabled at compile type by cmake option `OGRE_GLSUPPORT_USE_WAYLAND`, provided that EGL headers are found. Note that when Wayland is enabled, X11 is unavailable.
Some new packages will need to be installed to compile it (tested on Debian 12): `libwayland-dev libwayland-egl wayland-protocols pkg-config`
For Qt (in addition to default Core and Gui): At least also `qtwayland5` and `qttools5-private-dev`. Tested with Qt 5.15.8.

Once compiled and installed, the feature can be tested with the sample browser.
It is important to explicitly select wayland when starting the application as follows:
`SDL_VIDEODRIVER=wayland ./SampleBrowser`

Likewise when using ApplicationContextQt: `./MyOgreQt -platform wayland` or `QT_QPA_PLATFORM=wayland ./MyOgreQt`. *Note*: Ogre Qt and Wayland does not work yet.

This new feature has been tested on Debian 12 on WSL2. There are several things that should be tested and fixed before merging.
- The SampleBrowser looses mouse capture after a while for some unknown reason. Keyboard events are still functional.
- Maximising the window causes window decorations (shadows) to not update. It may be an external issue or WSL-specific.
- I do not understand the parent window mechanisms used in X11 and whether this should be properly replicated with wayland.
- How should the window creation/deletion etc. be tested for correctness/robustness?
- The Wayland support and window impl files use a different order of initialisation compared to the X11 counterpart. It may not be an issue, but it should be checked. The reason for reorganisation is due to both the expected order of initialisation by wayland, but also SDL. In particular, there is a workaround also used by `bgfx` to be able to create egl surface external to SDL (see link to issue in `ApplicationContextSDL.cpp` below).


BuildingOgre.md
```diff
@@ -44,7 +44,11 @@ For manually building the dependencies, please refer to the list below and get a
 On linux you additionally need the following system headers to build the GL, GL3+, GLES2 & Vulkan RenderSystems:
 * Ubuntu
 
-    sudo apt-get install libgles2-mesa-dev libvulkan-dev glslang-dev libxrandr-dev
+    sudo apt-get install libgles2-mesa-dev libvulkan-dev glslang-dev
+    # X11:
+    sudo apt-get install libxrandr-dev
+    # Wayland:
+    sudo apt-get install libwayland-dev libwayland-egl pkg-config
```
joakimono, line 51: Add `wayland-protocols`

Components/Bites/src/OgreApplicationContextSDL.cpp
```diff
@@ -68,7 +72,36 @@ NativeWindowPair ApplicationContextSDL::createWindow(const Ogre::String& name, O
...
+
+        // https://github.com/libsdl-org/SDL/issues/5386
+        // https://github.com/bkaradzic/bgfx/blob/00fa5ad179f5aa13c1e44d0bcbccdc535aba2d00/examples/common/entry/entry_sdl.cpp#L53-L64
+        if(!win_impl)
+        {
+          Ogre::LogManager::getSingleton().logMessage("[SDL] Creating wl_egl_window");
+          int width, height;
+          SDL_GetWindowSize(ret.native, &width, &height);
+          wl_surface* surface = wmInfo.info.wl.surface;
+          win_impl = wl_egl_window_create(surface, width, height);
```
paroj, line 89: I think we should do this unconditionally in `WaylandEGLWindow`. This way we dont need to use `SDL_GetWindowData` at all and we OgreBites will not depend on Wayland.

Components/Bites/src/OgreApplicationContextSDL.cpp
```diff
@@ -68,7 +72,36 @@ NativeWindowPair ApplicationContextSDL::createWindow(const Ogre::String& name, O
...
+          SDL_SetWindowData(ret.native, "wl_egl_window", win_impl);
+        }
+        p.miscParams["externalDisplay"] = Ogre::StringConverter::toString(size_t(wmInfo.info.wl.display));
+        p.miscParams["externalSurface"] = Ogre::StringConverter::toString(size_t(wmInfo.info.wl.surface));
+        p.miscParams["externalWindowHandle"] = Ogre::StringConverter::toString(size_t(win_impl)); //Ogre::StringConverter::toString(size_t(wmInfo.info.wl.egl_window));
+
+        p.miscParams["externalXdgSurface"] = Ogre::StringConverter::toString(size_t(wmInfo.info.wl.xdg_surface));
```
paroj, line 96: this stuff is only needed for fullscreen toggling, right?

I would recommend omitting its implementation in Ogre and instead utilizing SDL_SetWindowFullscreen. Our goal is to shift all windowing-related responsibilities away from Ogre, leaving it responsible only for managing the EGL context and render size

Components/Bites/src/OgreApplicationContextSDL.cpp
```diff
@@ -68,7 +72,36 @@ NativeWindowPair ApplicationContextSDL::createWindow(const Ogre::String& name, O
...
+        p.miscParams["externalXdgToplevel"] = Ogre::StringConverter::toString(size_t(wmInfo.info.wl.xdg_toplevel));
+#endif
```
paroj, line 98:
```suggestion
#else
OgreAssert(false, "recompile with Ogre with Wayland support");
#endif
```

RenderSystems/GLSupport/CMakeLists.txt
```diff
@@ -116,6 +134,16 @@ target_include_directories(OgreGLSupport PUBLIC
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/GLSL>"
     PRIVATE "$<BUILD_INTERFACE:${NATIVE_INCLUDES}>")
 
+  if(OGRE_GLSUPPORT_USE_WAYLAND)
+    target_compile_definitions(OgreGLSupport PUBLIC OGRE_WAYLAND)
+    # TODO should be part of include file?
+
+    add_dependencies(OgreGLSupport xdg_shell)
```
paroj, line 141: why is xdg_shell needed?

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLSupport.cpp
```diff
@@ -0,0 +1,222 @@
...
+void WaylandEGLSupport::output_handle_geometry(
+    void *data, wl_output *wl_output,
+    int32_t x, int32_t y,
+    int32_t physical_width, int32_t physical_height,
+    int32_t subpixel,
+    const char *make, const char *model,
+    int32_t output_transform)
+{}
+
+void WaylandEGLSupport::output_handle_mode(
```
paroj, line 61: I believe it's best to avoid implementing this feature within Ogre. As with fullscreen I would rather direct users to SDL_GetDisplayMode

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLWindow.cpp
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    initNativeCreatedWindow(miscParams);
+
+    mGLSupport->doInit();
+
+
+    if (!mEglConfig)
+    {
+        int minAttribs[] = {EGL_RED_SIZE,
```
paroj, line 315: it should be sufficient to specify those:
https://github.com/OGRECave/ogre/blob/b9fbf7da95a9de887944de0633ea77983988c375/RenderSystems/GLSupport/src/EGL/Emscripten/OgreEmscriptenEGLWindow.cpp#L318-L334

BuildingOgre.md
joakimono: No longer needed

Components/Bites/src/OgreApplicationContextSDL.cpp
joakimono: Done.

Components/Bites/src/OgreApplicationContextSDL.cpp
joakimono: Yes, I'll remove it.

RenderSystems/GLSupport/CMakeLists.txt
joakimono: It was needed for fullscreen stuff and xdg_ data structures. Removed in new commit.

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLSupport.cpp
joakimono: Removed.

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLWindow.cpp
joakimono: I added your suggestion. Your suggestion refer to some member attributes, e.g. mMaxBufferSize. I added these as variables local to the function.

joakimono:
Regarding loosing mouse motion events: It is lost if the user clicks on the window title line (or moves the window) and then clicks somewhere inside the window.

RenderSystems/GLSupport/CMakeLists.txt
```diff
@@ -85,6 +101,7 @@ elseif (UNIX)
 
     set(PLATFORM_LIBS ${X11_LIBRARIES} ${X11_Xrandr_LIB} ${OPENGL_glx_LIBRARY})
   endif()
+  #TODO: this should not be included if OGRE_GLSUPPORT_USE_WAYLAND
```
joakimono, line 104: Fix this.

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLWindow.cpp
```diff
@@ -0,0 +1,333 @@
...
+        {
+            StringVector tokens = StringUtil::split(opt->second, " :");
+
+            mGLSupport->mWlSurface = (wl_surface*)StringConverter::parseSizeT(tokens[0]);
+        }
+    }
+    mNativeDisplay = mGLSupport->getNativeDisplay();
+}
+
+void WaylandEGLWindow::createNativeWindow(uint& width, uint& height, String& title)
```
paroj, line 93: I would just assume that the window is provided by "parentWindowHandle" above. This way we dont have to deal with the compositor at all here.

Components/Bites/src/OgreApplicationContextQt.cpp
```diff
@@ -131,6 +144,23 @@ namespace OgreBites
...
+
+        if (!nativeInterface){
+          Ogre::LogManager::getSingleton().logMessage("[Qt] Native interface is nullptr!");
+        }
+
+        auto display = static_cast<wl_display*>(nativeInterface->nativeResourceForWindow("display", nullptr));
+        auto surface = static_cast<wl_surface*>(nativeInterface->nativeResourceForWindow("surface", window));
+
+        // Is window (QWindow) actually an instance of wl_egl_window?
+        p.miscParams["externalWindowHandle"] = Ogre::StringConverter::toString(size_t(window));
```
paroj, line 159: are display and surface not sufficient for integration as with SDL?

Components/Bites/src/OgreApplicationContextSDL.cpp
```diff
@@ -68,7 +68,22 @@ NativeWindowPair ApplicationContextSDL::createWindow(const Ogre::String& name, O
     p.miscParams["sdlwin"] = Ogre::StringConverter::toString(size_t(ret.native));
 
 #if OGRE_PLATFORM == OGRE_PLATFORM_LINUX
-    p.miscParams["externalWindowHandle"] = Ogre::StringConverter::toString(size_t(wmInfo.info.x11.window));
+    if (wmInfo.subsystem == SDL_SYSWM_WAYLAND)
+    {
+#ifdef OGRE_WAYLAND
```
paroj, line 73: I just realized that everything can be done at runtime here and we can move the assert to GLSupport

Components/Bites/src/OgreApplicationContextSDL.cpp
```diff
@@ -68,7 +68,22 @@ NativeWindowPair ApplicationContextSDL::createWindow(const Ogre::String& name, O
...
+        Ogre::LogManager::getSingleton().logMessage("[SDL] Creating Wayland window");
+
+        p.miscParams["externalDisplay"] = Ogre::StringConverter::toString(size_t(wmInfo.info.wl.display));
```
paroj, line 76:
```suggestion
        p.miscParams["externalWlDisplay"] = Ogre::StringConverter::toString(size_t(wmInfo.info.wl.display));
```

paroj: and then assert in X11 window, if this parameter is present

Components/Bites/src/OgreWindowEventUtilities.cpp
```diff
@@ -37,7 +37,9 @@ THE SOFTWARE.
 #  endif
 #  include <windows.h>
 #elif OGRE_PLATFORM == OGRE_PLATFORM_LINUX
```
paroj, line 39:
```suggestion
#elif OGRE_PLATFORM == OGRE_PLATFORM_LINUX && !defined(OGRE_WAYLAND)
```
as we are not going to write a wayland alternative for this

OgreMain/include/OgreRenderTarget.h
```diff
@@ -218,7 +218,7 @@ namespace Ogre {
 
             | Name        | Description                        |
             |-------------|------------------------------------|
-            | WINDOW      | The native window handle. (X11 Window XID/ HWND / NSWindow*) |
+            | WINDOW      | The native window handle. (Wayland wl_egl_window* /X11 Window XID/ HWND / NSWindow*) |
```
paroj, line 221: does it make sense to expose wl_egl_window*? Also you named it WL_WINDOW below

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLSupport.cpp
paroj: fore reference, this is the WIP code I had for this: https://github.com/OGRECave/ogre/pull/3144

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLWindow.cpp
joakimono: Agree.

Components/Bites/src/OgreApplicationContextQt.cpp
joakimono: Yes it is! I will remove it. There are still other issues with this implementation (and not yet functional), though.

Components/Bites/src/OgreApplicationContextSDL.cpp
joakimono: I understand. My latest commit does this.

Components/Bites/src/OgreApplicationContextSDL.cpp
joakimono: Done!

Components/Bites/src/OgreWindowEventUtilities.cpp
joakimono: Done

OgreMain/include/OgreRenderTarget.h
joakimono: The `WINDOW` parameter is only used in WindowEventUtilities. As long as we do not need to expose this for users of Ogre, there is no point in exposing `WINDOW` or any other pointers for that matter. I removed them in the latest commit.

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLSupport.cpp
joakimono: I tested your WIP code and it works fine with a wayland window.

joakimono:
With Wayland, if fullscreen is selected, but video mode is not selected to be the native resolution of the monitor, there will be `alpha channel` artifacts: e.g. windows below overlay elements will be partly visible. All open GL renderers are affected.

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLWindow.cpp
```diff
@@ -0,0 +1,297 @@
...
+{
+    mEglDisplay = mGLSupport->getGLDisplay();
+
+    mGLSupport->mWlRegion = wl_compositor_create_region(mGLSupport->mWlCompositor);
+    wl_region_add(mGLSupport->mWlRegion, 0, 0, width, height);
+    wl_surface_set_opaque_region(mGLSupport->mWlSurface, mGLSupport->mWlRegion);
```
paroj, line 73: do we have to do this ourselves? I would assume the surface is sufficiently configured by SDL/ Qt..

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLWindow.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+        if ((opt = miscParams->find("vsyncInterval")) != end)
+            vsyncInterval = StringConverter::parseUnsignedInt(opt->second);
+
+        if ((opt = miscParams->find("gamma")) != end)
+        {
+            mHwGamma = StringConverter::parseBool(opt->second);
+        }
+
+        if ((opt = miscParams->find("title")) != end)
```
paroj, line 217: title is unused now

RenderSystems/GLSupport/src/EGL/X11/OgreX11EGLWindow.cpp
```diff
@@ -94,6 +94,10 @@ namespace Ogre {
             NameValuePairList::const_iterator opt;
             NameValuePairList::const_iterator end = miscParams->end();
 
+            if ((opt = miscParams->find("externalWlDisplay")) != end)
```
paroj, line 97: you can do the same in OgreWaylandWindow for "externalWindowHandle" and "parentWindowHandle" as those indicate an X11 Window now

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLWindow.cpp
```diff
@@ -0,0 +1,297 @@
...
+    //mNativeDisplay = mGLSupport->getNativeDisplay();
+
+    if (miscParams)
+    {
+        NameValuePairList::const_iterator opt;
+        NameValuePairList::const_iterator end = miscParams->end();
+
+        if ((opt = miscParams->find("externalWlDisplay")) != end)
+        {
+          StringVector tokens = StringUtil::split(opt->second, " :");
```
paroj, line 50: this is not needed. The parameter will be always a single number.

Components/Bites/src/OgreApplicationContextSDL.cpp
```diff
@@ -68,7 +68,20 @@ NativeWindowPair ApplicationContextSDL::createWindow(const Ogre::String& name, O
...
 
 #if OGRE_PLATFORM == OGRE_PLATFORM_LINUX
-    p.miscParams["externalWindowHandle"] = Ogre::StringConverter::toString(size_t(wmInfo.info.x11.window));
+    if (wmInfo.subsystem == SDL_SYSWM_WAYLAND)
+    {
+#ifdef OGRE_WAYLAND
+        Ogre::LogManager::getSingleton().logMessage("[SDL] Creating Wayland window");
+
+        p.miscParams["externalWlDisplay"] = Ogre::StringConverter::toString(size_t(wmInfo.info.wl.display));
+        p.miscParams["externalSurface"] = Ogre::StringConverter::toString(size_t(wmInfo.info.wl.surface));
```
paroj, line 77:
```suggestion
        p.miscParams["externalWlSurface"] = Ogre::StringConverter::toString(size_t(wmInfo.info.wl.surface));
```
for consistency

Components/Bites/src/OgreApplicationContextSDL.cpp
```diff
@@ -68,7 +68,20 @@ NativeWindowPair ApplicationContextSDL::createWindow(const Ogre::String& name, O
     p.miscParams["sdlwin"] = Ogre::StringConverter::toString(size_t(ret.native));
 
 #if OGRE_PLATFORM == OGRE_PLATFORM_LINUX
-    p.miscParams["externalWindowHandle"] = Ogre::StringConverter::toString(size_t(wmInfo.info.x11.window));
+    if (wmInfo.subsystem == SDL_SYSWM_WAYLAND)
+    {
+#ifdef OGRE_WAYLAND
```
paroj, line 73: the #ifdef is not needed any more

Components/Bites/src/OgreApplicationContextQt.cpp
```diff
@@ -131,6 +144,21 @@ namespace OgreBites
 
         p.miscParams["externalWindowHandle"] = std::to_string(size_t(window->winId()));
 
+#ifdef OGRE_WAYLAND
```
paroj, line 147: rewrite without #ifdef as with SDL2

BuildingOgre.md
```diff
@@ -44,21 +44,29 @@ For manually building the dependencies, please refer to the list below and get a
 On linux you additionally need the following system headers to build the GL, GL3+, GLES2 & Vulkan RenderSystems:
 * Ubuntu
 
-    sudo apt-get install libgles2-mesa-dev libvulkan-dev glslang-dev libxrandr-dev
+    sudo apt-get install libgles2-mesa-dev libvulkan-dev glslang-dev
+    # X11:
```
paroj, line 48:
```suggestion
    * X11:
```
see: https://github.com/OGRECave/ogre/blob/1d077a2f2d23c1fa7d0e0aa6622835795ccf9fe8/BuildingOgre.md#linux

paroj:
I think this is in a pretty good shape for merging now. The comment spam is mostly a checklist for myself as it might take me a few weeks until I can fix this up. Of course you can do the remaining fixes yourself too :)

other things to watch out:
- scroll over the diff and check if everything looks good
- remove commented out code
- check that all docstrings are correct

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLWindow.cpp
joakimono: Removed

RenderSystems/GLSupport/src/EGL/X11/OgreX11EGLWindow.cpp
joakimono: Done

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLWindow.cpp
joakimono: Removed and also changed signature of affected function

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLWindow.cpp
joakimono: This is indeed not needed. By not doing this the `alpha channel` artifacts are removed too.

Components/Bites/src/OgreApplicationContextSDL.cpp
joakimono: Done

Components/Bites/src/OgreApplicationContextQt.cpp
joakimono: Done

joakimono: One problematic thing with this cpp-file is the need for including a private Qt header, which is unstable and requires the exact version number of Qt to include it.

RenderSystems/GLSupport/src/EGL/Wayland/OgreWaylandEGLSupport.cpp
```diff
@@ -0,0 +1,171 @@
...
+    if (!mGLDisplay)
+    {
+        if (!mNativeDisplay)
+            mNativeDisplay = getNativeDisplay();
+        struct wl_registry* registry = wl_display_get_registry(mNativeDisplay);
+        wl_registry_add_listener(registry, &WaylandEGLSupport::mRegistryListener, this);
+        wl_display_dispatch(mNativeDisplay);
+        wl_display_roundtrip(mNativeDisplay);
+
+        if (mWlCompositor == nullptr)
```
paroj, line 160: mWlCompositor is unused

joakimono: You are right, it is no longer used! I have removed it and affected code from the implementation.

joakimono:
`TestContext` fails with a segmentation fault at `RenderSystem::_createRenderWindow "OGRE VTest Context"`. It should be resolved before merging.

joakimono:
ApplicationContextQt is still not functional. An error similar to: https://bugreports.qt.io/browse/QTBUG-123866
But it may be some missing configuration in our end. I want to try a newer version of Qt also.

paroj:
> ApplicationContextQt is still not functional. An error similar to: https://bugreports.qt.io/browse/QTBUG-123866 But it may be some missing configuration in our end. I want to try a newer version of Qt also.

it would be totally fine by me if Wayland support would be only available with Qt >= 6.5

joakimono:
> > ApplicationContextQt is still not functional. An error similar to: https://bugreports.qt.io/browse/QTBUG-123866 But it may be some missing configuration in our end. I want to try a newer version of Qt also.
> 
> it would be totally fine by me if Wayland support would be only available with Qt >= 6.5

Ok. I will investigate further to see if a newer Qt version is beneficial

Components/Bites/src/OgreApplicationContextQt.cpp
```diff
@@ -129,7 +135,27 @@ namespace OgreBites
         p.width = window->width();
         p.height= window->height();
 
-        p.miscParams["externalWindowHandle"] = std::to_string(size_t(window->winId()));
+        if (QGuiApplication::platformName() != "wayland") {
```
joakimono, line 138: Need to investigate whether other platform names may be relevant too. 
https://github.com/qt/qtwayland/blob/dev/README

paroj:
> `TestContext` fails with a segmentation fault at `RenderSystem::_createRenderWindow "OGRE VTest Context"`. It should be resolved before merging.

I guess the problem is that it does not pass any miscParams:
https://github.com/OGRECave/ogre/blob/70d7c82530add72e1fe88e976a6c854352b1fce5/Tests/VisualTests/Context/src/TestContext.cpp#L126

which is a nice test case actually. We should assert instead of segfaulting in this case as on Vulkan:
https://github.com/OGRECave/ogre/blob/70d7c82530add72e1fe88e976a6c854352b1fce5/RenderSystems/Vulkan/src/OgreVulkanWindow.cpp#L444

but yeah.. for merging we should either revert the CI to X11 or fix that

I guess that keeping CI with X11 could be actually beneficial as X11 will probably remain the common case for at least some more years*. We could extend the "custombuild" target to merely compile-test wayland for now.

joakimono:
> Ok. I will investigate further to see if a newer Qt version is beneficial

Some of the interface is nicer for Qt 6.5+, but there will still be a need for including a private header, even at Qt 6.7. Specifically the `wl_surface` is not publically exposed in the new native interface.
ref: https://planet.kde.org/david-redondo-2022-12-09-wayland-native-interface-in-qt-6-5/

Since an implementation need to use QtGui::Private and deal with version-specific includes anyways, I do not see the point of restricting Ogre with Wayland to 6.5+ (unless there are other features unknown to me).
Qt 6 also requires c++17.

.github/workflows/ci-build.yml
```diff
@@ -111,4 +114,8 @@ jobs:
     - name: Build
       run: |
         cmake -Bbuild -DOGRE_BUILD_DEPENDENCIES=OFF -DSWIG_EXECUTABLE=none -DOGRE_CONFIG_DOUBLE=ON -DOGRE_ASSERT_MODE=1 -DOGRE_PROFILING=ON .
-        cmake --build build -- -j 4
\ No newline at end of file
+        cmake --build build -- -j 4
+    - name: Build with wayland
```
paroj, line 118: just add `-DOGRE_GLSUPPORT_USE_WAYLAND=ON` to the block above instead of introducing a new build just for this

joakimono: Done.

paroj:
tested this locally on Ubuntu 22.04 and added some final fixes. I would merge this next, if it is done from your side.

Out of curiosity as I did not test: does Wayland on Qt work in the current state?

joakimono:
> Out of curiosity as I did not test: does Wayland on Qt work in the current state?

No it does not. I suspect there is something missing regarding handling of the wl_surface but I do not know what. Maybe somehow initalise xdg_surface or indicate that it is a parent window. But these are guesses. I don't have time to look more into this for the time being.

paroj:
> No it does not. 

I just tested it on Ubuntu 22.04 & Qt 6.2.4 and I get an undecorated window with the wrong resolution. However input is already working. So likely not much is missing.

PR #3144, opened 2024-06-23
query SDL for video modes

paroj:


paroj:
this is currently incomplete, as we do not refresh resolutions if device changes like D3D11 does:
https://github.com/OGRECave/ogre/blob/21b46f2bd33228bc84d56545cf2ace148ed3284c/RenderSystems/Direct3D11/src/OgreD3D11RenderSystem.cpp#L513

PR #3154, opened 2024-07-08
Improve Wayland initialization

benjaminvdh:
There are some problems with Wayland initialization:
* An assertion fails, and the list of EGL extensions contains all client extensions twice and none of the display extensions, because WaylandEGLSupport::start() is called when mGLDisplay has not yet been set.
* GLNativeSupport::mFSAAlevels is duplicated every time a WaylandEGLWindow is created, because of the WaylandEGLSupport::doInit() call in WaylandEGLWindow::create().
* The wl_surface is stored in WaylandEGLSupport, but each WaylandEGLWindow should be able to have its own wl_surface.

I changed WaylandEGLSupport::start() to only print a message, and delayed the EGL initialization to the first time WaylandEGLSupport::newWindow() is called. This is similar to the way Win32GLSupport handles initialization.

I also fixed some minor errors I found when setting ENABLE_EGL_CHECK to 1 for testing purposes: eglTerminate could be called with a null mGLDisplay if more than one render system was loaded, causing an EGL_BAD_DISPLAY error, and trying to create an unsupported EGL context triggered an exception.

### DIFF
--- a/RenderSystems/GLSupport/include/EGL/Wayland/OgreWaylandEGLSupport.h
+++ b/RenderSystems/GLSupport/include/EGL/Wayland/OgreWaylandEGLSupport.h
@@ -22,21 +22,20 @@ public:
 
 protected:
     VideoModes mCbVideoModes;
+    EGLWindow* mInitialWindow = nullptr;
+
+    void start() override;
+
+    void doInit();
 
 public:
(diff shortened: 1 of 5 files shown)


paroj:
thanks

PR #3160, opened 2024-07-16
Either increment the iterator or assign when erasing

FrankGoyens:
Erasing the element invalidates the iterator so this is not possible in a simple range-based for loop causing a crash.

I encountered this when configuring a setup with multiple renderWindows, the program crashed during shutdown of the ogre root. We used `renderWindow->destroy()` to clean up the window, now we use `Ogre::Root::getSingleton().destroyRenderTarget(renderWindow)` instead because this circumvents the `erase` call in the loop. But I figured I would submit a pull request to fix the for loop anyway.


paroj:
rather reverting the commit that introduced the issue #3161

PR #3168, opened 2024-07-19
RenderSysten_GL: Prevent using a shared FBO renderbuffer if necessary

### LINKED ISSUE #3158: Two separate multisampled FBO's get the same renderbuffer due to caching, both FBO's contents are combined as a result
#### System Information
- Ogre Version: 14.3 (master)
- Operating System / Platform: Windows
- RenderSystem: RenderSystem_GL*

#### Detailled description
Hello, I encountered a problem when using multisampled FBO's.

I have two textures of the same size onto which I render different content using a "render to texture" FBO. 
The problem is that the contents of the first texture and the contents of the second texture are unintentionally combined, because they internally use the same OpenGL renderbuffer. They use the same buffer because the `GLFBOManager::requestRenderBuffer` uses a caching mechanism based on: format, width, height and nr of samples, which is the same for both textures. 

This is likely only a problem when "clearEveryFrame" is `false`. My rendering setup has a mechanism that requires "clearEveryFrame" to be false for both rendertargets, if "clearEveryFrame" is `true` then the contents wouldn't be combined as the unintentionally shared buffer gets cleared beforehand.

At my applications' side, I use `Ogre::TextureManager::getSingleton().createManual(GLNameGenerator::Generate(), Ogre::ResourceGroupManager::DEFAULT_RESOURCE_GROUP_NAME, Ogre::TEX_TYPE_2D,
        width, height, 0, Ogre::PF_R8G8B8A8, Ogre::TU_RENDERTARGET, 0, false, 8);` to create a texture with a renderTarget. 
And within the OGRE library it isn't until the `GLFrameBufferObject::initialise`, where on line 147 `mMultisampleColourBuffer = mManager->requestRenderBuffer(format, width, height, mNumSamples);` is called. 

So I wonder how this problem could be solved? I would try to solve this myself, but I think there's multiple approaches, so I could use some input.

FrankGoyens:
A shared renderbuffer might contain data from other FBO's, this is fine as long as all viewports have clearEveryFrame=true. When any viewport has clearEveryFrame=false, data would get unintentionally combined.

This fixes #3158

### DIFF
--- a/RenderSystems/GL/include/OgreGLFBORenderTexture.h
+++ b/RenderSystems/GL/include/OgreGLFBORenderTexture.h
@@ -89,6 +89,10 @@ namespace Ogre {
         /** Request a render buffer. If format is GL_NONE, return a zero buffer.
         */
         GLSurfaceDesc requestRenderBuffer(GLenum format, uint32 width, uint32 height, uint fsaa);
+
+        /** Creates a new render buffer. Caller takes ownership.
+        */
+        GLSurfaceDesc createNewRenderBuffer(GLenum format, uint32 width, uint32 height, uint fsaa);
         /** Get a FBO without depth/stencil for temporary use, like blitting between textures.
         */
         GLuint getTemporaryFBO() { return mTempFBO; }
--- a/RenderSystems/GL/include/OgreGLFBORenderTexture.h
+++ b/RenderSystems/GL/include/OgreGLFBORenderTexture.h
@@ -85,14 +85,9 @@ namespace Ogre {
 
         GLFBORenderTexture *createRenderTexture(const String &name,
             const GLSurfaceDesc &target, bool writeGamma, uint fsaa) override;
-        
-        /** Request a render buffer. If format is GL_NONE, return a zero buffer.
-        */
-        GLSurfaceDesc requestRenderBuffer(GLenum format, uint32 width, uint32 height, uint fsaa);
 
-        /** Creates a new render buffer. Caller takes ownership.
-        */
-        GLSurfaceDesc createNewRenderBuffer(GLenum format, uint32 width, uint32 height, uint fsaa);
+        GLSurfaceDesc createNewRenderBuffer(unsigned format, uint32 width, uint32 height, uint fsaa) override;
+
         /** Get a FBO without depth/stencil for temporary use, like blitting between textures.
         */
         GLuint getTemporaryFBO() { return mTempFBO; }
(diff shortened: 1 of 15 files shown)


FrankGoyens:
This approach switches between a cached and an owned (uncached) renderbuffer.
So far only implemented for RenderSystem_GL which is easier if some changes have to be made, since GL3Plus and GLES2 are very similar in this regard. So GL3Plus and GLES2 would be added later in this Pull Request if GL implementation is OK.

RenderSystems/GL/include/OgreGLFrameBufferObject.h
```diff
@@ -59,6 +59,10 @@ namespace Ogre {
     private:
         GLFBOManager *mManager;
         GLSurfaceDesc mMultisampleColourBuffer;
+        // mMultisampleColourBuffer.buffer is either shared through caching, or owned,
+        // if owned, mOwnedMultisampleColourBuffer contains mMultisampleColourBuffer.buffer
+        // otherwise, mOwnedMultisampleColourBuffer == nullptr
+        std::unique_ptr<GLHardwarePixelBufferCommon> mOwnedMultisampleColourBuffer;
```
paroj, line 65: move this definition to the `GLRenderTexture` parent

RenderSystems/GL/include/OgreGLFBORenderTexture.h
```diff
@@ -89,6 +89,10 @@ namespace Ogre {
         /** Request a render buffer. If format is GL_NONE, return a zero buffer.
         */
         GLSurfaceDesc requestRenderBuffer(GLenum format, uint32 width, uint32 height, uint fsaa);
+
+        /** Creates a new render buffer. Caller takes ownership.
+        */
+        GLSurfaceDesc createNewRenderBuffer(GLenum format, uint32 width, uint32 height, uint fsaa);
```
paroj, line 95: make this virtual and move the requestRenderBuffer impl to `GLRTTManager`

paroj:
looks good to me. However, I would directly implement this in the common classes, instead of adding GL3+ & GLES2 later

RenderSystems/GL/include/OgreGLFrameBufferObject.h
FrankGoyens: To the base class `GLFrameBufferObjectCommon` (defined in GLRenderTexture.h/.cpp)?

paroj: yes. sorry, mixed them up

FrankGoyens:
Thanks for the help, have a nice weekend

paroj:
had to revert this due to #3452. The new way is to call setDepthBufferPool with different pool ids on the rendertargets

PR #3184, opened 2024-08-06
fix crashes when reusing a Dx11 program pointer.  

creadmefford:
Fix crash due to accessing freed memory in Dx11 HLSL program class.

Min repro case:
1. compile a D3D11HLSLProgram based shader via prepare().
2. trigger an unload while state=LOADSTATE_PREPARED.  (see Resource::unload () snip below)
3. this calls unprepareImpl() which *only* frees the cached string data and NOT the other 8 or so std::vectors.
4. call prepare() again. the program will crash due while iterating the std::vector data containing freed strings.

Resource::unload(void) calls either unprepareImpl() or unloadImpl () depending on the 'old' load state of a shader program.
 if (old==LOADSTATE_PREPARED) {
         unprepareImpl();
 else
         unloadImpl();

### DIFF
--- a/RenderSystems/Direct3D11/src/OgreD3D11HLSLProgram.cpp
+++ b/RenderSystems/Direct3D11/src/OgreD3D11HLSLProgram.cpp
@@ -79,6 +79,16 @@ namespace Ogre {
         if(!mCompileError)
             analizeMicrocode();
     }
+
+    inline char* copyString(const char* src)
+    {
+        auto len = strlen(src);
+        char* dst = new char[len+1];
+        memcpy(dst, src, len);
+        dst[len] = '\0';
+        return dst;
+    }
+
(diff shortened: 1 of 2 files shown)


RenderSystems/Direct3D11/src/OgreD3D11HLSLProgram.cpp
```diff
@@ -79,6 +79,16 @@ namespace Ogre {
         if(!mCompileError)
             analizeMicrocode();
     }
+
+    inline char* copyString(const char* src)
+    {
+        auto len = strlen(src);
+        char* dst = new char[len+1];
+        strcpy(dst, src);
```
paroj, line 87: you can use memcpy here, because you already did strlen above.

PR #3194, opened 2024-08-19
switching to static builds.  INSTALL project can't find any of the PDBs.

creadmefford:
(this is NOT intended for merge/checkin)

QUESTION 1
several previously protected OgreSceneManager members were made private breaking my custom SceneManager derived class.  Any guidance on how I should approach moving forward?  

(EDIT: removed QUESTION 2 since i've learned enough to avoid the static route and will continue with dynamic linking)

thanks!

### DIFF
--- a/OgreMain/include/OgreSceneManager.h
+++ b/OgreMain/include/OgreSceneManager.h
@@ -699,10 +699,14 @@ namespace Ogre {
         void firePreUpdateSceneGraph(Camera* camera);
         /// Internal method for firing post update scene graph event
         void firePostUpdateSceneGraph(Camera* camera);
+
+        protected:
         /// Internal method for firing find visible objects event
         void firePreFindVisibleObjects(Viewport* v);
         /// Internal method for firing find visible objects event
         void firePostFindVisibleObjects(Viewport* v);
+
+        private:
         /// Internal method for firing destruction event
         void fireSceneManagerDestroyed();
 
@@ -742,8 +746,12 @@ namespace Ogre {
         /** Flag that indicates if all of the scene node's bounding boxes should be shown as a wireframe. */
         bool mShowBoundingBoxes;
 
+        protected:
+
         /// Utility class for calculating automatic parameters for gpu programs
         std::unique_ptr<AutoParamDataSource> mAutoParamDataSource;
+
+        private:
 
         GpuProgramParametersPtr mFixedFunctionParams;
 
@@ -760,6 +768,8 @@ namespace Ogre {
         void updateDirtyInstanceManagers(void);
 
         void _destroySceneNode(SceneNodeList::iterator it);
+
+        protected:
 
         struct _OgreExport ShadowRenderer
         {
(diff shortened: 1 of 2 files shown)


OgreMain/include/OgreSceneManager.h
```diff
@@ -742,9 +746,13 @@ namespace Ogre {
         /** Flag that indicates if all of the scene node's bounding boxes should be shown as a wireframe. */
         bool mShowBoundingBoxes;
 
+        protected:
+
         /// Utility class for calculating automatic parameters for gpu programs
         std::unique_ptr<AutoParamDataSource> mAutoParamDataSource;
```
paroj, line 752: there is `_getAutoParamDataSource()` to access this one now

creadmefford: From a design perspective, i'd like to use Ogre's code wherever I can AND yet have the flexibility to make my own customizations as needed.  My scene manager is essentially a refactor of SceneManager::_renderScene () so I need access to a similar set of members as that official function.

I'm not a OO proponent but it seems to me that either SceneManager needs to be an abstract interface (that I can implement however I like) *OR* SceneManager needs to give me access to it's members so I can create my own version in whatever way I need.  I could breakaway and make my own completely new SceneManager but that's a huge invasive change (i.e. Ogre objects all over store/accept SceneManager*) and Ogre's code does so much that I need anyway so there's no point in rewriting it all myself.

I can remove calls to firePreFindVisibleObjects, firePostFindVisibleObjects, and mShadowRenderer for now.  if i decide to refactor / change my shadow rendering I would need access to the mShadowRenderer though.
It looks like _getAutoParamDataSource () is const only and I need write access.

Thanks so much!

paroj: > From a design perspective, i'd like to use Ogre's code wherever I can AND yet have the flexibility to make my own customizations as needed. 

Understood. The problem is that Ogre never had a proper extension API for SceneManager. In fact Ogre never properly made anything `private`.
However, in order to be able to give API guarantees *and* get anything done, I need to lock down some parts. In lack of any other implementations, I just look at the SM plugins, which in the case of `_renderScene` is just this:

https://github.com/OGRECave/ogre/blob/8cd501b66c2658af864ba1fb1c31dd7944523b3c/PlugIns/PCZSceneManager/src/OgrePCZSceneManager.cpp#L449-L459

if I had thought more about it, probably I could have made `_renderScene` non-virtual and move the above code into `preUpdateSceneGraph` of a Listener.

Thats why I have been asking about example code. If you merely swap two lines in your `_renderScene`, maybe we can come up with an API to do just that without exposing all the other internals.

Particularly `mShadowRenderer` is not meant to be stable at all. In fact I want to replace it by the Compositor at some point.

creadmefford: ok this all makes sense.  for now, i'll just remove/minimize my private usage.  once you're stable with 14 i can revisit and work within those bounds.  then if there's something i still can't do i'll reach out again.  thanks!

creadmefford: i really like your approach of a significantly simpler scene manager.  maybe one that just creates scene nodes, manages visibility, and what not.  seems like you're saying PCZ SM could just be a listener.  very cool.  no need to have shadows in there, skyboxes, etc.  looking forward to the future!

paroj:
closing for now

PR #3216, opened 2024-09-07
fix crash in the ControllerManager::updateAllControllers when mControllers is changed while iterating over it

Altren:
More details: https://forums.ogre3d.org/viewtopic.php?t=97339

### DIFF
--- a/OgreMain/src/OgreControllerManager.cpp
+++ b/OgreMain/src/OgreControllerManager.cpp
@@ -77,7 +77,8 @@ namespace Ogre {
         unsigned long thisFrameNumber = Root::getSingleton().getNextFrameNumber();
         if (thisFrameNumber != mLastFrameNumber)
         {
-            for (auto *ci : mControllers)
+            auto controllersCopy = mControllers;
+            for (auto *ci : controllersCopy)
             {
                 ci->update();
             }


paroj:
can you elaborate on the use-case where you need this?

Also, we typically use this pattern to allow remove/ add during iteration:
https://github.com/OGRECave/ogre/blob/d111e9d5e5b4820c4adc01ffd9c3db1c6b451afd/OgreMain/src/OgreSceneManager.cpp#L2418-L2419

OgreMain/src/OgreControllerManager.cpp
```diff
@@ -77,9 +77,10 @@ namespace Ogre {
         unsigned long thisFrameNumber = Root::getSingleton().getNextFrameNumber();
         if (thisFrameNumber != mLastFrameNumber)
         {
-            for (auto *ci : mControllers)
+            // intentionally do a loop using indices, because mControllers might change within the loop
+            for (size_t i = 0; i < mControllers.size(); ++i)
             {
-                ci->update();
+                mControllers[i]->update();
```
paroj, line 83: this will skip a controller i+1, if controller i removes itself.

Altren:
I do not know why this crash started happening and how to reproduce it on a minimal example

paroj:
crashing was likely introduced by f7cd9dd73c3894cb42ab5e33a002262752b018a8 hence this is fixing a regression

PR #3217, opened 2024-09-08
add support for structed bindings for Ogre::Vector

ribbon-otter:
It would be nice to be able to write 
```
auto [x,y,z] = myVector;
```
to conveniently pull `myVector` apart into its components for some algorithms. This pull request makes turns `Ogre::Vector` into a tuple-like (as defined in the C++ std library), so that structured binding is supported.  Given that `Ogre::Vector` is a fixed length container, it is in many ways an specialized alternative to `std::array`, and `std::array` is also a tuple-like.

I invite feedback, particularly since `Ogre::Vector` is such a fundamental part of the Ogre project, and I am venturing into unfamiliar parts of the C++ language with this pull request.

### DIFF
--- a/OgreMain/include/OgreVector.h
+++ b/OgreMain/include/OgreVector.h
@@ -66,7 +66,7 @@ namespace Ogre
         const T* ptr() const { return data; }
         /** to help make VectorBase enum-like */
         template<std::size_t N> T get() {
-            static_assert(N < dims);
+            static_assert(N < dims, "invalid index");
             return data[N];
         }
     };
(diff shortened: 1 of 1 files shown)


ribbon-otter:
looks like I forgot (and didn't find) how to run the tests that include trailing whitespace locally.

paroj:
```cpp
auto& [x,y,z] = myVector;
x = 1;
```
this should modify myVector.x, but currently will not work, right?

paroj:
I did some tests:

for types that go through the VectorBase specialization things just work without this PR:
```cpp
    Vector3 v;
    auto& [a,b,c] = v;
    a = 1;
```

the only problem are types that use the generic path:
```cpp
    Vector3i v;
    auto& [a,b,c] = v.data;
    a = 1;
```
they need `.data` added

paroj:
closing for now

PR #3236, opened 2024-09-28
Fix get current surface for Android if use SDL

HadesD:
On android, i'm using SDL for create window
But without this, app will be crash, because SDL already created surface, so just get the surface from current

### DIFF
--- a/RenderSystems/GLSupport/src/EGL/Android/OgreAndroidEGLWindow.cpp
+++ b/RenderSystems/GLSupport/src/EGL/Android/OgreAndroidEGLWindow.cpp
@@ -119,9 +119,12 @@ namespace Ogre {
                                 "currentGLContext was specified with no current GL context",
                                 "EGLWindow::create");
                 }
-#if defined(OGRE_BITES_HAVE_SDL) && OGRE_BITES_HAVE_SDL
-                mEglSurface = eglGetCurrentSurface(EGL_DRAW);
-#endif
+
+                if((opt = miscParams->find("currentEGLSurface")) != end)
+                {
+                    mEglSurface = eglGetCurrentSurface(EGL_DRAW);
+                }
+
                 mEglDisplay = eglGetCurrentDisplay();
             }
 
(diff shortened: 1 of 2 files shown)


RenderSystems/GLSupport/src/EGL/Android/OgreAndroidEGLWindow.cpp
```diff
@@ -119,7 +119,9 @@ namespace Ogre {
                                 "currentGLContext was specified with no current GL context",
                                 "EGLWindow::create");
                 }
-
+#if defined(OGRE_BITES_HAVE_SDL) && OGRE_BITES_HAVE_SDL
```
paroj, line 122: do not use defines here, but rather pass `currentGLSurface=true` via miscParams. See the `currentGLContext` code just above.
You can use #ifdef in OgreBites to conditionally pass this flag.

HadesD: @paroj 
where should I put `if (currentGLSurface=true)`?
is inside the `if (currentGLContext = true)` above accepable?

RenderSystems/GLSupport/src/EGL/Android/OgreAndroidEGLWindow.cpp
```diff
@@ -120,6 +120,11 @@ namespace Ogre {
                                 "EGLWindow::create");
                 }
 
+                if((opt = miscParams->find("currentEGLSurface")) != end)
```
paroj, line 123: also add the parseBool part, as currently `currentEGLSurface=false` would be accepted

HadesD: @paroj 
Thanks,
I've added

paroj:
one final request:

please add the new property to the property table:
https://github.com/OGRECave/ogre/blob/185c369e8f2bcdd98ade5e7649c4d11bd6db258b/OgreMain/include/OgreRenderSystem.h#L395

PR #3237, opened 2024-09-29
Prevent crash on Animation

HadesD:


paroj:
can you provide some details where this crash happens? inside _getTimeIndex?

Also this is only triggered, if the animation is active without having any keyframes, right?

HadesD:
I'm not check so deep in to that crash, but i'm using old skel and maybe it is issue that make this crash happened
anyway, it crash at this line
```
auto it = std::lower_bound(mKeyFrameTimes.begin(), mKeyFrameTimes.end() - 1, timePos);
```
so I think in every case, we should check `mKeyFrameTimes` is `empty` or not.

paroj:
I could not reproduce the crash with no keyframes on libstdc++ on linux. But maybe libc++ on android behaves differently.

Could you attach the skeleton for me to verify?

HadesD:
sorry for the project, that can not share skeleton
But I think you can check UT if `mKeyFrameTimes.empty() == true` then I think it will be exception

PR #3238, opened 2024-10-04
Allows multi-threading when rendering.

andre-caldas:
SWIG is capable of generating multi-threading code.

We allow locking the GIL when python code is called from C++ through some virtual method. However, we do not automatically allow the GIL to be released when C++ code is called from python, for that would happen in abut 9000 places and the performance penalty would not be negligible.

This is done with `thread="1"` and `%feature("nothreadallow");`.

However, for some selected methods we do release the GIL, allowing multi-threading. This is done with `%threadsallow` only for the methods `Root::startRendering()`, `Root::renderOneFrame()` and `Root::_updateAllRenderTargets`.

### DIFF
--- a/Components/Bites/include/OgreBites.i
+++ b/Components/Bites/include/OgreBites.i
@@ -1,4 +1,4 @@
-%module(package="Ogre", directors="1") Bites
+%module(package="Ogre", directors="1", threads="1") Bites
 %{
 /* Includes the header in the wrapper code */
 #include "Ogre.h"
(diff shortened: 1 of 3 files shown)


andre-caldas:
Linux build is failing because of this:
https://github.com/swig/swig/issues/1996

paroj:
python 3.14 now allows a free-thread execution, that is also supported by swig:
https://docs.python.org/3/whatsnew/3.14.html#whatsnew314-free-threaded-cpython

which is superior to this approach

Issue #3248, opened 2024-10-29
Build fails with recent Vulkan/glslang shared library build


Apteryks:
#### System Information
- Ogre Version: 14.3.1
- Operating System / Platform: GNU/Linux
- RenderSystem: Wayland

#### Detailled description
Attempting to build Ogre 14.3.1 with `-DOGRE_BUILD_RENDERSYSTEM_VULKAN=ON` and shared system libraries of glslang, I get:

```
[ 43%] Building CXX object PlugIns/BSPSceneManager/CMakeFiles/Plugin_BSPSceneManager.dir/src/OgreBspSceneManagerDll.cpp.o
cd /tmp/guix-build-ogre-14.3.1.drv-0/build/PlugIns/BSPSceneManager && /gnu/store/86fc8bi3mciljxz7c79jx8zr4wsx7xw8-gcc-11.4.0/bin/c++ -DPlugin_BSPSceneManager_EXPORTS -I/tmp/guix-build-ogre-14.3.1.drv-0/source/PlugIns/BSPSceneManager/include -I/tmp/guix-build-ogre-14.3.1.drv-0/source/OgreMain/include -I/tmp/guix-build-ogre-14.3.1.drv-0/build/include -I/tmp/guix-build-ogre-14.3.1.drv-0/source/OgreMain/src/zip -Wall -Winit-self -Wcast-qual -Wwrite-strings -Wextra -Wundef -Wmissing-declarations -Wno-unused-parameter -Wshadow -Wno-missing-field-initializers -Wno-long-long -Wsuggest-override  -msse -O2 -g -DNDEBUG -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -std=c++14 -MD -MT PlugIns/BSPSceneManager/CMakeFiles/Plugin_BSPSceneManager.dir/src/OgreBspSceneManagerDll.cpp.o -MF CMakeFiles/Plugin_BSPSceneManager.dir/src/OgreBspSceneManagerDll.cpp.o.d -o CMakeFiles/Plugin_BSPSceneManager.dir/src/OgreBspSceneManagerDll.cpp.o -c /tmp/guix-build-ogre-14.3.1.drv-0/source/PlugIns/BSPSceneManager/src/OgreBspSceneManagerDll.cpp
[ 43%] Linking CXX shared library ../../lib/Plugin_GLSLangProgramManager.so
[...]

PR #3249, opened 2024-10-29
build/glslang: Fix build when OSDependent library is missing.

Apteryks:
The OSDependent glslang library is only available when it is built as a static library, not a shared one; look for the library but do not fail the  build when it doesn't exist.

Fixes: <https://github.com/OGRECave/ogre/issues/3248>


PlugIns/GLSLang/CMakeLists.txt
```diff
@@ -19,7 +19,11 @@ elseif(DEFINED ENV{VULKAN_SDK})
     target_link_directories(Plugin_GLSLangProgramManager PUBLIC $ENV{VULKAN_SDK}/lib)
     target_link_libraries(Plugin_GLSLangProgramManager PUBLIC OgreMain shaderc_combined)
 else()
-  set(GLSLANG_LIBS glslang OSDependent SPIRV SPIRV-Tools-opt SPIRV-Tools)
+  set(GLSLANG_LIBS glslang SPIRV SPIRV-Tools-opt SPIRV-Tools)
+  find_library(GLSLANG_OSDependent OSDependent)
+  if(GLSLANG_OGLCompiler)
```
paroj, line 24: this condition is wrong

Apteryks: Oops! Should be fixed now.

PR #3269, opened 2025-01-12
ApplicationContextBase: changing raw pointer to smart ptr

Joilnen:


paroj:
as the member is merely protected, this may break compilation of users. Therefore scheduling for Ogre 15 release

PR #3271, opened 2025-01-12
Tests: range loop refactoring and change raw pointer to smart pointer

Joilnen:


Tests/OgreMain/src/RootWithoutRenderSystemFixture.cpp
```diff
@@ -25,23 +25,20 @@ void RootWithoutRenderSystemFixture::SetUp()
...
         }
     }
 }
 
 void RootWithoutRenderSystemFixture::TearDown()
 {
-    delete mRoot;
-    delete mHBM;
-    delete mFSLayer;
+    // enforce this is removed by the last
```
paroj, line 42: misleading: it is actually deleted first

Tests/OgreMain/include/RootWithoutRenderSystemFixture.h
```diff
@@ -36,9 +36,9 @@ THE SOFTWARE.
 
 class RootWithoutRenderSystemFixture : public ::testing::Test {
 public:
-    Ogre::Root* mRoot;
-    Ogre::HardwareBufferManager* mHBM;
-    Ogre::FileSystemLayer* mFSLayer;
+    std::shared_ptr<Ogre::Root> mRoot;
+    std::unique_ptr<Ogre::HardwareBufferManager> mHBM;
```
paroj, line 40: you can fix the error by instead moving mHBM on top of mRoot

Joilnen: fixed

PR #3273, opened 2025-01-15
D3D11: _create2DTex() with mSurface.

Arthurfernades:
Created a variable called mSurface that can be used to create a texture from an already existing surface. Inside the _create2DTex() function, there is an if statement to check whether this variable exists or not. If it exists, it will be manipulated to be used as an Ogre texture.

### DIFF
--- a/Docs/src/high-level-programs.md
+++ b/Docs/src/high-level-programs.md
@@ -331,6 +331,9 @@ Set the optimisation level, which can be one of ’default’, ’none’, ’0�
 
 # Assembler Shaders {#Assembler-Shaders}
 
+You can specify assembly shaders directly as the syntax code of the program, if you compiled them ahead of time.
+Alternatively, you can use the syntax code to specify the @c target of a high-level shader, which will be compiled on-the-fly to the specified syntax.
+
 The current supported syntaxes are:
 
 <dl compact="compact">


RenderSystems/Direct3D11/src/OgreD3D11Texture.cpp
```diff
@@ -26,252 +26,297 @@ THE SOFTWARE.
 -----------------------------------------------------------------------------
```
paroj, line 26: this file seems to have many unrelated changes, please check

OgreMain/include/OgreTexture.h
```diff
@@ -146,6 +146,14 @@ namespace Ogre {
         */
         TextureType getTextureType(void) const { return mTextureType; }
 
+        /** Sets the surface when you already have one
+         */
+        virtual void _setSurface(void* surface) { mSurface = surface; }
+
+        /** Gets the surface
+         */
+        void* getSurface(void) const { return mSurface; }
```
paroj, line 155: not needed

OgreMain/include/OgreTexture.h
```diff
@@ -146,6 +146,14 @@ namespace Ogre {
         */
         TextureType getTextureType(void) const { return mTextureType; }
 
+        /** Sets the surface when you already have one
+         */
+        virtual void _setSurface(void* surface) { mSurface = surface; }
```
paroj, line 151: make this an empty implementation

OgreMain/include/OgreTexture.h
```diff
@@ -473,6 +483,8 @@ namespace Ogre {
 
         TextureType mTextureType;
 
+        void* mSurface;
```
paroj, line 486: move this to D3D11 as it is the only user

RenderSystems/Direct3D11/src/OgreD3D11Texture.cpp
Arthurfernades: There is something weird here because everything is correct in my project; it's just a basic change in the _create2DTex() method.

paroj: maybe you had auto indentation enabled or something. You could try copying over the original file and then adding your changes again

Components/Overlay/include/ImGui.i
```diff
@@ -49,7 +49,7 @@
 #ifdef SWIGPYTHON
 // match the signature of the by value variants
 %typemap(argout) float[4], float[3], float[2] {
-    $result = SWIG_AppendOutput($result, SWIG_NewPointerObj($1, $descriptor(ImVec4*), 0));
+    $result = SWIG_Python_AppendOutput($result, SWIG_NewPointerObj($1, $descriptor(ImVec4*), 0));
```
paroj, line 52: this changes causes the windows CI to fail

RenderSystems/Direct3D11/src/OgreD3D11Texture.cpp
```diff
@@ -232,46 +232,106 @@ namespace Ogre
     inline bool IsPowerOfTwo(unsigned int n)                { return ((n&(n-1))==0);                    }
     //---------------------------------------------------------------------
     void D3D11Texture::_create2DTex()
+    {
+            if (NULL != mSurface)
```
paroj, line 236: instead of if else around everything I suggest you do:
```cpp
if(mSurface)
{
   _createShared2DTex();
  return;
}
```

RenderSystems/Direct3D11/src/OgreD3D11Texture.cpp
```diff
@@ -232,46 +232,106 @@ namespace Ogre
...
+    {
+        HRESULT hr = S_OK;
+
+        IUnknown* pUnk = (IUnknown*)mSurface;
+
+        IDXGIResource* pDXGIResource;
+        hr = pUnk->QueryInterface(__uuidof(IDXGIResource), (void**)&pDXGIResource);
+        if (FAILED(hr))
+        {
+            throw std::runtime_error("Failed to query IDXGIResource interface from the provided object.");
```
paroj, line 246: please use the `OGRE_EXCEPT_EX` macro instead of throwing directly

paroj:
accidently closed this while cleaning up. Continued in #3275

PR #3275, opened 2025-01-19
D3D11: _create2DTex() from shared surface.

paroj:
dropped Flush as you can get the device via getCustomAttribute("D3DDEVICE") and flush yourself


Arthurfernades:
@paroj Is it necessary to create another pull request because the CI Build/Linux failed?

paroj:
no, I can fix this later.

PR #3291, opened 2025-02-21
Force the DDS/DXT stream decompression.

CsabaStupak:
The DDSCodec keeps the DXT format if the hardware supports it. However using this stack object it is possible to enforce the DXT decompression into RGB(A) format. This is useful e.g. when we have DDS texture without mipmaps and we need them. DirectX11 does not generate mipmaps if the texture is compressed (DirectX9 does it automatically).

Also added support to obtain the DDS stream (not only saving to file).

### DIFF
--- a/OgreMain/src/OgreDDSCodec.cpp
+++ b/OgreMain/src/OgreDDSCodec.cpp
@@ -160,27 +160,6 @@ namespace {
 }
 
     //---------------------------------------------------------------------
-    // Force the DXT stream decompression
-    //---------------------------------------------------------------------
-    int DDSDecodeEnforcer::mCount = 0;
-
-    DDSDecodeEnforcer::DDSDecodeEnforcer()
-    {
-        ++mCount;
-    }
-
-    DDSDecodeEnforcer::~DDSDecodeEnforcer()
-    {
-        --mCount;
-        assert(mCount >= 0);
-    }
-
-    bool DDSDecodeEnforcer::isEnabled()
-    {
-        return mCount > 0;
-    }
-
-    //---------------------------------------------------------------------
     DDSCodec* DDSCodec::msInstance = 0;
     //---------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


OgreMain/src/OgreDDSCodec.h
```diff
@@ -44,6 +44,23 @@ namespace Ogre {
     struct DXTExplicitAlphaBlock;
     struct DXTInterpolatedAlphaBlock;
 
+    /** Force the DXT stream decompression.
+
+        The DDSCodec keeps the DXT format if the hardware supports it.
+        However this stack object can enforce the DXT decompression into RGB(A) format.
+    */
+    class _OgreExport DDSDecodeEnforcer
```
paroj, line 52: You are in a private header here which is not accessible for client code.

Please add a generic virtual `setParameter` method to Codec like this:
https://github.com/OGRECave/ogre/blob/57fc7dfbf5e45017923a8e8beb54edf6d8aa653d/Components/RTShaderSystem/include/OgreShaderSubRenderState.h#L161-L162

and override it in DDSCodec. You can create a stack object based on this in your application code.

CsabaStupak: Changed the code according to review.
Thanks for the review!

OgreMain/src/OgreDDSCodec.cpp
```diff
@@ -1043,6 +1066,17 @@ namespace {
         return BLANKSTRING;
 
     }
+    //---------------------------------------------------------------------
+    bool DDSCodec::setParameter(const String& name, const String& value)
+    {
+        if (name == "decode_enforce")
+        {
+            StringConverter::parse(value, mDecodeEnforce);
```
paroj, line 1074:
```suggestion
            return StringConverter::parse(value, mDecodeEnforce);
```

in case the value string contains some garbage

PR #3292, opened 2025-02-21
Update vulkansdk into Dockerfile

mosfet80:
https://vulkan.lunarg.com/sdk/home


paroj:
why is this necessary?

also, if this is necessary all references (PIP, Windows) should be updated to the same version.

paroj:
no reply

PR #3298, opened 2025-03-11
fix: SWIG binding TerrainSlotMap.

Arthurfernades:
SWIG created a generic type for TerrainSlotMap; this apparently fixes it to the right type.

### DIFF
--- a/Components/Terrain/include/OgreTerrain.i
+++ b/Components/Terrain/include/OgreTerrain.i
@@ -32,6 +32,7 @@
 %include "OgreTerrainQuadTreeNode.h"
 
 %template(LayerInstanceList) std::vector<Ogre::Terrain::LayerInstance>;
+%template(TerrainSlotMap) std::map<uint32_t, Ogre::TerrainGroup::TerrainSlot*>;
 %template(TerrainRayResult) std::pair<bool, Ogre::Vector3>;
 %ignore Ogre::Terrain::getBlendTextureCount;
 %ignore Ogre::Terrain::getBlendTextureName;
--- a/Components/Terrain/include/OgreTerrain.i
+++ b/Components/Terrain/include/OgreTerrain.i
@@ -32,6 +32,17 @@
 %include "OgreTerrainQuadTreeNode.h"
 
 %template(LayerInstanceList) std::vector<Ogre::Terrain::LayerInstance>;
+#ifdef SWIGPYTHON
+%{
+    // this is a workaround for the following map
+    namespace swig {
+    template<> struct traits<Ogre::TerrainGroup::TerrainSlot> {
+        typedef pointer_category category;
+        static const char* type_name() { return "Ogre::TerrainGroup::TerrainSlot"; }
+    };
+    }
+%}
+#endif
 %template(TerrainSlotMap) std::map<uint32_t, Ogre::TerrainGroup::TerrainSlot*>;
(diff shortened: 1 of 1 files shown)


paroj:
to fix the build, you need to add a workaround like here:
https://github.com/OGRECave/ogre/commit/5d8188845ed1ffa3877633b7c78bf22257d6f0b3

Arthurfernades:
> to fix the build, you need to add a workaround like here: [5d81888](https://github.com/OGRECave/ogre/commit/5d8188845ed1ffa3877633b7c78bf22257d6f0b3)

Can you show me exactly how to do it? I fixed other bindings but without workaround. I've tried in differente ways to implement workaround but everytime I get a different error.

Components/Terrain/include/OgreTerrain.i
```diff
@@ -32,6 +32,7 @@
 %include "OgreTerrainQuadTreeNode.h"
 
 %template(LayerInstanceList) std::vector<Ogre::Terrain::LayerInstance>;
+%template(TerrainSlotMap) std::map<uint32_t, Ogre::TerrainGroup::TerrainSlot*>;
```
paroj, line 35:
```suggestion
#ifdef SWIGPYTHON
%{
    // this is a workaround for the following map
    namespace swig {
    template<> struct traits<Ogre::TerrainGroup::TerrainSlot> {
        typedef pointer_category category;
        static const char* type_name() { return "Ogre::TerrainGroup::TerrainSlot"; }
    };
    }
%}
#endif
%template(TerrainSlotMap) std::map<uint32_t, Ogre::TerrainGroup::TerrainSlot*>;
```

paroj:
please rebase on master to fix the build

Arthurfernades:
Oops, that pull request was unintentional :(

Arthurfernades:
In fact, if Copilot is correct, the error is due to the TerrainSlotMap binding in Java. Is that right?

Components/Terrain/include/OgreTerrain.i
```diff
@@ -32,6 +32,18 @@
...
+    // this is a workaround for the following map
+    namespace swig {
+    template<> struct traits<Ogre::TerrainGroup::TerrainSlot> {
+        typedef pointer_category category;
+        static const char* type_name() { return "Ogre::TerrainGroup::TerrainSlot"; }
+    };
+    }
+%}
+#endif
+%template(TerrainSlotMap) std::map<uint32_t, Ogre::TerrainGroup::TerrainSlot*>;
```
paroj, line 46:
```suggestion
#ifndef SWIGJAVA
%template(TerrainSlotMap) std::map<uint32_t, Ogre::TerrainGroup::TerrainSlot*>;
#endif
```

Arthurfernades:
There we go. So, I made other bindings. Do I need the workaround and this #ifndef SWIGJAVA for everyone?

paroj:
not generally. Just let the CI test it and then add fixes as needed.

Arthurfernades:
Just one more question: do you think it's better to make each pull request separately, or can I just send all bindings at once?

paroj:
if you make separate PRs, I can merge the ones that dont fail straight ahead. However, it will be more work switching branches etc. if you have to fix things

Arthurfernades:
Okay, thanks a lot, Paroj!

PR #3315, opened 2025-04-07
Use constexpr for static Vector constants

scpeters:
This fixes the initialization-order-fiasco for static Vector constants while preserving API (though breaking ABI) through the following steps:

* Make `constexpr` the `VectorBase` and `Vector` constructors
* Instantiate global `constexpr` variables in a hidden namespace in .cpp file
* Change static class constants to references (ABI break) and assign to global `constexpr` variables

This matches the approach taken in #3313. Note that the `Vector(T)` constructor could not be made `constexpr`, so the `ZERO` and `SCALE` variables were constructed with repeated arguments instead of a single argument.

### DIFF
--- a/OgreMain/include/OgreVector.h
+++ b/OgreMain/include/OgreVector.h
@@ -46,17 +46,17 @@ namespace Ogre
     template <int dims, typename T> struct VectorBase
     {
         VectorBase() {}
-        VectorBase(T _x, T _y)
+        constexpr VectorBase(T _x, T _y)
         {
             static_assert(dims > 1, "must have at least 2 dimensions");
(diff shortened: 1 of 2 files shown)


paroj:
> Note that the Vector(T) constructor could not be made constexpr

why not? works with clang for me..

scpeters:
> > Note that the Vector(T) constructor could not be made constexpr
> 
> why not? works with clang for me..

in my initial testing adding `constexpr` to the constructor didn't cause a compilation error if the static variables are not modified, but I haven't found a way to make it work when trying to do both (see #3316)

PR #3330, opened 2025-04-30
Update OgreD3D11HardwarePixelBuffer.h

rpgplayerrobin:
Part 1 of 3 commits.
More about this here:
https://forums.ogre3d.org/viewtopic.php?t=97512


paroj:
you have to edit the files on your ogre fork on the same branch. This way everything ends up in this PR.
e.g. for the cpp go here:
https://github.com/rpgplayerrobin/ogre/edit/patch-2/RenderSystems/Direct3D11/src/OgreD3D11HardwarePixelBuffer.cpp

PR #3332, opened 2025-04-30
Update OgreD3D11Texture.cpp

rpgplayerrobin:
Part 3 of 3 commits.
More about this here:
https://forums.ogre3d.org/viewtopic.php?t=97512


RenderSystems/Direct3D11/src/OgreD3D11Texture.cpp
```diff
@@ -603,6 +603,7 @@ namespace Ogre
         case D3D11_SRV_DIMENSION_TEXTURECUBE:
             RTVDesc.ViewDimension = D3D11_RTV_DIMENSION_TEXTURE2DARRAY;
             RTVDesc.Texture2DArray.FirstArraySlice = buffer->getFace();
+            RTVDesc.Texture2DArray.MipSlice = buffer->getMipLevel();
```
paroj, line 606: I guess the mipslice paramteer is missing from all RTVs and not just 2darray?

rpgplayerrobin: Yeah I guess so. But I have not tried to do it for non-cube textures so I have not checked if it works for those.

PR #3359, opened 2025-05-29
Main: AnimationTrack - use integer comparison to avoid overflow

### LINKED ISSUE #3354: Animation track without keyframes crashes animation system
#### System Information
- Ogre Version: master branch
- Operating System / Platform: Ubuntu 22.04, gcc 11
- RenderSystem: GL ES 2.0

#### Detailled description
deleting all keyframes on animation track makes animation system crash with SIGSEGV. The crash is 50% times for me.
for some reason not under gdb. To debug run executale a few times to crash and produce core and debug using gdb -c core /path/to/executable.
```c++
track->removeAllKeyFrames();
```

paroj:
when AnimationTrack has no keyframes. The parent animation can still have keyFrameTimes.

fixes #3354

### DIFF
--- a/OgreMain/src/OgreAnimationTrack.cpp
+++ b/OgreMain/src/OgreAnimationTrack.cpp
@@ -192,12 +192,12 @@ namespace Ogre {
         // Pre-allocate memory
         mKeyFrameIndexMap.resize(keyFrameTimes.size());
 
-        size_t i = 0, j = 0;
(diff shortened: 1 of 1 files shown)


slapin:
Tested it - the problem is no longer reproducible, hurray!

slapin:
But I guess you meant not 'integer comparisons' but 'signed comparisons'

Issue #3358, opened 2025-05-29
OgreBullet does not support btCompoundShape


slapin:
... which is the only way to offset collision shapes and have multiple of them per RigidBody.
Additionally all the CollisionShape utility functions are hidden which makes it hard to create
shapes as needed from shapes and entities, so I think the following should be done:

- [ ] Add btCompoundShape support to OgreBullet.cpp
- [ ] make shape creation functions available as API.
- [ ] a API function is needed which would allow btRigidBody/btCollisionObject created outside to have Ogre listener attached for Ogre to handle the motion.

slapin:
Created one of 3 PRs.

slapin:
All PRs created. Then I go back to implementing move_and_slide...

slapin:
added "implement Godot's move_and_slide()" it would be easier for me to track.

slapin:
*achievement* made btPairCachingGhostShape work and still sane. Muhahahahagh!!!111

PR #3363, opened 2025-05-31
OgreBullet: Support btCompound shape; collider creation

slapin:
btCompoundShape allows to create more than 1 collider per body. Additionally it is the only way to offset colliders against body.

Additionally allow creation of Trimesh and ConvexHull colliders out of creating RigidBody. This will help filling-up btCompoundShape, for example.

Related to #3358

### DIFF
--- a/Components/Bullet/include/OgreBullet.h
+++ b/Components/Bullet/include/OgreBullet.h
@@ -31,7 +31,8 @@ enum ColliderType
     CT_CYLINDER,
     CT_CAPSULE,
     CT_TRIMESH,
-    CT_HULL
+    CT_HULL,
+    CT_COMPOUND
 };
 
 inline btQuaternion convert(const Quaternion& q) { return btQuaternion(q.x, q.y, q.z, q.w); }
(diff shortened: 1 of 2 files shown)


Components/Bullet/include/OgreBullet.h
```diff
@@ -72,6 +73,12 @@ _OgreBulletExport btBoxShape* createBoxCollider(const MovableObject* mo);
 _OgreBulletExport btCapsuleShape* createCapsuleCollider(const MovableObject* mo);
 /// create capsule collider using ogre provided data
 _OgreBulletExport btCylinderShape* createCylinderCollider(const MovableObject* mo);
+/// create triMesh collider
+_OgreBulletExport btBvhTriangleMeshShape* createTrimeshCollider(const MovableObject* mo);
```
paroj, line 77:
```suggestion
_OgreBulletExport btBvhTriangleMeshShape* createTrimeshCollider(const Entity* ent);
```
same for createConvexHullCollider.

They will not work with any MovableObject - no need to pretend.

Components/Bullet/src/OgreBullet.cpp
```diff
@@ -91,6 +91,12 @@ btCylinderShape* createCylinderCollider(const MovableObject* mo)
     return shape;
 }
 
+/// create compound shape because we can
+btCompoundShape* createCompoundShape()
+{
+	return new btCompoundShape;
```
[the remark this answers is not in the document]
slapin: Just for self-similarity sake i.e for easier intuition coding. Because other shape constructors look similar.

Components/Bullet/include/OgreBullet.h
slapin: Fixed.

PR #3367, opened 2025-06-04
Samples: range loop refactoring

Joilnen:


Joilnen:
I can't see any error in windows build

PR #3373, opened 2025-06-07
Main: range loop refactoring

Joilnen:


OgreMain/src/OgreEntity.cpp
```diff
@@ -2198,14 +2198,13 @@ namespace Ogre {
 #if !OGRE_NO_MESHLOD
         // if manual LOD is in use, visit those too
         ushort lodi = 1;
-        for (LODEntityList::iterator e = mLodEntityList.begin(); 
-            e != mLodEntityList.end(); ++e, ++lodi)
```
paroj, line 2202: this is not equivalent

Joilnen: please check it

PR #3392, opened 2025-06-23
Implement height field terrain collider for Ogre terrain

slapin:
Added simple implementation of integration of btHeightfieldTerrainShape with Ogre Terrain
data for easy building of colliders.


Components/Bullet/src/OgreBullet.cpp
```diff
@@ -97,6 +100,37 @@ btCompoundShape* createCompoundShape()
 	return new btCompoundShape;
 }
 
+#ifdef OGRE_BUILD_COMPONENT_TERRAIN
+/// create height field collider
+btHeightfieldTerrainShape* createHeightfieldTerrainShape(const Terrain* terrain, struct HeightFieldData *data)
+{
+    int i;
```
paroj, line 107: if OGRE_BUILD_COMPONENT_TERRAIN is not available, you can add a
```
OgreAssert(false, "recompile with Terrain component");
```
here

Components/Bullet/include/OgreBullet.h
```diff
@@ -81,6 +85,19 @@ _OgreBulletExport btBvhTriangleMeshShape* createTrimeshCollider(const Entity* en
...
 _OgreBulletExport btCompoundShape* createCompoundShape();
+#ifdef OGRE_BUILD_COMPONENT_TERRAIN
+/// height field data
+struct HeightFieldData {
+    /** the position for a center of the shape, i.e. where to place btRigidBody
+     *  or a child of btCompoundShape */
+    Vector3 bodyPosition;
+    /** a heightfield pointer to be freed when
+     * btHeightfieldTerrainShape is freed */
+    float *terrainHeights;
```
paroj, line 96: you could use std::unique_ptr to enforce deletion

Components/Bullet/include/OgreBullet.h
```diff
@@ -12,9 +12,13 @@
 #include "Ogre.h"
 #include "btBulletCollisionCommon.h"
 #include "btBulletDynamicsCommon.h"
+#include "BulletCollision/CollisionShapes/btHeightfieldTerrainShape.h"
 
 namespace Ogre
 {
+#ifdef OGRE_BUILD_COMPONENT_TERRAIN
```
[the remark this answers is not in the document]
slapin: Fixed.

Components/Bullet/include/OgreBullet.h
slapin: The problem is I found no way to automagically delete colliders without user-side code help. Would be nice to have some event for that. For now I just create node and store collider in there, however it is not user accessible currently... Need to solve architectural issue here... Would be nice to have some terrain metadata or something to keep shared_ptr there...

slapin:
Thanks a lot!

PR #3393, opened 2025-06-25
OgreBullet: terrain: ensure colliders get destroyed with terrain.

slapin:
Attach terrain collider's SceneNode to internal terrain's SceneNode.


paroj:
I did not closely look at this part initially, but each terrain already has a SceneNode associated with it here:
https://github.com/OGRECave/ogre/blob/0dd66045b5223a6726594c7571dbf730218b1a5c/Components/Terrain/src/OgreTerrain.cpp#L194

probably you can use that instead

slapin:
Updated and it looks like it works, Looked at code and did not find much use for that SceneNode so I hope the way I used it won't break anything for anyone.

PR #3398, opened 2025-06-28
Overlay: update imgui to 1.92

paroj:


tomconder:
Build fails due to a trailing whitespace
```
Components/Overlay/src/OgreImGuiOverlay.cpp:240: trailing whitespace.
+        updateTextureData(*draw_data->Textures); 
```
Also imgui v1.92.1 is now available. You may want to check this PR against it.

PR #3402, opened 2025-06-29
Delete Other/add_crlf.py

mosfet80:
the functions contained in the file are triplicated. files Other/tab2spc.py and Other/rem_endspc.py contain the same functions


paroj:
> contain the same functions

not true

Issue #3387, opened 2025-06-19
OgreBullet: debug drawing of btHeightfieldShape is extremely slow


slapin:
Hi, all!.
When using Ogre terrain and adding collider to it using btHeightfieldShape, if debug drawing enabled, on my laptop the FPS drop is from 3000 FPS unlocked to 3FPS. Which is quite strange as this never happens with Bullet demo. Otherwise (if debug drawing is disabled) the FPS drop is from 3000 to 2000 which is expected. Something needs to be fixed...

paroj:
do you have a reproducer for this? (maybe with a PR to add btHeightfieldShape to OgreBullet ^^)

slapin:
Will add as soon as I get to my workstation, around June 22-23. I have a code of course, but now I'm travelling :(

slapin:
An yes, the PR is happenning too as I got how to integrate colliders and paged terrain.

slapin:
Debugged that some more. It looks like the issue is due to large number of line draws when debugging runs for heightfields. I guess that is a shortcoming with constant updates of ManualObject. It is kind of usable with terrain size of 129, but becomes unusable with size of 513. Perhaps this can be solved by using several ManualObjects...

paroj:
you could try moving this into the header and see whether it helps with performance:https://github.com/OGRECave/ogre/blob/8af1c8b7a329477f9adca39e92fe95cc66ecf190/Components/Bullet/src/OgreBullet.cpp#L776

PR #3432, opened 2025-08-22
[OgreBullet] Remove heightmap colliders from visualization

slapin:
OgreBullet's debug rendering for physics leaves a lot to be desired regarding performance. Heightmap colliders are dense enough to put whole rendering to crawl and be unusable.

CF_DISABLE_VISUALIZE_OBJECT flag allows to exclude objects from debug rendering which is used here to remove heightmap colliders from being drawn.

By discussion I add parameter which allows enabling heightmap collider
debug anyway while it is being disabled by default.

References #3387

### DIFF
--- a/Components/Bullet/src/OgreBullet.cpp
+++ b/Components/Bullet/src/OgreBullet.cpp
@@ -351,7 +351,7 @@ public:
         delete hfdata.terrainHeights;
     }
 };
-btRigidBody* DynamicsWorld::addTerrainRigidBody(TerrainGroup* terrainGroup, long x, long y, int group, int mask)
+btRigidBody* DynamicsWorld::addTerrainRigidBody(TerrainGroup* terrainGroup, long x, long y, int group, int mask, bool debugDraw)
 {
     Terrain* terrain = nullptr;
 #ifdef OGRE_BUILD_COMPONENT_TERRAIN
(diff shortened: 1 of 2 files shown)


Components/Bullet/src/OgreBullet.cpp
```diff
@@ -376,7 +376,8 @@ btRigidBody* DynamicsWorld::addTerrainRigidBody(Terrain* terrain, int group, int
     // transfer ownership to node
     auto objWrapper = std::make_shared<TerrainRigidBody>(rb, mBtWorld, hfdata);
     node->getUserObjectBindings().setUserAny("BtCollisionObject", objWrapper);
-    rb->setCollisionFlags(rb->getCollisionFlags() | btCollisionObject::CF_STATIC_OBJECT);
+    rb->setCollisionFlags(rb->getCollisionFlags() | btCollisionObject::CF_STATIC_OBJECT |
```
paroj, line 379: its better to add a flag to `addTerrainRigidBody(..., debugDraw=true)`, so it is still possible to enable this when needed.

slapin: Ah, right, will do just that.

slapin: Should be fixed now

paroj:
you dont have to manually rebase this unless there are merge conflicts. I can rebase on merge automatically

slapin:
Ok, just did that with other two

paroj (approved):

PR #3449, opened 2025-09-18
Update ImGui to 1.92.3 (this is an update of paroj's diff for current master and latest ImGui)

OllyFunkster:
Not trying to take any credit for this work, just saw that paroj's pull request was stalled and I needed a fork with ImGui v1.92.3 in it so I made one and applied paroj's diff (from https://github.com/paroj/ogre/tree/imgui192)

image


paroj:
imgui 1.92 breaks some core API, so Ogre will stick to 1.91 until Ogre 15.

OllyFunkster:
Ah, thanks for the context. I'll try to keep my fork moderately up-to-date in case any others want an Ogre 14 / newer ImGui combo in the meantime.

paroj:
closing, as I rebased my branch on master now

PR #3459, opened 2025-09-29
OgreBullet: Support manual narrow phase for ghost objects

slapin:
When implementing characters as ghost shapes, they don't interfere  normally (
pass through each other and collision is not registered). To make them collide,
either custom reaction or manual narrow phase is needed. I decided to implement
manual narrow phase, as custom reaction is to be implemented by the user I guess
(but a few helpers are to be implemented for that).


Components/Bullet/src/OgreBullet.cpp
```diff
@@ -971,32 +994,54 @@ bool KinematicMotionSimple::recoverFromPenetration(btCollisionWorld* collisionWo
...
 
         mManifolds += mManifoldArray.size();
-        for (int j = 0; j < mManifoldArray.size(); j++)
+        if (mManifoldArray.size() == 0 && mAllowManualNarrowPhase)
         {
-            btPersistentManifold* manifold = mManifoldArray[j];
-            btScalar directionSign = manifold->getBody0() == mGhostObject ? btScalar(-1.0) : btScalar(1.0);
-            for (int p = 0; p < manifold->getNumContacts(); p++)
+            /* Manual narrow phase: because Bullet does not collide ghosts by default */
+            std::cout << "no manifolds\n";
```
paroj, line 1000:
```suggestion
            LogManager::getSingleton().stream() << "no manifolds\n";
```

paroj (changes requested):

Issue #3467, opened 2025-10-09
Ogre gltf imports miss UV layers after first


slapin:
#### System Information
- Ogre Version: master branch
- Operating System / Platform: GNU/Linux (Ubuntu 22.04)
- RenderSystem: GL ES 2.0

#### Detailled description
When I load mesh geometry and pass through it I can't find any more UVs except first.

```c++
static void getSubmeshUVs(const Ogre::Mesh *mesh, const Ogre::SubMesh *submesh,
			  std::vector<Ogre::Vector2> &uvs, int index)
{
	int j;
	float *pReal;
	Ogre::HardwareVertexBufferSharedPtr vbuf;
	Ogre::VertexData *vertex_data = submesh->useSharedVertices ?
						mesh->sharedVertexData :
						submesh->vertexData;
	const Ogre::VertexElement *uvElem =
		vertex_data->vertexDeclaration->findElementBySemantic(
			Ogre::VES_TEXTURE_COORDINATES, index);
	int vertex_count = 0;
	if (submesh->useSharedVertices)
		vertex_count += mesh->sharedVertexData->vertexCount;
	else
		vertex_count += submesh->vertexData->vertexCount;
	if (!uvElem)
		return;
	OgreAssert(uvs.size() == 0 || uvs.size() == vertex_count,
		   "bad vertex count");
	uvs.resize(vertex_count);
	vbuf = vertex_data->vertexBufferBinding->getBuffer(uvElem->getSource());
	unsigned char *uv = static_cast<unsigned char *>(
		vbuf->lock(Ogre::HardwareBuffer::HBL_READ_ONLY));
	for (j = 0; j < vertex_data->vertexCount; ++j) {
		uvElem->baseVertexPointerToElement(uv, &pReal);
		uvs[j] = Ogre::Vector2(pReal[0], pReal[1]);
		uv += vbuf->getVertexSize();
	}
	vbuf->unlock();
}

```
This function always returns empty std::vector from .glb exported from Blender for indices 1 and above.
Meshes in Blender do have 2 UV layers.

paroj:
the assimp codec only considers uv0 currently:
https://github.com/OGRECave/ogre/blob/0f01724bb8f25cf8813047253d43867450f5553a/PlugIns/Assimp/src/AssimpLoader.cpp#L1132

if assimp sees uv1 too, fixing this is trivial

slapin:
Looks like AssetImporter does not support multiple UV layers.
The code 
```diff
[...]

PR #3468, opened 2025-10-19
Support multiple UV layers with Assimp

slapin:
Support multiple UV texture coordinate layers. The elements are added for all layers found by Assimp.


slapin:
This seems to break nothing but there are issues https://github.com/OGRECave/ogre/issues/3467

PR #3475, opened 2025-11-03
Build Bullet without "_Debug" postfix so find_package finds it

### LINKED ISSUE #3474: OGRE_BUILD_COMPONENT_BULLET does not appear as an option when configuring Ogre in debug
#### System Information
- Ogre Version: 14.4.1
- Operating System / Platform: Windows 11 
- RenderSystem: N/A

#### Detailled description

My steps:
1. Run cmake-gui, point source dir to Ogre sources, binaries dir to empty folder
2. Add a CMAKE_BUILD_TYPE entry set to "Debug"
3. Press "Configure"

This builds Bullet correctly (with __Debug_ postfix) but `find_package(Bullet)` fails to find it. I believe the root cause it that the [FindBullet.cmake](https://cmake.org/cmake/help/latest/module/FindBullet.html) script shipped with CMAKE only sets `Bullet_FOUND` if it finds the release libraries (i.e. with no postfix). 

I think this is basically the same issue as https://github.com/OGRECave/ogre/issues/2511. I've worked around it using the same approach, adding the following to the command to build Bullet in Dependencies.cmake:

`-DCMAKE_DEBUG_POSTFIX=`

I'll put up a PR with my workaround soon.

valverl:
Also got rid of CMAKE warning about using BULLET_ROOT instead of Bullet_ROOT

Fixes #3474

### DIFF
--- a/CMake/Dependencies.cmake
+++ b/CMake/Dependencies.cmake
@@ -198,13 +198,14 @@ if(OGRE_BUILD_DEPENDENCIES AND NOT EXISTS ${OGREDEPS_PATH})
         -DBUILD_ENET=OFF
         -DBUILD_UNIT_TESTS=OFF
         -DCMAKE_RELWITHDEBINFO_POSTFIX= # fixes FindBullet on MSVC
+        -DCMAKE_DEBUG_POSTFIX= # need to remove postfix so default find_package detects Bullet correctly
         -DBUILD_CLSOCKET=OFF
         -DCMAKE_POLICY_VERSION_MINIMUM=3.5
         ${PROJECT_BINARY_DIR}/bullet3-3.25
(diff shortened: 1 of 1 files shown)


paroj:
thanks!

paroj:
on a second look, FindBoost should support the _Debug postfix, no?

https://github.com/Kitware/CMake/blob/6b0663f8604996a68df2b5b6588e49a550dbb091/Modules/FindBullet.cmake#L109

valverl:
I think it does set the `*_LIBRARY_DEBUG` variables correctly but only uses the `*_LIBRARY` ones for the call to `find_package_handle_standard_args()`, which is what sets `Bullet_FOUND`.

PR #3527, opened 2025-12-20
terrain task std::future support

slapin:
Support std::future for Ogre::Terrain tasks.
Stability under valgrind drastically increased with this change.
But it is in no way complete fix but a tiny step forward.

### DIFF
--- a/Components/Terrain/include/OgreTerrain.h
+++ b/Components/Terrain/include/OgreTerrain.h
@@ -38,6 +38,7 @@ THE SOFTWARE.
 #include "OgreTerrainLayerBlendMap.h"
 #include "OgreWorkQueue.h"
 #include "OgreTerrainLodManager.h"
+#include <future>
 
 namespace Ogre
 {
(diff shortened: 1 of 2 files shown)


slapin:
Need to mention that I have very rare hangs with this change which might mean that sometimes tasks are not executed which makes future.wait() wait forever. I don't know haw this can happen, but will investigate...

slapin:
Yes, move semantics for containers in MSVS is not supporterd, so it won't ever work in windows.

slapin:
To support windows I guess std::shared_ptr is needed...

paroj:
> Need to mention that I have very rare hangs with this change which might mean that sometimes tasks are not executed which makes future.wait() wait forever. I don't know haw this can happen, but will investigate...

could you fix that?

slapin:
Well, I can't reproduce it anymore so I guess that would be solved as WorkQueue have test coverage. Other logic like removing futures which was never running would complicate things and still have no 100% warranty.

Issue #3508, opened 2025-12-10
Explicit Texture Unit handling


paroj:
currently we only have `ShaderGenerator::_markNonFFP` to opt out from FFP Texturing.

Instead move `TextureUnitParams` to `RenderState` so we have a global view of the textures and each SRS can mark the TUs it handles itself.

PR #3551, opened 2026-01-09
Docs: programmatically set "texture_index" for SRS_NORMALMAP

chillywillysoft:


chillywillysoft:
The parameter is otherwise undocumented. Should we mention it in [RTSS Texture Unit properties / normal_map](https://ogrecave.github.io/ogre/api/latest/rtss.html#normal_map)? While it could be overridden in script, it wouldn't really make sense to do so...

paroj:
> Should we mention it in RTSS Texture Unit properties

no - I want to change this for Ogre 15 #3508. That parameter is merely needed because the SRS cannot communicate the TU responsibilities properly right now..

PR #3557, opened 2026-01-19
Gizmos: Provide editing gizmos and camera orbit gizmo including sample

AFWEdmonds:


paroj:
also run clang-format on the added .cpp/ .h files

Components/Bites/src/OgreGizmos.cpp
```diff
@@ -0,0 +1,1074 @@
+#include "../include/OgreGizmos.h"
```
paroj:
```suggestion
#include "OgreGizmos.h"
```

paroj:
please *rebase* on master

Components/Bites/src/OgreGizmos.cpp
```diff
@@ -0,0 +1,1124 @@
...
+    pass->setLightingEnabled(false);
+    pass->setVertexColourTracking(TVC_AMBIENT | TVC_DIFFUSE);
+    pass->setDepthCheckEnabled(false);
+    pass->setCullingMode(CULL_NONE);
+    pass->setManualCullingMode(MANUAL_CULL_NONE);
+    pass->setPolygonMode(PM_SOLID);
+    pass->setPolygonModeOverrideable(false);
+    if (alpha)
+        pass->setSceneBlending(SBT_TRANSPARENT_ALPHA);
+}
```
AFWEdmonds: I'm having trouble getting the material in the .material file to work. I've checked that it loads correctly with the requireMaterial from Sample.h, and it's loading fine, but unless I generate the material programmatically as is done here, it doesn't work and just renders white.

paroj:
scaling along 2 axes does not work. Everything is picked correctly though :+1:

paroj:
CI is failing in the Real=double build. You might need to use Vector3f explicitly and cast it to Vector3

ArthurFriedri-ch:
Ready for review again. All selection logic is out of the sample and in the Gizmo classes. This has resulted in a little duplication of camera to viewport ray logic for the selection, for which the gizmos have to have pointers for the camera and window to know viewport size.

paroj:
please resolve the custombuild issues, where Real=double

Samples/Simple/include/Gizmos.h
```diff
@@ -0,0 +1,153 @@
...
+        {
+            mGizmo->setMode(G_ROTATE);
+        }
+        else if (name == "Scale")
+        {
+            mGizmo->setMode(G_SCALE);
+        }
+    }
+
+    // custom shader parameter bindings
```
paroj: review whether all these members are still used

ArthurFriedri-ch:
Ready for approval

paroj:
still failing. you need to cast Vector3f to Vector3 at some places

ArthurFriedri-ch:
should be fixed

PR #3568, opened 2026-01-26
Moving skeletal animation sample

chillywillysoft:
Modifies existing Skeletal Animation sample to add option to hoist animated transform from "root" Bone to SceneNode.

Includes minor modifications to OgreMain:
- Entity::setUpdateBoundingBoxFromSkeleton() only calls Mesh::computeBoneBoundingRadius() if enabling
- OgreMesh.cpp's _computeBoneBoundingRadiusHelper() creates a temporary shadow buffer if necessary
- debug display of Entity's bounding spheres
- select which SceneNodes have their debug axes drawn

Commits are in haphazard order. If desired, I can make a new branch and apply commits in more logical order.

### DIFF
--- a/Samples/Simple/include/SkeletalAnimation.h
+++ b/Samples/Simple/include/SkeletalAnimation.h
@@ -284,11 +284,11 @@ public:
 
         if (method == AnimationUpdater::kMovePeriodic)
         {
-            mEntities[0]->getMesh()->_setBounds(mSneakBoundsMoving);
+            mEntities[0]->getMesh()->_setBounds(mSneakBoundsRootBoneMoving);
         }
         else if (method == AnimationUpdater::kMoveContinuous)
         {
-            mEntities[0]->getMesh()->_setBounds(mSneakBoundsStationary);
+            mEntities[0]->getMesh()->_setBounds(mSneakBoundsRootBoneStatic);
         }
 
         // update move method mode for all models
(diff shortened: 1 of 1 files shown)


paroj:
I will probably have to cherry pick some stuff anyway. I want to do another patch release for 14.5, which means I cannot merge ABI breaks - i.e. anything that touches headers in Main.

paroj:
I could merge the sample changes without the debug drawer additions for the patch release.
Maybe you could squash some of the sample changes: e.g. "tame the lame name" could be squashed into "introduce MovingAnimationStateControllerValue".
If this is too much git for you, I can also do this myself

chillywillysoft:
I'll start a new branch and clean it up. I'll probably wait a couple days and see if I have any other ideas. No rush to put it in the patch release.

chillywillysoft:
Since the main point of these changes is to demonstrate the different way of moving the SceneNode for purposes of tighter bounding boxes and spheres, I think it makes sense to wait until we can implement all the visualization rather than merge the movement without the visualization. Sound right? (I'll still make a new branch with a clean commit history, just trying to plan it out.)

paroj:
ah.. I thought the main point here was to demonstrate "root motion animation", where the motion is driven by the skeleton in contrast to the scene node

chillywillysoft:
> ah.. I thought the main point here was to demonstrate "root motion animation", where the motion is driven by the skeleton in contrast to the scene node

Good point. I'm thinking to get rid of the "periodic" option and just present the "continuous" method as *the* way. It would make the sample easier to understand.

slapin:
Can't wait, looking forward to it so much!

creadmefford:
badass, thank you chillywillysoft!

PR #3582, opened 2026-02-03
Skeletal sample 2

chillywillysoft:


chillywillysoft:
Separating the AnimationUpdater from the RootMotionApplier may seem pointless now but it is in anticipation of another feature I plan to share in the next couple days.

PR #3583, opened 2026-02-03
Devel

paroj:


OgreMain/src/OgreDefaultDebugDrawer.cpp
```diff
@@ -121,7 +121,7 @@ void DefaultDebugDrawer::drawSphere(const Sphere & sphere)
     // TODO: only draw visible arcs (very big or close spheres could have very many facets)
```
chillywillysoft: I was thinking to use the camera frustum for this...

PR #3584, opened 2026-02-04
Animation Time Events

chillywillysoft:
Demonstrates triggering callbacks at certain times in an animation.

### DIFF
--- a/Samples/Simple/include/SkeletalAnimation.h
+++ b/Samples/Simple/include/SkeletalAnimation.h
@@ -535,18 +535,12 @@ static void generateBoundingBox(Entity * entity)
 
 static void createSoundwaveMaterial(const String & material_name, const String & group_name)
 {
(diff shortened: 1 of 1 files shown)


Samples/Simple/include/SkeletalAnimation.h
```diff
@@ -328,6 +533,73 @@ static void generateBoundingBox(Entity * entity)
     entity->getMesh()->_setBounds(sneakBounds);
 }
 
+static void createSoundwaveMaterial(const String & material_name, const String & group_name)
+{
+    // Make texture
```
paroj: check

https://github.com/OGRECave/ogre/blob/a24c3daeaa900efe49d40a51fc745aa2073e1438/Samples/Water/include/Water.h#L40-L63

for a more portable way to do this

PR #3595, opened 2026-02-27
Fix bug crash when write zip

HadesD:


paroj:
this file comes from: https://github.com/kuba--/zip

can you commit your changes there? Otherwise they will get lost the next time I sync the file

paroj:
I have updated zip from upstream. Please check if the problem still persists and open a PR against upstream if necessary

PR #3625, opened 2026-05-01
GLSupport: scope depth buffers to creator context

### LINKED ISSUE #3618: Error prone depth buffer management when using multiple renderwindows
#### System Information
- Ogre Version: 14.3 and 14.5
- Operating System / Platform: Linux (Wayland)
- RenderSystem: OpenGL ES 2

#### Detailled description
When the OpenGL ES 2 rendersystem creates a depth buffer, the current OpenGL context is used. When using multiple render windows, this could be the context from the second window. When choosing which depth buffer gets attached to an FBO; regarding the OGL context, any depth buffer can be chosen. 

We encountered a problem where the (opaque) objects in a scene suddenly get rendered in the wrong order. This occurs when the second window gets closed and gets destroyed through `Ogre::Root::getSingletonPtr()->destroyRenderTarget(...)`. What probably happens is that along with the second window context, the corresponding depth buffers get destroyed as well, while these are attached to FBO's shown on the main window.

Our current workaround is to set the main window's context active any time an FBO is made, which ensures the main window's context is active during depth buffer creation. Setting the main window's context active is done through  `Ogre::Root::getSingletonPtr()->getRenderSystem()->_setViewport(<main_window_vp>)`. After creating the FBO, we restore the previous viewport through another `_setViewport` call.

But perhaps a better solution is to let the `GLES2RenderSystem::_createDepthBufferFor` function always use the main context before actually creating the depth buffer? Or maybe there's a more elegant way to solve this in our own non-Ogre code?

paroj:
fixes #3618

### DIFF
--- a/RenderSystems/GLSupport/src/OgreGLRenderSystemCommon.cpp
+++ b/RenderSystems/GLSupport/src/OgreGLRenderSystemCommon.cpp
@@ -47,6 +47,37 @@ namespace Ogre {
         return StringUtil::format("%4d x %4d", width, height);
     }
 
+    void GLRenderSystemCommon::_unregisterContext(GLContext *context)
+    {
+        for(auto & rt : mRenderTargets)
+        {
+            if(auto target = dynamic_cast<GLRenderTarget*>(rt.second))
+            {
+                if(auto fbo = target->getFBO())
+                    fbo->notifyContextDestroyed(context);
+            }
+        }
+
+        if (mCurrentContext == context)
+        {
+            // Change the context to something else so that a valid context
+            // remains active. When this is the main context being unregistered,
+            // we set the main context to 0.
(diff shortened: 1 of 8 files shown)


RenderSystems/GLSupport/src/OgreGLRenderSystemCommon.cpp
```diff
@@ -47,6 +48,57 @@ namespace Ogre {
...
+            {
+                _switchContext(mMainContext);
+            }
+            else
+            {
+                /// No contexts remain
+                mCurrentContext->endCurrent();
+                mCurrentContext = 0;
+                mMainContext = 0;
+                //mStateCacheManager = 0;
```
FrankGoyens, line 97: Is this intentionally made into a comment?

PR #3645, opened 2026-06-02
this fixes this error

Joilnen:
/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/16.1.1/../../../../include/c++/16.1.1/bits/stl_vector.h:1109:44: error:
      arithmetic on a pointer to an incomplete type 'Ogre::MeshLodUsage'
 1109 |         ptrdiff_t __dif = this->_M_impl._M_finish - this->_M_imp...
      |                           ~~~~~~~~~~~~~~~~~~~~~~~ ^
/usr/local/include/OGRE/OgreMesh.h:497:71: note: in instantiation of member
      function 'std::vector<Ogre::MeshLodUsage>::size' requested here
  497 |   ...getNumLodLevels(void) const { return mMeshLodUsageList.size(); }
      |                                                             ^
/usr/local/include/OGRE/OgreMesh.h:56:12: note: forward declaration of
      'Ogre::MeshLodUsage'
   56 |     struct MeshLodUsage;
      |            ^
1 error generated.

### DIFF
--- a/OgreMain/include/OgreMesh.h
+++ b/OgreMain/include/OgreMesh.h
@@ -68,7 +68,6 @@ namespace Ogre {
             Transformed from user-supplied values with LodStrategy::transformUserValue.
         */
         Real value;
-        
 
         /// Only relevant if mIsLodManual is true, the name of the alternative mesh to use.
         String manualName;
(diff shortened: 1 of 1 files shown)


Joilnen:
it is now working again even without this workaround but the Vulkan render is broken again.

PR #3686, opened 2026-07-28
Main: const Node::convert functions and helper functions to get Pose indices from Mesh

chillywillysoft:
Makes some functions on Node const.

Adds helper functions to get Pose indices from Mesh. Can't add them to an animation by name or pointer so we need to go through the Mesh's PoseList to find the indices.

### DIFF
--- a/OgreMain/src/OgreMesh.cpp
+++ b/OgreMain/src/OgreMesh.cpp
@@ -2278,6 +2278,35 @@ namespace Ogre {
 
     }
     //---------------------------------------------------------------------
+    size_t Mesh::getPoseIndex(const String& name) const
+    {
+        for (size_t i = 0; i < mPoseList.size(); ++i)
+        {
+            if (mPoseList[i]->getName() == name)
+                return i;
+        }
+        StringStream str;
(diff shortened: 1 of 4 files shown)


OgreMain/src/OgreMesh.cpp
```diff
@@ -2278,6 +2278,35 @@ namespace Ogre {
...
+    size_t Mesh::getPoseIndex(const String& name) const
+    {
+        for (size_t i = 0; i < mPoseList.size(); ++i)
+        {
+            if (mPoseList[i]->getName() == name)
+                return i;
+        }
+        StringStream str;
+        str << "No pose called " << name << " found in Mesh " << mName;
+        OGRE_EXCEPT(Exception::ERR_ITEM_NOT_FOUND,
```
paroj, line 2277: 1. there is a newer macro that auto-derives the function name
2. format makes it more readable and is faster

```suggestion
        OGRE_EXCEPT(Exception::ERR_ITEM_NOT_FOUND, StringUtil::format("No pose called %s found in Mesh %s", name.c_str(), mName.c_str());
```

PR #3690, opened 2026-07-29
Fix Skybox material invalidation

chillywillysoft:
The code looked wrong. I think this is what was intended.

Possible remaining issue is that it first disables the depth-writing on the provided material but then might switch to the default material and use it without disabling depth-writing. I don't think we want to modify the default material. Would anything bad happen if we just used the provided material even if it doesn't have cube map texture (and still issue a warning)?

### DIFF
--- a/OgreMain/src/OgreSkyRenderer.cpp
+++ b/OgreMain/src/OgreSkyRenderer.cpp
@@ -173,7 +173,7 @@ void SceneManager::SkyBoxRenderer::create(
     mSceneNode->attachObject(mSkyBoxObj.get());
 
     mSkyBoxObj->setRenderQueueGroup(renderQueue);
-    mSkyBoxObj->begin(materialName, RenderOperation::OT_TRIANGLE_STRIP, groupName);
+    mSkyBoxObj->begin(m, RenderOperation::OT_TRIANGLE_STRIP);
 
     // rendering cube, only using 14 vertices
     const Vector3 cube_strip[14] = {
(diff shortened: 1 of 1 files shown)


slapin:
We reject "invalid" material and then still use different "invalid" material. Does it make sense?
I think we should reject the material only if it is complete non-runner and issue the warning about what is the problem in particular (about having cubemap texture). The default material should be used only if the original is unusable, i.e. invalid material, IMHO.
The other approach would be to create default cubemap+noDT material and use it as fallback, otherwise it will be confusing. In this case we can reject original material based on lack of cubemap in it.

OgreMain/src/OgreSkyRenderer.cpp
```diff
@@ -153,8 +153,10 @@ void SceneManager::SkyBoxRenderer::create(
     if(valid)
     {
         Pass* pass = m->getBestTechnique()->getPass(0);
-        valid = valid && pass->getNumTextureUnitStates() &&
-                pass->getTextureUnitState(0)->getTextureType() == TEX_TYPE_CUBE_MAP;
+        if (pass->getNumTextureUnitStates())
```
paroj, line 156: this allows a material without a texture now, previously a cubemap was required. why did you change that?

paroj:
> We reject "invalid" material and then still use different "invalid" material. Does it make sense?

the default material is used to continue rendering, while issuing a warning, not to salvage the results

OgreMain/src/OgreSkyRenderer.cpp
chillywillysoft: For my sky I use a solid blue untextured material.

paroj: great, this means you dont need a skybox at all. The clear color or a constant plane should be enough

PR #3707, opened 2026-08-12
Main: Root - make sure first frame updates animation and controllers

chillywillysoft:
SceneManagers and ControllerManager, etc. initialize their frame number to 0 and only perform their updates when the Root's frame number is different. But the Root also initializes its frame number to 0 and doesn't increment it until after the first frame, so animations (etc.) are not applied until the second frame, causing visible glitches on the first frame.


paroj:
nice catch!

PR #3725, opened 2026-08-27
Main: SkeletonInstance - remove broken overrides and pass info to ctor

chillywillysoft:
Commit c285df6 de-virtualised Resource::getName(), getHandle(), and getGroup(), which broke SkeletonInstance’s overrides of those (when called virtually). SkeletonInstance just delegated those to its master Skeleton. It can instead grab the info in its constructor and pass it to Skeleton’s private constructor and Resource’s protected constructor which exist only for SkeletonInstance anyway.

Potential drawback: if lots of instances and strings are big, could waste a little memory.

### DIFF
--- a/OgreMain/include/OgreResource.h
+++ b/OgreMain/include/OgreResource.h
@@ -161,8 +161,8 @@ namespace Ogre {
     protected:
         /** Protected unnamed constructor to prevent default construction. 
         */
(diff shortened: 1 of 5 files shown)


paroj:
nice catch! the extra memory is negligible compared to the vfunction calls that we save

PR #3733, opened 2026-09-08
Add platform-specific thread naming in worker threads

HadesD:
Set thread names based on platform in worker thread creation. Currently windows is researching, so implement some os first.

### DIFF
--- a/OgreMain/src/OgreDefaultWorkQueueStandard.cpp
+++ b/OgreMain/src/OgreDefaultWorkQueueStandard.cpp
@@ -64,7 +64,15 @@ namespace Ogre
         mNumThreadsRegisteredWithRS = 0;
         for (size_t i = 0; i < mWorkerThreadCount; ++i)
         {
-            OGRE_THREAD_CREATE(t, [this]() { _threadMain(); });
+            OGRE_THREAD_CREATE(t, [this]() {
+#  if OGRE_PLATFORM == OGRE_PLATFORM_WIN32
+                // std::wstring wname(name.begin(), name.end());
+                // SetThreadDescription(static_cast<HANDLE>(t.native_handle()), wname.c_str());
+#  elif OGRE_PLATFORM == OGRE_PLATFORM_LINUX || OGRE_PLATFORM == OGRE_PLATFORM_APPLE_IOS || OGRE_PLATFORM == OGRE_PLATFORM_ANDROID
+                pthread_setname_np(pthread_self(), mName.substr(0, 15).c_str());
+#  endif
+                _threadMain();
+            });
             mWorkers.push_back(t);
         }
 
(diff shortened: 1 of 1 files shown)


HadesD:
hmm, maybe windows # preprocessor is differ than other OS
idk how to fix it, should ask AI ...

OgreMain/src/OgreDefaultWorkQueueStandard.cpp
```diff
@@ -64,7 +64,18 @@ namespace Ogre
         mNumThreadsRegisteredWithRS = 0;
         for (size_t i = 0; i < mWorkerThreadCount; ++i)
         {
-            OGRE_THREAD_CREATE(t, [this]() { _threadMain(); });
+            auto threadProc = [this]() {
+#  if OGRE_PLATFORM == OGRE_PLATFORM_WIN32
+                // std::wstring wname(mName.begin(), mName.end());
+                // SetThreadDescription(GetCurrentThread(), wname.c_str());
```
paroj, line 70: there is no need to rush this, so I would wait for the windows bits

OgreMain/src/OgreDefaultWorkQueueStandard.cpp
```diff
@@ -64,7 +64,18 @@ namespace Ogre
...
+#  elif OGRE_PLATFORM == OGRE_PLATFORM_LINUX || OGRE_PLATFORM == OGRE_PLATFORM_ANDROID
+                pthread_setname_np(pthread_self(), mName.substr(0, 15).c_str());
```
paroj, line 72: why are you limiting the name to 15 chars here but not in the other branches? Why limit at all?

OgreMain/src/OgreDefaultWorkQueueStandard.cpp
```diff
@@ -64,7 +64,18 @@ namespace Ogre
         mNumThreadsRegisteredWithRS = 0;
         for (size_t i = 0; i < mWorkerThreadCount; ++i)
         {
-            OGRE_THREAD_CREATE(t, [this]() { _threadMain(); });
+            auto threadProc = [this]() {
+#  if OGRE_PLATFORM == OGRE_PLATFORM_WIN32
+                // std::wstring wname(mName.begin(), mName.end());
+                // SetThreadDescription(GetCurrentThread(), wname.c_str());
+#  elif OGRE_PLATFORM == OGRE_PLATFORM_LINUX || OGRE_PLATFORM == OGRE_PLATFORM_ANDROID
```
paroj, line 71: use `defined(__linux__) || defined(__ANDROID__)`

otherwise we end up with BSD into this, which have yet another signature

OgreMain/src/OgreDefaultWorkQueueStandard.cpp
paroj: also mName is the name of the WorkQueue, I think calling this  "WorkerFunc" would be more useful

OgreMain/src/OgreDefaultWorkQueueStandard.cpp
```diff
@@ -64,7 +64,18 @@ namespace Ogre
         mNumThreadsRegisteredWithRS = 0;
         for (size_t i = 0; i < mWorkerThreadCount; ++i)
         {
-            OGRE_THREAD_CREATE(t, [this]() { _threadMain(); });
```
paroj, line 67: finally, move this code into a separate function and call it from _threadMain instead of the lambda

OgreMain/src/OgreDefaultWorkQueueStandard.cpp
HadesD: > why are you limiting the name to 15 chars here but not in the other branches? Why limit at all?

Don't know why but gemini show that linux only allow 16 char (included \0)
But I 've re-searched, no where wrote that, so I will update.
thanks for your comment

> also mName is the name of the WorkQueue, I think calling this "WorkerFunc" would be more useful

 I think only WorkerFunc name is so limit, when user implement another queue, they should be confused, so I decided  better to use mName

